PR #18, opened 2021-02-13
add dockerfile 

arbuilder:
add dockerfile and compose


AriaMoradi:
I rather not merge this because if I do, I have to take responsibility of maintaining it, so I'll just un-comment the link to your repo again.

PR #39, opened 2021-03-25
Standardize Line endings

Syer10:
Standardize line endings
Fixes the issue where if you clone on windows you cannot use the getAndroid.sh in WSL.
Let git know what files are binary, so that git doesnt try to update the line endings for them


AriaMoradi:
That's hell stupid behavior

PR #45, opened 2021-04-01
Fix getRepo function

Syer10:
Just a issue I found when testing other things


AriaMoradi:
It's a manual revert anyways.

Issue #54, opened 2021-04-04
[Bug] Manga details can't be stored to database column because exceeds length (4096)


arbuilder:
## Device information
- Tachidesk version: Docker [arbuild/tachidesk-docker:latest/slim](https://hub.docker.com/r/arbuild/tachidesk-docker)
- Server Operating System: Win 10 Home
- Server Desktop Environment: WSL2 Ubuntu 20.04
- Server JVM version: bundled with docker image 
- Client Web Browser: Google Chrome Version 89.

## Steps to reproduce
1.  browse one piece manga page of mangakakalot or manganelo
2. it hangs and details of manga page doesn't load

### Expected behavior
It shouldn't hang

### Actual behavior
manga details should load

## Other details
screenshot and error log attached below

![Screenshot 2021-04-04 061609](https://user-images.githubusercontent.com/78140816/113504757-bf5a6d80-9557-11eb-8190-e4716e8a9936.png)
[error.txt](https://github.com/Suwayomi/Tachidesk/files/6254291/error.txt)

AriaMoradi:
fixed in #63

PR #63, opened 2021-04-23
Truncate manga description if it's too long

Forgenn:
Fix for #54


AriaMoradi:
This would fix some cases, but truncating the text would be a better solution.

PR #68, opened 2021-05-03
Update everything

Syer10:
- Update everything.
- The update to SystemTray removed some utility function, like the CacheUtil, and the browse url, I replaced the browse Url with a standard java way, which should work for all desktop platforms.
- Remove Jcenter.
- Remove unneeded java gradle plugin, its handled by kotlins.
- Cleanup the build.gradle.kts's, put all tasks under the task block.
- The update to kotlinter now has ktlint checking filenames to classes, so it complained about LegacyBackupRestoreValidator being in a LegacyBackupValidator named file, so I named the object LegacyBackupValidator, I can do the other way if requested.


server/build.gradle.kts
```diff
@@ -114,24 +109,34 @@ val TachideskRevision = Runtime
 
         }
 
+launch4j { //used for windows
+    mainClassName = MainClass
+    bundledJrePath = "jre"
+    bundledJre64Bit = true
```
AriaMoradi, line 129: 64-bit or 32-bit?
It's bundling 32-bit, why turn this flag to true?

Syer10:
Also forgot to mention that this fixes the manifest values, fixing the key of Implementation-Title, and adding the real Implementation-Version

server/src/main/kotlin/ir/armor/tachidesk/server/util/SystemTray.kt
```diff
@@ -10,16 +10,17 @@ package ir.armor.tachidesk.server.util
 import dorkbox.systemTray.MenuItem
 import dorkbox.systemTray.SystemTray
 import dorkbox.systemTray.SystemTray.TrayType
-import dorkbox.util.CacheUtil
-import dorkbox.util.Desktop
```
AriaMoradi, line 14: the library still has this package in it's tests, are you sure it's removed?
https://github.com/dorkbox/SystemTray/blob/6b4893b223312cbecc60d1544a97c48b85ee2a22/test/dorkbox/TestTray.java#L30

server/src/main/kotlin/ir/armor/tachidesk/server/util/SystemTray.kt
```diff
@@ -10,16 +10,17 @@ package ir.armor.tachidesk.server.util
 import dorkbox.systemTray.MenuItem
 import dorkbox.systemTray.SystemTray
 import dorkbox.systemTray.SystemTray.TrayType
-import dorkbox.util.CacheUtil
```
AriaMoradi, line 13: the library still has this package in it's tests, are you sure it's removed?

https://github.com/dorkbox/SystemTray/blob/6b4893b223312cbecc60d1544a97c48b85ee2a22/test/dorkbox/TestTray.java#L29

server/src/main/kotlin/ir/armor/tachidesk/server/util/SystemTray.kt
Syer10: Ya I checked, there is no alternatives in the library. He probably hasnt updated the tests. If you look at the actual code its not there.

server/src/main/kotlin/ir/armor/tachidesk/server/util/SystemTray.kt
Syer10: Ya I checked, there is no alternatives in the library. He probably hasnt updated the tests. If you look at the actual code its not there.

server/build.gradle.kts
Syer10: That config I just moved here because the tasks are now in the tasks block, I didnt edit any of your values other then changing the main class variable name

server/src/main/kotlin/ir/armor/tachidesk/server/util/SystemTray.kt
AriaMoradi: It has never been there(I checked commits from 2017-18) it seems to be

server/src/main/kotlin/ir/armor/tachidesk/server/util/SystemTray.kt
AriaMoradi: It has never been there(I checked commits from 2017-18) it seems to be added on build time.

Syer10: I think I found it, its a implementation thats not visible to projects that just implement the system tray library now
implementation("com.dorkbox:Utilities:1.9")
I can add it

PR #70, opened 2021-05-03
Download directly to file instead of a dir

Syer10:
Should fix Cl build


AriaMoradi:
workaround for https://github.com/michel-kraemer/gradle-download-task/issues/164

PR #82, opened 2021-05-15
Load next chapter when scrolled to the bottom

Forgenn:


AriaMoradi:
A feature like this needs more work
1- A setting to turn it off
2- some kind of transition area to tell the user they are going to the next chapter
3- this will only work on VerticalReader variants, how about others? 
4- when in doubt, steal from Tachiyomi, specially AZ

Forgenn:
Yes I see now it needs some works, I will close the pull request for now.

PR #84, opened 2021-05-16
Load next chapter when getting to the last page

Forgenn:
Load next chapter when getting to the last page, in vertical reader and paged reader.

### DIFF
--- a/webUI/react/src/components/ReaderNavBar.tsx
+++ b/webUI/react/src/components/ReaderNavBar.tsx
@@ -183,11 +183,6 @@ export default function ReaderNavBar(props: IProps) {
             setHideOpenButton(currentScrollPos > prevScrollPos);
             setPrevScrollPos(currentScrollPos);
         }
-        if (settings.loadNextonEnding && (settings.readerType === 'ContinuesVertical' || settings.readerType === 'Webtoon')) {
-            if ((window.innerHeight + window.scrollY) >= document.body.offsetHeight) {
-                window.location.href = `/manga/${manga.id}/chapter/${chapter.index + 1}`;
-            }
-        }
     };
 
     useEffect(() => {
--- a/webUI/react/src/components/ReaderNavBar.tsx
+++ b/webUI/react/src/components/ReaderNavBar.tsx
@@ -143,6 +143,7 @@ export const defaultReaderSettings = () => ({
     staticNav: false,
     showPageNumber: true,
     continuesPageGap: false,
+    loadNextonEnding: false,
     readerType: 'ContinuesVertical',
 } as IReaderSettings);
 
@@ -181,6 +182,11 @@ export default function ReaderNavBar(props: IProps) {
         if (Math.abs(currentScrollPos - prevScrollPos) > 20) {
             setHideOpenButton(currentScrollPos > prevScrollPos);
             setPrevScrollPos(currentScrollPos);
+        }
+        if (settings.loadNextonEnding && (settings.readerType === 'ContinuesVertical' || settings.readerType === 'Webtoon')) {
+            if ((window.innerHeight + window.scrollY) >= document.body.offsetHeight) {
+                window.location.href = `/manga/${manga.id}/chapter/${chapter.index + 1}`;
+            }
         }
     };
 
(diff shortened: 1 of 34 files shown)


AriaMoradi:
There's a problem that we have to address, settings is saved in local storage, so for users upgrading, they won't have the new keys, and when you access `settings.loadNextonEnding`, you'll get `undefined`

I think it can be solved by making sure we have all the keys at the start and set the missing ones to the default

Forgenn:
> There's a problem that we have to address, settings is saved in local storage, so for users upgrading, they won't have the new keys, and when you access `settings.loadNextonEnding`, you'll get `undefined`
> 
> I think it can be solved by making sure we have all the keys at the start and set the missing ones to the default

Then, after retrieving the settings in Reader() in Reader.tsx, should we check if every setting has been initialized and if not, put it to the default?

AriaMoradi:
> Then, after retrieving the settings in Reader() in Reader.tsx, should we check if every setting has been initialized and if not, put it to the default?

I'll handle it myself

webUI/react/src/components/ReaderNavBar.tsx
```diff
@@ -182,6 +183,11 @@ export default function ReaderNavBar(props: IProps) {
             setHideOpenButton(currentScrollPos > prevScrollPos);
             setPrevScrollPos(currentScrollPos);
         }
+        if (settings.loadNextonEnding && (settings.readerType === 'ContinuesVertical' || settings.readerType === 'Webtoon')) {
+            if ((window.innerHeight + window.scrollY) >= document.body.offsetHeight) {
+                window.location.href = `/manga/${manga.id}/chapter/${chapter.index + 1}`;
+            }
+        }
```
AriaMoradi: Each reader should provide some functionality based on the context, put this in `VerticalReader`

AriaMoradi (changes requested):

webUI/react/src/components/ReaderNavBar.tsx
Forgenn: I have moved this to the Vertical reader, but now the gradle check is not working even though I can compile everything locally.

AriaMoradi: I made some changes, your code is not merge-able any more.

AriaMoradi: > I made some changes, your code is not merge-able any more.

actually is, surprising!

webUI/react/src/components/reader/pager/PagedPager.tsx
```diff
@@ -23,13 +23,17 @@ const useStyles = makeStyles({
...
     } = props;
 
     const classes = useStyles();
 
     function nextPage() {
-        if (curPage < pages.length - 1) { setCurPage(curPage + 1); }
+        if (curPage < pages.length - 1) {
+            setCurPage(curPage + 1);
+        } else if (settings.loadNextonEnding) {
+            window.location.href = `/manga/${manga.id}/chapter/${chapter.index + 1}`;
```
AriaMoradi: use this instead: https://reactrouter.com/web/api/Hooks/usehistory

webUI/react/src/components/reader/pager/VerticalPager.tsx
```diff
@@ -20,10 +20,24 @@ const useStyles = makeStyles({
...
 
-export default function VerticalPager(props: IReaderProps) {
-    const { pages, settings, setCurPage } = props;
+export default function VerticalReader(props: IReaderProps) {
+    const {
+        pages, settings, setCurPage, manga, chapter,
+    } = props;
 
     const classes = useStyles();
+    const handleScroll = () => {
```
AriaMoradi: rename to `handleLoadNextonEnding` or something similar

webUI/react/src/components/reader/pager/VerticalPager.tsx
```diff
@@ -20,10 +20,24 @@ const useStyles = makeStyles({
...
+        if ((window.innerHeight + window.scrollY) >= document.body.offsetHeight) {
+            window.location.href = `/manga/${manga.id}/chapter/${chapter.index + 1}`;
+        }
+    };
+    useEffect(() => {
+        window.addEventListener('scroll', handleScroll);
```
AriaMoradi: check for `settings.loadNextonEnding` being true, then add the event listener

AriaMoradi (changes requested):

PR #87, opened 2021-05-18
Smarter Chapters and cleanup

Syer10:
- If chapters are empty and onlineFetch was not set to false, get chapters from the source
 - Replace .firstOrNull()!! with .first() and handle NoSuchElementException in JavaLin
 - Cleanup some other things

### DIFF
--- a/server/src/main/kotlin/ir/armor/tachidesk/impl/Chapter.kt
+++ b/server/src/main/kotlin/ir/armor/tachidesk/impl/Chapter.kt
@@ -27,96 +27,105 @@ import org.jetbrains.exposed.sql.update
 
 object Chapter {
     /** get chapter list when showing a manga */
-    suspend fun getChapterList(mangaId: Int, onlineFetch: Boolean): List<ChapterDataClass> {
-        return if (!onlineFetch) {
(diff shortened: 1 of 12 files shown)


Syer10:
Build error because of your new windows package stuff, not my fault lol

server/src/main/kotlin/ir/armor/tachidesk/impl/Chapter.kt
```diff
@@ -27,96 +27,105 @@ import org.jetbrains.exposed.sql.update
...
+        return if (onlineFetch == true) {
+            getSourceChapters(mangaId)
+        } else {
             transaction {
                 ChapterTable.select { ChapterTable.manga eq mangaId }.orderBy(ChapterTable.chapterIndex to DESC)
                     .map {
                         ChapterTable.toDataClass(it)
                     }
+            }.ifEmpty {
+                // If it was explicitly set to offline dont grab chapters
```
AriaMoradi, line 40: why?

Syer10: If you wanted to be specifically offline you could set it to false, could remove that part though

PR #88, opened 2021-05-18
getAndroid fixes

Syer10:
Also fixed some lints that IntelliJ was complaining about

### DIFF
--- a/AndroidCompat/getAndroid.sh
+++ b/AndroidCompat/getAndroid.sh
@@ -71,7 +71,7 @@ zip --delete android.jar android/text/Html.class
 # Dedup overridden Android classes
 ABS_JAR="$(realpath android.jar)"
 function dedup() {
-    pushd "$1" || exit
+    pushd "$1"
     CLASSES="$(find ./* -type f)"
     echo "$CLASSES" | while read -r class
     do
--- a/AndroidCompat/getAndroid.sh
+++ b/AndroidCompat/getAndroid.sh
@@ -20,7 +20,7 @@ fi
 
 
 # foolproof against running from AndroidCompat dir instead of running from project root
-if [ "$(basename $(pwd))" = "AndroidCompat" ]; then
+if [ "$(basename "$(pwd)")" = "AndroidCompat" ]; then
   cd ..
 fi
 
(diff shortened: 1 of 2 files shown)


AndroidCompat/getAndroid.sh
```diff
@@ -20,15 +20,15 @@ fi
...
+if [ "$(basename "$(pwd)")" = "AndroidCompat" ]; then
   cd ..
 fi
 
 
 echo "Getting required Android.jar..."
 rm -rf "tmp"
 mkdir -p "tmp"
-pushd "tmp"
+pushd "tmp" || exit
```
AriaMoradi: what does this line do?

Syer10: If pushd fails(folder doesnt exist or similar) it will exit the script instead of continuing

AriaMoradi: why would it fail?!

Syer10: Maybe no permissions? It was a lint in IntelliJ

AndroidCompat/getAndroid.sh
```diff
@@ -20,15 +20,15 @@ fi
 
 
 # foolproof against running from AndroidCompat dir instead of running from project root
-if [ "$(basename $(pwd))" = "AndroidCompat" ]; then
+if [ "$(basename "$(pwd)")" = "AndroidCompat" ]; then
```
AriaMoradi, line 23: why not `\"` ?

AndroidCompat/getAndroid.sh
```diff
@@ -59,36 +59,33 @@ zip --delete android.jar javax/*
 echo "Removing java..."
 zip --delete android.jar java/*
 
-echo "Removing overriden classes..."
+echo "Removing overridden classes..."
 zip --delete android.jar android/app/Application.class
 zip --delete android.jar android/app/Service.class
 zip --delete android.jar android/net/Uri.class
-zip --delete android.jar 'android/net/Uri$Builder.class'
+zip --delete android.jar "android/net/Uri\$Builder.class"
```
AriaMoradi: it was better before!

AriaMoradi: readability!

AndroidCompat/getAndroid.sh
AriaMoradi: it will fail if there's space in pwd, good find

Syer10: \\" in that place would actually make the string end instead of just having a quote in the string

AndroidCompat/getAndroid.sh
Syer10: Resolved

Issue #83, opened 2021-05-16
[Feature Request] Better Desktop keyboard controls


AriaMoradi:
Qoute from Asinine#2585 on discord

> Well, simply going to next chapter when pushing right/down arrow (depending on view) on the last page, at the bottom/end of the page could go to the next chapter?
And same for going back to previous, but first page, and at the top/start of the page.
Would be nice if space-bar went to the next page/chapter (if its the last page) once its scrolled down to the bottom of the page too, some other viewers also do that.

AriaMoradi:
I assume that #95 satisfied requirments of this issue, if there are more ideas that you have, open a new issue.
@vignakke

PR #95, opened 2021-05-24
Added some key mappings to navigate pages

Manchewable:
Added functionality:
- mapped left and right array keys to navigate pages
- mapped spacebar to jump to next page

Pretty much mimics Mangadex's vertical reader behaviour.


webUI/react/src/components/reader/pager/VerticalPager.tsx
```diff
@@ -29,36 +29,87 @@ export default function VerticalReader(props: IReaderProps) {
...
+    const pagesRef = useRef<HTMLDivElement[]>([]);
+    let readerCurPage = curPage;
+
+    function nextPage() {
+        if (curPage < pages.length - 1) {
+            setCurPage((x) => {
+                readerCurPage = x + 1;
+                return readerCurPage;
+            });
+            pagesRef.current[readerCurPage].scrollIntoView();
```
AriaMoradi: the lambda function might be slow to execute and `pagesRef.current[readerCurPage].scrollIntoView();` could generate unexpected results.

webUI/react/src/components/reader/pager/VerticalPager.tsx
```diff
@@ -29,36 +29,87 @@ export default function VerticalReader(props: IReaderProps) {
...
+
+    function prevPage() {
+        if (readerCurPage > 0) {
+            setCurPage((page) => {
+                const rect = pagesRef.current[page].getBoundingClientRect();
+                readerCurPage = (rect.y < 0 && rect.y + rect.height > 0) ? page : page - 1;
+                return readerCurPage;
+            });
+        }
+        pagesRef.current[readerCurPage].scrollIntoView();
```
AriaMoradi: same as the other comment.

webUI/react/src/components/reader/pager/VerticalPager.tsx
```diff
@@ -29,36 +29,87 @@ export default function VerticalReader(props: IReaderProps) {
...
+    }
+
+    function keyboardControl(e:KeyboardEvent) {
+        switch (e.key) {
```
AriaMoradi: use https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code
because `case ' '` is a bit confusing to read.

webUI/react/src/components/reader/pager/VerticalPager.tsx
```diff
@@ -29,36 +29,87 @@ export default function VerticalReader(props: IReaderProps) {
     const classes = useStyles();
     const history = useHistory();
 
-    const [initialScroll, setInitialScroll] = useState(-1);
-    const initialPageRef = useRef<HTMLDivElement>(null);
+    const pageRef = useRef<HTMLDivElement>(null);
+    const pagesRef = useRef<HTMLDivElement[]>([]);
+    let readerCurPage = curPage;
```
AriaMoradi: see if you can write this with only const variables?

Generally in React Functional components you should have everything as const(following functional programming rules) and not do things like this. specially you have to give extra attention when dealing with refs.

AriaMoradi (changes requested):

AriaMoradi:
Does this pull request completely satisfy #83? or is there more work to be done after it?

webUI/react/src/components/reader/pager/VerticalPager.tsx
Manchewable: Do you mean that I should avoid using scrollIntoView() or double check that pagesRef.current[readerCurPage] is not null?

AriaMoradi: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous so `pagesRef.current[readerCurPage].scrollIntoView();` might not have the correct `readerCurPage` at the time as it might be set in another thread.

AriaMoradi: Also this https://reactjs.org/docs/refs-and-the-dom.html#dont-overuse-refs your use of refs makes the code nondeterministic and harder to debug, and prone to unexpected states

webUI/react/src/components/reader/pager/VerticalPager.tsx
Manchewable: The reason I added an additional check in there is because if an image is being cut of by the top of the screen, it should reset the view back to top of the image.
Like this:
prev

AriaMoradi: the comment is about readerCurPage

Manchewable:
> Does this pull request completely satisfy #83? or is there more work to be done after it?

I've only added key mappings to the vertical reader, I'm not sure if #83 is asking the feature to be implemented in all readers.

webUI/react/src/components/reader/pager/VerticalPager.tsx
Manchewable: Do you have any suggestions on how I can reduce my use of refs?

AriaMoradi:
I fixed the mentioned issues, but the commit history got a bit polluted, look at the last 3 commits of mine for the real needed changes.

PR #97, opened 2021-05-25
Add reading direction functionality

Manchewable:
## Description
Added an option in the reader to choose the reading direction.

chrome_EsujNvJw0e

## Changes
- Combined the single page reader types into one.
- Added a new attribute ReaderDirection into IReaderSettings.

### DIFF
--- a/webUI/react/src/components/navbar/ReaderNavBar.tsx
+++ b/webUI/react/src/components/navbar/ReaderNavBar.tsx
@@ -141,6 +141,7 @@ export const defaultReaderSettings = () => ({
     continuesPageGap: false,
     loadNextonEnding: false,
     readerType: 'ContinuesVertical',
+    readerDirection: 'LeftToRight',
 } as IReaderSettings);
 
 interface IProps {
@@ -285,18 +286,10 @@ export default function ReaderNavBar(props: IProps) {
                                     value={settings.readerType}
                                     onChange={(e) => setSettingValue('readerType', e.target.value)}
                                 >
-                                    <MenuItem value="SingleLTR">
-                                        Left to right
+                                    <MenuItem value="SinglePage">
+                                        Single Page
 
                                     </MenuItem>
-                                    {/* <MenuItem value="SingleRTL">
-                                        Right to left(WIP)
-
-                                    </MenuItem> */}
-                                    {/* <MenuItem value="SingleVertical">
-                                        Vertical(WIP)
-
-                                    </MenuItem> */}
                                     <MenuItem value="Webtoon">
                                         Webtoon
 
@@ -309,6 +302,22 @@ export default function ReaderNavBar(props: IProps) {
                                         Horizontal(WIP)
 
                                     </MenuItem> */}
+                                </Select>
+                            </ListItem>
+                            <ListItem>
+                                <ListItemText primary="Reading Direction" />
+                                <Select
+                                    value={settings.readerDirection}
+                                    onChange={(e) => setSettingValue('readerDirection', e.target.value)}
+                                >
+                                    <MenuItem value="LeftToRight">
+                                        Left to right
+
+                                    </MenuItem>
+                                    <MenuItem value="RightToLeft">
+                                        Right to left
+
+                                    </MenuItem>
                                 </Select>
                             </ListItem>
                         </List>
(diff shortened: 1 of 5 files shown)


AriaMoradi:
> Combined the single page reader types into one.

They are meant to be the same component, having it as it was is better and less confusing.(Actually I'm not sure, I'll put it to vote)

You can infer control directions form reader type itself i.e. `RTL` or `LTR`

Manchewable:
> You can infer control directions form reader type itself i.e. `RTL` or `LTR`

I think it's cleaner if we separate the control directions from the reader type. 

If we do bundle them all together, then we'll have a whole bunch of options in the "Reader Type" drop down menu. 

If there are more reader types that supports multiple reading directions in the future (i.e. Double Page mode), then there'll be even more reader types.

webUI/react/src/components/navbar/ReaderNavBar.tsx
```diff
@@ -141,6 +141,7 @@ export const defaultReaderSettings = () => ({
     continuesPageGap: false,
     loadNextonEnding: false,
     readerType: 'ContinuesVertical',
+    readerDirection: 'LeftToRight',
```
AriaMoradi, line 144: `Continues Vertical` + `LeftToRight`? this doesn't make sense

Maybe make it `Continues` + `Vertical`?

webUI/react/src/components/navbar/ReaderNavBar.tsx
```diff
@@ -311,6 +304,22 @@ export default function ReaderNavBar(props: IProps) {
...
+                                    onChange={(e) => setSettingValue('readerDirection', e.target.value)}
+                                >
+                                    <MenuItem value="LeftToRight">
+                                        Left to right
+
+                                    </MenuItem>
+                                    <MenuItem value="RightToLeft">
+                                        Right to left
+
+                                    </MenuItem>
```
AriaMoradi, line 320: You kind of need a matrix to handle all possible cases, maybe put it in an object, 
```
{
SinglePage: ['LeftToRight','RightToLeft','Vertical'],
Vertical: ['Continues', 'Webtoon'],
Horizantal: ['LTR','RTL']
}
```

This naming conversion breaks when you have booth Webtoon and Continues?

webUI/react/src/components/reader/pager/VerticalPager.tsx
```diff
@@ -52,17 +52,33 @@ export default function VerticalReader(props: IReaderProps) {
...
+            nextPage();
+        }
+    }
+
+    function goRight() {
+        if (settings.readerDirection === 'LeftToRight') {
+            nextPage();
+        } else {
+            prevPage();
+        }
```
AriaMoradi, line 68: `LeftToRight` and `RightToLeft` for VerticalPager? doesn't make sense?

you'll probably want this in Horizantal?

webUI/react/src/typings.d.ts
```diff
@@ -91,16 +91,19 @@ interface INavbarOverride {
...
 'Webtoon' |
-'SingleVertical' |
-'SingleRTL' |
-'SingleLTR' |
+'SinglePage' |
 'ContinuesHorizontal';
 
+type ReaderDirection =
+'LeftToRight'|
+'RightToLeft';
```
AriaMoradi, line 99: doesn't make sense for vertical

AriaMoradi (changes requested):

webUI/react/src/components/navbar/ReaderNavBar.tsx
Manchewable: What `Continues Vertical` + `LeftToRight` does is it moves to `next/prev` page when `right/left` arrow keys are pressed.

And `Continues Vertical` + `RightToLeft` does the opposite when the arrow keys are pressed.

webUI/react/src/components/navbar/ReaderNavBar.tsx
Manchewable: That'll indeed break the naming convention yeah.

I'm treating the reader types as how the images are being layout in the reader, and the reading direction as the direction you need to press or click to go towards next page. 

What's your opinion on it?

Manchewable:
Closing this as #105 addresses the issues.

PR #105, opened 2021-05-28
Add a Double Page Viewer

Manchewable:
## Description
Added a Double Page Viewer for reading manga.

chrome_ztWOl4VLXQ
chrome_vgj2jHuZfJ

## Changes
- Added Double Page reading mode.
- Added Right to Left reading direction for Single Page reading mode.
- Added an attribute onImageLoad in Page.

### DIFF
--- a/webUI/react/src/components/manga/reader/Page.tsx
+++ b/webUI/react/src/components/manga/reader/Page.tsx
@@ -7,9 +7,10 @@
 
 import CircularProgress from '@material-ui/core/CircularProgress';
 import { makeStyles } from '@material-ui/core/styles';
+import { CSSProperties } from '@material-ui/core/styles/withStyles';
 import React, { useEffect, useRef, useState } from 'react';
 
-function imageStyle(settings: IReaderSettings): any {
+function imageStyle(settings: IReaderSettings): CSSProperties {
     if (settings.readerType === 'DoubleLTR' || settings.readerType === 'DoubleRTL') {
         return {
             display: 'block',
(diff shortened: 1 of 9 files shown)


webUI/react/src/components/manga/reader/Page.tsx
```diff
@@ -9,6 +9,28 @@ import CircularProgress from '@material-ui/core/CircularProgress';
 import { makeStyles } from '@material-ui/core/styles';
 import React, { useEffect, useRef, useState } from 'react';
 
+function imageStyle(settings: IReaderSettings): any {
```
AriaMoradi: return type is `React.CSSProperties` 
ref: https://stackoverflow.com/questions/45019775/type-for-style-attribute-passed-to-function

webUI/react/src/components/manga/reader/pager/DoublePagedPager.tsx
```diff
@@ -0,0 +1,223 @@
...
+        flexDirection: (settings.readerType === 'DoubleLTR') ? 'row' : 'row-reverse',
+        justifyContent: 'center',
+        margin: '0 auto',
+        width: 'auto',
+        height: 'auto',
+        overflowX: 'scroll',
+    },
+});
+
+export default function DoublePagedReader(props: IReaderProps) {
```
AriaMoradi: DoublePagedPager?

AriaMoradi (changes requested):

webUI/react/src/components/manga/reader/Page.tsx
Manchewable: I think the correct return type is actually `CSSProperties` from `@material-ui/core/styles/withStyles`. 
Getting an error from the compiler saying `Type 'CSSProperties' is not assignable` when I use `React.CSSProperties`.

PR #107, opened 2021-05-28
Update README.md

arbuilder:
add docker download info


AriaMoradi:
no, that's not a good place for it

PR #109, opened 2021-05-28
Add Last Page Read offset

Syer10:
This is something I can use in JUI to go to the spot the person left the chapter at, exactly

### DIFF
--- a/server/src/main/kotlin/suwayomi/server/database/migration/M0007_AddPageOffset.kt
+++ b/server/src/main/kotlin/suwayomi/server/database/migration/M0007_AddPageOffset.kt
@@ -1,0 +1,25 @@
+package suwayomi.server.database.migration
+
+import org.jetbrains.exposed.sql.transactions.TransactionManager
+import org.jetbrains.exposed.sql.transactions.transaction
+import org.jetbrains.exposed.sql.vendors.currentDialect
+import suwayomi.server.database.migration.lib.Migration
+
+/*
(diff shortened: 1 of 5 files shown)


AriaMoradi:
what's the unit that offset is measured in? I proposed a float percentage value

AriaMoradi:
Isn't this going to make the reader make too many update requests?

Syer10:
> 
> 
> Isn't this going to make the reader make too many update requests?

I only send a update request when the window is closed, not sure how the WebUI does it



> 
> 
> what's the unit that offset is measured in? I proposed a float percentage value

It seems to be a unit thats set based the offset of the window compared to the top of the image, a quick test gave me 538 in compose. I dont believe I could make it work with a float percentage value as easily

AriaMoradi:
> It seems to be a unit thats set based the offset of the window compared to the top of the image, a quick test gave me 538 in compose. I dont believe I could make it work with a float percentage value as easily

We have to find a cross implementation solution, either a percent offset or a real(not window) pixels from start offset

Syer10:
Maybe we could find a way to store client data in the db per manga instead? Like some endpoints to store and request key value info per manga

Syer10:
Maybe a table like this
```
--------------------------------
| key       | value | chapterRef |
| juioffset | 538   | 10         |
--------------------------------
```

Syer10:
Superseded by https://github.com/Suwayomi/Tachidesk/pull/113

PR #112, opened 2021-05-29
Working shared preferences

Syer10:
Working version of shared preferences using Java Preferences and a Kotlin wrapper

### DIFF
--- a/AndroidCompat/src/main/java/xyz/nulldev/androidcompat/io/sharedprefs/JavaSharedPreferences.kt
+++ b/AndroidCompat/src/main/java/xyz/nulldev/androidcompat/io/sharedprefs/JavaSharedPreferences.kt
@@ -1,0 +1,168 @@
+package xyz.nulldev.androidcompat.io.sharedprefs
+
+import android.content.SharedPreferences
+import com.russhwolf.settings.ExperimentalSettingsApi
+import com.russhwolf.settings.ExperimentalSettingsImplementation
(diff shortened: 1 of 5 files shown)


AndroidCompat/src/main/java/xyz/nulldev/androidcompat/io/sharedprefs/JavaSharedPreferences.kt
```diff
@@ -0,0 +1,168 @@
...
+import java.util.prefs.PreferenceChangeListener
+import java.util.prefs.Preferences
+
+@OptIn(ExperimentalSettingsImplementation::class, ExperimentalSerializationApi::class, ExperimentalSettingsApi::class)
+class JavaSharedPreferences(key: String) : SharedPreferences {
+    private val javaPreferences = Preferences.userRoot().node("suwayomi/tachidesk/$key")
+    private val preferences = JvmPreferencesSettings(javaPreferences)
+    private val listeners = mutableMapOf<SharedPreferences.OnSharedPreferenceChangeListener, PreferenceChangeListener>()
+
+    // TODO: 2021-05-29 Need to find a way to get this working with all pref types
```
AriaMoradi, line 24: now or later?

Syer10: I would say later, I dont see it being used at all

PR #113, opened 2021-05-29
Add meta info for clients to store custom data in

Syer10:
Also change some objects to classes so that they are not taking up memory for the lifetime of the application.
The custom data is key value pairs, they will be available through the referenced manga or chapter object as normal json key values
For example if you added reader mode(with the key juiReaderMode) such as webtoon to a manga object, this is what will show up when you request that manga from the api again
```json
{
    "id": 10,
    "title": "Isekai manga",
    "meta": {
        "juiReaderMode": "webtoon"
    }
}
```


Syer10:
Gradle wrapper check doesnt seem to be my fault

server/src/main/kotlin/suwayomi/server/database/migration/M0007_MangaAndChapterMeta.kt
```diff
@@ -0,0 +1,38 @@
...
+import suwayomi.tachidesk.model.table.MangaTable
+
+/*
+ * Copyright (C) Contributors to the Suwayomi project
+ * 
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+class M0007_MangaAndChapterMeta : Migration() {
```
AriaMoradi: M0007 is taken, update your clone

server/src/main/kotlin/suwayomi/tachidesk/TachideskAPI.kt
```diff
@@ -207,6 +221,19 @@ object TachideskAPI {
             ctx.status(200)
         }
 
+        // used to modify a chapter's meta paramaters
```
AriaMoradi, line 224: chapter -> manga

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/server/database/migration/M0007_MangaAndChapterMeta.kt
Syer10: Done

server/src/main/kotlin/suwayomi/tachidesk/TachideskAPI.kt
Syer10: Done

server/src/main/kotlin/suwayomi/server/database/migration/M0007_MangaAndChapterMeta.kt
Syer10: Actually not, once sec

Syer10: Now its done

PR #123, opened 2021-06-01
Update publish.yml

arbuilder:
add docker build workflow

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -88,6 +88,5 @@ jobs:
 
       - name: Run Docker build workflow
         run: |
-          cd master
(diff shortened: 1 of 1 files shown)


arbuilder:
@AriaMoradi u have to create and add DOCKER_BUILD_ACCESS_TOKEN (public_repo scope will do) , github_token wont work (https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token)

AriaMoradi:
secrets.DEPLOY_PREVIEW_TOKEN should have the perms, use that

arbuilder:
> secrets.DEPLOY_PREVIEW_TOKEN should have the perms, use that

If u are sure about that I will just do that, since it would just need public_repo scope to run

AriaMoradi:
I

> > secrets.DEPLOY_PREVIEW_TOKEN should have the perms, use that
> 
> If u are sure about that I will just do that, since it would just need public_repo scope to run

Yes, it hast that permission

PR #163, opened 2021-07-30
ConfigurableExtension(PreferenceScreen) support

AriaMoradi:


AriaMoradi:
@Syer10 What do you think about this way of implementation?

Syer10 (approved):

PR #168, opened 2021-08-03
the plural of Manga is Manga, not Mangas

KraXen72:


AriaMoradi:
Plural of manga *is* mangas.

AriaMoradi:
I'll merge because that reads better

K0-RR:
Well technically there is no plural since this is a Japanese word
but it's commonly used in English so yeah both Manga and Mangas will be correct.

K0-RR:
For reference, not that anyone asked
"In more general, commonly used, contexts, the plural form will also be manga. However, in more specific contexts, the plural form can also be mangas e.g. in reference to various types of mangas or a collection of mangas."
"With loan words, singular/plural distinction can be left as-is in the originating language, or English grammar rules can be applied to create a plural form. “Sheeps” is grammatically valid, but unusual."

So yeah just create an en_jp.json / en_jp.po or whatever format Tachidesk uses and make it use "Manga" as plural. There, everyone's happy.

PR #171, opened 2021-08-04
Compressed assets.

### LINKED ISSUE #169: I compressed the icon.
Normally I open prs with optimized assets (lossless compression by `oxipng`) but this time I noticed the https://github.com/Suwayomi/Tachidesk/blob/master/webUI/src/public/favicon.png and https://github.com/Suwayomi/Tachidesk/blob/master/server/src/main/resources/icon/faviconlogo.png are huge (in dimensions and weight). Does this icon need to be this big? I don't think it is displayed anywhere in its full resolution.

I also tried some lossy compression - reduced the color palette to 10 since the icon is simple enough. It went from 579 KB to 48,0 KB. Then resized it to 2000x and got this 18.7 KB file:
favicon (1)

K0-RR:
Logos got cropped to 512x and their color palettes were reduced to 256.
Closes #169.


AriaMoradi:
compressed versions are already in use, thanks for the effort though.

K0-RR:
Ok, but the icon is still that huge, I suppose that's on purpose but it can be optimized losslessly with `oxipng -arvZ -o max --strip all --fix` and the filesize goes from 579 KiB to 287 KiB.
I committed it to a new branch on my fork - https://github.com/RDKRACZ/Tachidesk-Server/blob/oxipng/server/src/main/resources/icon/faviconlogo.png and also tried Github bot http://imgbot.net/ here - https://github.com/RDKRACZ/Tachidesk-Server/pull/1, and here is the file for comparison - https://github.com/RDKRACZ/Tachidesk-Server/blob/imgbot/server/src/main/resources/icon/faviconlogo.png.

PR #182, opened 2021-09-01
add config option for base url

mgn-norm:
this is to allow for adding a base url path for servers where it is not the root url.


AriaMoradi:
closing as this PR is no longer required.

PR #184, opened 2021-09-03
Create docker_build_stable

arbuilder:
Add docker build workflow


arbuilder:
Needs to add on trigger for stable and create the same workflow for preview

arbuilder:
@AriaMoradi Not sure if [published] includes draft releases, you can create a stable release then only we would know if it works

AriaMoradi:
I'll hold up on this until we confirm it works.

arbuilder:
> I'll hold up on this until we confirm it works.

It will work since preview triggers which is the same code u can merge it

PR #185, opened 2021-09-03
update base url extension name based on talk

mgn-norm:
base url extension take 2


AriaMoradi:
Still not sure if we want this feature or not...

AriaMoradi:
I've finally decided that I don't want to deal with this in WebUI, closing.

PR #186, opened 2021-09-04
Change type of Manga.genre to a List<String>

Syer10:
Since it makes more sense for it to be a list of genre instead of a comma delimited string in the API I changed it. I did not change the DB since there isn't a better way to store it in a column.
Extensions have the freedom to add a space after a comma so I added a utility function that trims all strings in the list.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/anime/model/table/AnimeTable.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/anime/model/table/AnimeTable.kt
@@ -11,8 +11,8 @@ import eu.kanade.tachiyomi.animesource.model.SAnime
 import org.jetbrains.exposed.dao.id.IntIdTable
 import org.jetbrains.exposed.sql.ResultRow
 import suwayomi.tachidesk.manga.impl.MangaList.proxyThumbnailUrl
-import suwayomi.tachidesk.manga.impl.util.lang.trimAll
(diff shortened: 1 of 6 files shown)


server/src/main/kotlin/suwayomi/tachidesk/anime/model/table/AnimeTable.kt
```diff
@@ -48,7 +49,7 @@ fun AnimeTable.toDataClass(mangaEntry: ResultRow) =
         mangaEntry[artist],
         mangaEntry[author],
         mangaEntry[description],
-        mangaEntry[genre],
+        mangaEntry[genre]?.split(",")?.trimAll().orEmpty(),
```
AriaMoradi: how about adding a `.toGenreList()` helper function for this?

Syer10: That function would be available to every String if we did it that way, I can do it though

AriaMoradi: yeah, and make it `inline` and `internal`

Syer10: `inline` as well? It wont give any performance benefits :pepehmm:

Syer10: Done

PR #192, opened 2021-09-11
Update Proto models to match Tachi preview

Syer10:
There was a recent PR that fixes the Proto number mistake, I ported it to Tachidesk. But we might want to wait for it to be released in Tachi stable before we merge this.


AriaMoradi:
Is this a breaking change? do we have to detect older backup files and deal with them somehow?

Syer10:
It isn't breaking no, we just don't populate the broken source list now, only the fixed version. This is the same as how Tachi preview does it.

PR #196, opened 2021-09-17
add Local Source

AriaMoradi:
Should be tested with https://github.com/Suwayomi/Tachidesk-WebUI/pull/33


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
```diff
@@ -75,7 +75,7 @@ object Manga {
 
             transaction {
                 MangaTable.update({ MangaTable.id eq mangaId }) {
-
+                    it[MangaTable.title] = fetchedManga.title
```
Syer10: I think we should guard against title changes before we implement this, downloads rely on the title

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
```diff
@@ -151,14 +155,20 @@ object Manga {
...
         return getCachedImageResponse(saveDir, fileName) {
-            getManga(mangaId) // make sure is initialized
-
             val mangaEntry = transaction { MangaTable.select { MangaTable.id eq mangaId }.first() }
 
             val sourceId = mangaEntry[MangaTable.sourceReference]
             val source = getHttpSource(sourceId)
 
-            val thumbnailUrl = mangaEntry[MangaTable.thumbnail_url]!!
+            val thumbnailUrl: String = mangaEntry[MangaTable.thumbnail_url]
```
Syer10: Is there a specific reason why we need a thumbnail url, Tachi allows null urls without issue

server/src/main/kotlin/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -1,153 +1,170 @@
...
-//        const val HELP_URL = "https://tachiyomi.org/help/guides/reading-local-manga/"
-//
-//        private const val COVER_NAME = "cover.jpg"
-//        private val SUPPORTED_ARCHIVE_TYPES = setOf("zip", "rar", "cbr", "cbz", "epub")
-//
-//        private val POPULAR_FILTERS = FilterList(OrderBy())
-//        private val LATEST_FILTERS = FilterList(OrderBy().apply { state = Filter.Sort.Selection(1, false) })
-//        private val LATEST_THRESHOLD = TimeUnit.MILLISECONDS.convert(7, TimeUnit.DAYS)
-//
+        const val LANG = "localsourcelang"
```
Syer10, line 47: Why not `all` or `other`?

server/src/main/kotlin/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -168,117 +185,123 @@ class LocalSource(private val context: Context) : CatalogueSource {
...
+        return Observable.just(MangasPage(mangas.toList(), false))
+    }
+
+    override fun fetchLatestUpdates(page: Int) = fetchSearchManga(page, "", LATEST_FILTERS)
+
+    override fun fetchMangaDetails(manga: SManga): Observable<SManga> {
+        File(applicationDirs.localMangaRoot, manga.url).listFiles().orEmpty().toList()
+            .firstOrNull { it.extension == "json" }
+            ?.apply {
+                val reader = this.inputStream().bufferedReader()
```
Syer10: Grab the latest manga info code from Tachi to use Kotlinx.Serialization for this

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
AriaMoradi: JSON can change title of a manga: https://tachiyomi.org/help/guides/local-manga/#advanced

And we only get the title when the manga is first introduced into the database and there's no way to update it, the source might update the name and `fetchMangaDetails` can give the new name? I'm not sure how it works in Tachiyomi though...

Syer10: In main Tachi the json allows deciding the name of the manga, but it does not update it for future refreshes

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
AriaMoradi: thumbnail url can absolutely be null, that NPE is for triggering a HTTP 404 in javalin.

Syer10: That works then 👍

server/src/main/kotlin/eu/kanade/tachiyomi/source/LocalSource.kt
AriaMoradi: sources use both those langs and having a special lang makes categorizing it into a different section easier.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
Syer10: Basically no title changes are copied over to the db in main Tachi, most forks allow title changes because they properly rename the downloads folder.

server/src/main/kotlin/eu/kanade/tachiyomi/source/LocalSource.kt
AriaMoradi: hmm, my clone is old then.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
AriaMoradi: > most forks allow title changes because they properly rename the downloads folder.

We can do the same, rename the download dir when old title is not the same as the new title?

Syer10: Yep, that would work

PR #206, opened 2021-09-25
Workaround StdLib issue and add KtLint to all modules

Syer10:
This adds a workaround for the BuildConfig plugin messing with the StdLib, and it adds KtLint to all Modules. Dont ask me why this fixes it, it just works 😄 .

### DIFF
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -5,7 +5,7 @@ import java.time.Instant
 plugins {
     application
     id("com.github.johnrengelman.shadow") version "7.0.0"
-    id("com.github.gmazzo.buildconfig") version "3.0.3"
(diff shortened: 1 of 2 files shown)


server/build.gradle.kts
```diff
@@ -127,20 +126,13 @@ tasks {
...
     test {
         useJUnit()
     }
 
-    withType<ShadowJar> {
-        destinationDirectory.set(File("$rootDir/server/build"))
-    }
-
-    named("run") {
-        dependsOn(":formatKotlin", ":lintKotlin")
```
AriaMoradi, line 141: will run still call formatKotlin?

Syer10: Yep, this will be done whenever a CompileKotlin task is run, which is done for any build task

PR #225, opened 2021-10-23
Create view for unread and download badges

### LINKED ISSUE #224: [Feature Request] Implement Unread Count on Manga level
The Api should return how many Chapters of a Manga in the library are unread.

This would enable the UIs to implement a Unread Badge. Furthermore a proper implementation would make it possible to implement the Tachiyomi Style Filter Unread/Read on the Category Level.

mrsOwlex:
Implements unread and download counters for Mangas in Libraries Closes #224


AriaMoradi:
So https://github.com/JetBrains/Exposed/issues/798 , I'm not really sure how you could do it with Exposed.

mrsOwlex:
Cheating. Exposed is actually just not able to create a View. But in sql a
view is like a table, so the resulting statement is the same. I wanna make
sure no one will try updating on the data entity

Aria Moradi ***@***.***> schrieb am Sa. 23. Okt. 2021 um
22:05:

mrsOwlex:
So extended /api/v1/category/{categoryId} to return the MangaViewDataClass.
This I mapped by MangaView which is a sql-view.

Feature of it are the `download_count` and `unread_count`

As previous discussions were talking about performance I included a basic test, so we can switch around to a different solution in the future.

PR #227, opened 2021-10-24
Expose unread and download count of Manga in category api

### LINKED ISSUE #224: [Feature Request] Implement Unread Count on Manga level

mrsOwlex:
Implements unread and download counters for Mangas in Libraries Closes #224

extended /api/v1/category/{categoryId} to return the MangaViewDataClass.
This I mapped by MangaView which is a sql-view.

Feature of it are the download_count and unread_count

As previous discussions were talking about performance I included a basic test, so we can switch around to a different solution in the future.


server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/MangaDataClass.kt
```diff
@@ -2,7 +2,7 @@ package suwayomi.tachidesk.manga.model.dataclass
 
 /*
  * Copyright (C) Contributors to the Suwayomi project
- * 
+ *
```
AriaMoradi: hmm, fix please.

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/model/view/MangaView.kt
```diff
@@ -0,0 +1,75 @@
...
+import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
+import suwayomi.tachidesk.manga.model.dataclass.toGenreList
+import suwayomi.tachidesk.manga.model.table.MangaStatus.Companion
+
+/**
+ * This is a dirty hack to implement View with exposed.
+ * Exposed has actually no concept of a View, so it has been manually created by a sql Migration
+ * Selects on this view work fine, but modifying queries are neither adviced nor tested
+ * One could override the modifying functions or just don't use them
+ */
```
AriaMoradi: I'm not really a fan of introducing Views to the codebase specially when the ORM doesn't support them. It's a neat trick and works for this case, but the solution is likely to be reused in the future and I don't want it to spread.

Sorry, we have to make due with simple queries.

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/model/view/MangaView.kt
mrsOwlex: Fine. I'll create it as a query then. Shouldn't be that hard now that I have a test 😋

If we would implement filters on the API I'd recommend revisiting this solution as it is much more straightforward.

mrsOwlex: Implemented it now i a fully supported way. Ready for a new review

AriaMoradi:
I may do some cleanup afterwards

PR #232, opened 2021-10-28
Fix category reorder Endpoint. Added Test for Category Reorder

mrsOwlex:
`PATCH /api/v1/category/reorder` was failing with a NumberFormatException.
Cause was wrong order of the Routes in the MangaApi.
Wrote a Test while Debugging, so adding this here as well


server/src/test/kotlin/suwayomi/tachidesk/manga/controller/CategoryControllerTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        System.setProperty("$CONFIG_PREFIX.server.rootDir", dataRoot)
+        applicationSetup()
+    }
+
+    @Test
+    fun categoryReorder() {
+        Category.createCategory("foo")
+        Category.createCategory("bar")
+        val cats = Category.getCategoryList()
+        val foo = cats.stream().filter { c -> c.name == "foo" }.findFirst().orElseThrow()
```
Syer10: Do this in a kotlin way(avoid streams) 😉 
```suggestion
        val foo = cats.asSequence().filter { it.name == "foo" }.first()
```

mrsOwlex: Thanks for the hint :)

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -77,11 +77,13 @@ object MangaAPI {
             get("", CategoryController::categoryList)
             post("", CategoryController::categoryCreate)
 
+            // The order here is important {categoryId} needs to be applied last 
+            // or throws a NumberFormatException
+            patch("reorder", CategoryController::categoryReorder)
+
```
AriaMoradi, line 83: I don't understand the change here really, changing the order of declarations and also the meaning of the comment.

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
mrsOwlex: So a Little down below there's a patch route, which starts with {categoryId}. Javelin is trying to apply the routes in order of definition. So when the patch with categoryId is defined before the reorder, then reorder will never be called because the wild card {categoryId} will always match before.

Issue #234, opened 2021-11-01
[Feature Request] Library Update Option


Sehraan:
## What feature should be added to Tachidesk?
Library update option.

##  Why/Project's Benefit/Existing Problem
Library update option needs to be added because it will help the user to get new chapters easily and fast. Currently in Tachidesk, to update a manga we have to go open the manga in library section and reload the page or wait to get new chapter because it doesn't show updates until a comic (or Item in library) is opened so you have to go to all the manga in library to know if there is a new chapter released for that manga.

AriaMoradi:
This feature is planned for the next release.

Sehraan:
on preview or stable version of Tachidesk. @AriaMoradi

AriaMoradi:
I think it's not implemented yet, you can track the work on our discord, and help here and there if you are interested.

Sehraan:
@AriaMoradi i am already a member of your discord server.
Is your username on discord armor

AriaMoradi:
> @AriaMoradi i am already a member of your discord server. Is your username on discord armor

yes

PR #235, opened 2021-11-01
Implement Update of Library

mrsOwlex:
Early early Stages of Implementation of #234

This PR implements Library and Categorie Update

We are introducing a singleton service here, which is processing new jobs.
At every progress change a Flow will be update with a summarizing object of the current status.
The Route /api/v1/update/summary also provides a way to get a summary object about the most current job.


mrsOwlex:
There could be some tests added and a maybe some documentation. What's most important: It works. Looking forward to your feedback guys!

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
```diff
@@ -22,4 +38,54 @@ object UpdateController {
             }
         )
     }
+
+    fun categoryUpdate(ctx: Context) {
+        val categoryId = Optional.ofNullable(ctx.formParam("category")).map { it.toIntOrNull() }.orElse(null)
```
Syer10:
```suggestion
        val categoryId = ctx.formParam("category")?.toIntOrNull()
```
Kotlins null system is basically Optional

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
```diff
@@ -22,4 +38,54 @@ object UpdateController {
...
+        addCategoriesToUpdateQueue(categoriesForUpdate, true)
+        ctx.status(HttpCode.OK)
+    }
+
+    private fun addCategoriesToUpdateQueue(categories: List<CategoryDataClass>, clear: Boolean = false) {
+        val updater by DI.global.instance<IUpdater>()
+        if (clear) {
+            runBlocking { updater.reset() }
+        }
+        for (category in categories) {
```
Syer10:
```suggestion
        categories.forEach { category ->
```

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
```diff
@@ -22,4 +38,54 @@ object UpdateController {
...
+            val mangas = CategoryManga.getCategoryMangaList(category.id)
+            for (manga in mangas) {
```
Syer10:
```suggestion
            mangas.forEach { manga ->
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/UpdateStatus.kt
```diff
@@ -0,0 +1,35 @@
+package suwayomi.tachidesk.manga.impl.update
+
+import mu.KotlinLogging
+import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
+
+var logger = KotlinLogging.logger {}
+class UpdateStatus {
+    var statusMap = HashMap<JobStatus, ArrayList<MangaDataClass>>()
```
Syer10:
```suggestion
    var statusMap = mutableMapOf<JobStatus, MutableList<MangaDataClass>>()
```
Avoid raw java types, unless you specifically need a non-standard function from them

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/UpdateStatus.kt
```diff
@@ -0,0 +1,35 @@
...
+    var running: Boolean = false
+    var numberOfJobs: Int = 0
+
+    constructor(jobs: List<UpdateJob>, running: Boolean) {
+        this.running = running
+        this.numberOfJobs = jobs.size
+        jobs.forEach {
+            val list = statusMap.getOrDefault(it.status, ArrayList())
```
Syer10:
```suggestion
            val list = statusMap.getOrDefault(it.status, mutableListOf())
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -0,0 +1,72 @@
...
+import suwayomi.tachidesk.manga.impl.Chapter
+import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
+
+class Updater : IUpdater {
+    private val logger = KotlinLogging.logger {}
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+
+    private var isRunning: Boolean = false
+
+    private var tracker = HashMap<String, UpdateJob>()
```
Syer10:
```suggestion
    private var tracker = mutableMapOf<String, UpdateJob>()
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private fun createUpdateJob(): Job {
+        return scope.launch {
+            while (true) {
+                val job = updateChannel.receive()
+                process(job)
+                isRunning = !updateChannel.isEmpty
+                val value = UpdateStatus(tracker, isRunning)
+                statusChannel.emit(value)
```
Syer10:
```suggestion
                statusChannel.value = value
```
Assigning the stateflows state, which will automatically emit if the old value and the new value is not equal

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+    }
+
+    private suspend fun process(job: UpdateJob) {
+        isRunning = true
+        job.status = JobStatus.RUNNING
+        tracker["${job.manga.id}"] = job
+        statusChannel.emit(UpdateStatus(tracker, isRunning))
```
Syer10:
```suggestion
        statusChannel.value = UpdateStatus(tracker, isRunning)
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -0,0 +1,72 @@
...
+        try {
+            logger.info { "Updating ${job.manga.title}" }
+            Chapter.getChapterList(job.manga.id, true)
+            job.status = JobStatus.COMPLETE
+        } catch (e: Exception) {
```
Syer10, line 48: Re-throw cancellation exceptions here

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -0,0 +1,72 @@
...
+            logger.error(e) { "Error while updating ${job.manga.title}" }
+            job.status = JobStatus.FAILED
+        }
+        tracker["${job.manga.id}"] = job
+    }
+
+    @OptIn(DelicateCoroutinesApi::class)
+    override fun addMangaToQueue(manga: MangaDataClass) {
+        GlobalScope.launch { updateChannel.send(UpdateJob(manga)); }
```
Syer10:
```suggestion
        scope.launch { updateChannel.send(UpdateJob(manga)); }
```
Avoid GlobalScope, so if the regular scope is cancelled send wont crash the app

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -0,0 +1,72 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import mu.KotlinLogging
+import suwayomi.tachidesk.manga.impl.Chapter
+import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
+
+class Updater : IUpdater {
+    private val logger = KotlinLogging.logger {}
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+
+    private var isRunning: Boolean = false
```
Syer10: Maybe check if channel.isActive or updateJob.isActive instead?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/UpdateStatus.kt
```diff
@@ -0,0 +1,35 @@
+package suwayomi.tachidesk.manga.impl.update
+
+import mu.KotlinLogging
+import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
+
+var logger = KotlinLogging.logger {}
+class UpdateStatus {
```
Syer10: Maybe make the default constructor a constructor with fields since your just assigning it anyway.

Syer10 (changes requested):
Here is my initial look into it, looks good, can be cleaned up a bit.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
mrsOwlex: Wouldn't that cancel the whole job? My intention here is to accept the error and mark the specific manga as failed.

Syer10: `CancellationException`'s are thrown when the job(thread) is cancelled, so they need to be rethrown in any place where you catch exceptions

Syer10: Like so
```kotlin
try {
    println(channel.receive())
} catch(e: Exception) {
    if (e is CancellationException) throw e
    // log error
}
```

server/build.gradle.kts
```diff
@@ -70,6 +70,9 @@ dependencies {
     // uncomment to test extensions directly
 //    implementation(fileTree("lib/"))
     implementation(kotlin("script-runtime"))
+
+    testImplementation("io.mockk:mockk:1.9.3")
+    testImplementation("io.javalin:javalin:4.0.0")
```
Syer10: Just wondering, how does this help?

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -112,7 +112,11 @@ object MangaAPI {
         }
 
         path("update") {
+            get("recentChapters", UpdateController::recentChapters)
```
Syer10: Rebase error?

server/build.gradle.kts
mrsOwlex: Tested the controller. Needed them to compile :)

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
mrsOwlex: don't actually need isRunning so i'll remove it completely

server/build.gradle.kts
mrsOwlex: for this more explicitly
```kotlin
    @Test
    fun `POST non existent Category Id should give error`() {
        every { ctx.formParam("category") } returns "1"
        UpdateController.categoryUpdate(ctx)
        verify { ctx.status(HttpCode.BAD_REQUEST) }
        val updater by DI.global.instance<IUpdater>()
        assertEquals(0, updater.getStatus().value.numberOfJobs)
    }
 ```

Syer10: Weird, tests should use regular dependencies + test dependencies, not sure why it needs javalin

mrsOwlex: Just tried a `:server:clean :server:test` and it works without it. So i guess i just had a weird run back then.

Syer10: 👍

Syer10 (approved):
Looks good 👍

PR #240, opened 2021-11-04
batch download enqueue

### LINKED ISSUE #181: [Feature Request] Batch download endpoint
Batch download endpoint to request downloading for multiple manga in a single request

AriaMoradi:
Closes #181

Some notes:
- ~~I used the term "bulk" instead of "batch" as a non-native English speaker, it might not be a better term?~~
- I'm not sure how to write unit tests for this one, It's a couple of functions that depend on a much larger library...

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -107,6 +107,7 @@ object MangaAPI {
         }
 
         path("download") {
+            get("bulk", DownloadController::bulkQueueChapter)
             get("{mangaId}/chapter/{chapterIndex}", DownloadController::queueChapter)
             delete("{mangaId}/chapter/{chapterIndex}", DownloadController::unqueueChapter)
         }
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -107,7 +107,7 @@ object MangaAPI {
         }
 
         path("download") {
-            get("bulk", DownloadController::bulkQueueChapter)
+            get("{mangaId}/batch", DownloadController::batchQueueChapter)
             get("{mangaId}/chapter/{chapterIndex}", DownloadController::queueChapter)
             delete("{mangaId}/chapter/{chapterIndex}", DownloadController::unqueueChapter)
         }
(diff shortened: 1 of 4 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/MangaChapterDataClass.kt
```diff
@@ -11,3 +11,8 @@ data class MangaChapterDataClass(
     val manga: MangaDataClass,
     val chapter: ChapterDataClass,
 )
+
+data class MangaIdChapterIdDataClass(
+    val magnaId: Int,
```
Syer10:
```suggestion
    val mangaId: Int,
```

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -107,6 +107,7 @@ object MangaAPI {
         }
 
         path("download") {
+            get("bulk", DownloadController::bulkQueueChapter)
```
Syer10: Why not something like `{mangaId}/download/bulk`? and use a List<Int>

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
AriaMoradi: You can select multiple manga and queue for download in Tachiyomi, this takes care of that as well.

Syer10: I would suggest something like
```kotlin
data class MangaBulkDownload(
    val mangaId: Int
    val all: Boolean = false,
    val startIndex: Int = -1,
    val endIndex: Int = -1 // Or maybe count?
)
```
Since in some places where we would want bulk download we wouldn't have the chapter list available.

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -107,6 +107,7 @@ object MangaAPI {
         }
 
         path("download") {
+            get("{mangaId}/batch", DownloadController::batchQueueChapter)
```
mrsOwlex, line 110: When using a body in the request it's an AnitPattern to do HTTP-GET.

AriaMoradi:
To be suppressed by the download queue rewrite?

Syer10:
Sure

AriaMoradi:
closing for now as this is stale

PR #246, opened 2021-11-08
add pagination to recentChapters

AriaMoradi:


server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/PaginatedList.kt
```diff
@@ -0,0 +1,40 @@
+package suwayomi.tachidesk.manga.model.dataclass
+
+/*
+ * Copyright (C) Contributors to the Suwayomi project
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import java.lang.Integer.min
```
Syer10:
```suggestion
import kotlin.math.min
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -321,17 +323,19 @@ object Chapter {
...
-                        MangaTable.toDataClass(it),
-                        ChapterTable.toDataClass(it)
-                    )
-                }
+    fun getRecentChapters(pageNum: Int): PaginatedList<MangaChapterDataClass> {
+        return paginatedFrom(pageNum) {
+            transaction {
+                (ChapterTable innerJoin MangaTable)
+                    .select { (MangaTable.inLibrary eq true) and (ChapterTable.fetchedAt greater MangaTable.inLibraryAt) }
+                    .orderBy(ChapterTable.fetchedAt to SortOrder.DESC)
```
Syer10, line 331: I would suggest adding this query instead of using a custom function that handles the data after the fact(though I would be fine without it)
```suggestion
                    .orderBy(ChapterTable.fetchedAt to SortOrder.DESC)
                    .limit(50, 50 * pageNum)
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
AriaMoradi: That's what I intend to do in the following pull requests, I devised this design so we can quickly add pagination to various api endpoints before the next major release, all of these pagination changes are breaking, the more we pack the better.

Syer10: Got it 👍

Syer10 (approved):

PR #251, opened 2021-11-15
cleanup directory names

AriaMoradi:
Critically breaking change on users side, maybe having some migration method is desired?
Opinions please.


Syer10 (approved):

mrsOwlex (approved):

Syer10:
Personally redownloading covers would be fine. We could make a quick migration that deletes the original folder or renames it. I would suggest storing that we have done the migration in the preferences or somewhere.

mrsOwlex (approved):

Syer10 (approved):

AriaMoradi:
We can try to do the migration on startup, then remove the code for migration on the next major release.

PR #274, opened 2021-12-19
Attempt to fix #237

### LINKED ISSUE #237: [Bug] Unable to install any extension
## Device information
- Tachidesk version: v0.5.4-r995
- Server Operating System: Windows 10 21H1 (x64)
- Server JVM version: Java 8 Update 311
- Client Operating System: Windows 10 21H1 (x64)
- Client Web Browser: Firefox 94.0 (64-bit)

## Steps to reproduce
1. Try to install an extension.

### Expected behavior
The extension installs successfully and shows up in the sources tab.

### Actual behavior
The extension doesn't install, it just freezes on 'installing' status.

## Other details
The APK file and jar file for the extension gets created successfully, but the extension still doesn't install. The 'install from external source' feature also doesn't work. So in short, I'm unable to install any extensions. The existing extensions work fine, but when I uninstall an existing extension and try to install it again, it doesn't install.

Syer10:
This PR attempts to fix #237 by explicitly setting the standard open options that Files.write is supposed to set by default. I cannot reproduce the issue locally so I cannot test to see if this actually helps.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/BytecodeEditor.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/BytecodeEditor.kt
@@ -18,6 +18,7 @@ import org.objectweb.asm.Opcodes
 import java.nio.file.FileSystems
 import java.nio.file.Files
 import java.nio.file.Path
+import java.nio.file.StandardOpenOption
 import kotlin.streams.asSequence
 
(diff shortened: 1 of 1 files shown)


AriaMoradi:
There's no point in merging this unless we are confidant if it actually fixes anything

AriaMoradi:
closing for now as this is stale

PR #340, opened 2022-04-16
[Feature] Ability to change default WebUi in server.conf

### LINKED ISSUE #325: [Feature Request] Ability to change default WebUi in server.conf
Ability to change default WebUi to different Client's Web UI (Like Sorayomi Web Ui) in server.conf file.

As we have different clients, the ability to change default WebUi will give users more choices.

JiPaix:
closes #325

- add `server.webUIPath` to `server.conf`
- revert to default webUI if no `server.webUIPath` option is provided.
- revert to default webUI if `server.webUIPath = ""`
- relative and absolute path are both valid values.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -29,6 +29,7 @@ class ServerConfig(config: Config, moduleName: String = MODULE_NAME) : SystemPro
 
     // webUI
     val webUIEnabled: Boolean by overridableConfig
+    val webUIPath: String? by overridableConfig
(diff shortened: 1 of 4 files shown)


server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -29,6 +29,7 @@ class ServerConfig(config: Config, moduleName: String = MODULE_NAME) : SystemPro
 
     // webUI
     val webUIEnabled: Boolean by overridableConfig
+    val webUIPath: String? by overridableConfig
```
JiPaix, line 32: optional configuration: no migration needed for users

AriaMoradi:
what exactly does this PR achieve?

PR #341, opened 2022-04-16
[WIP] token auth

AriaMoradi:
The implementation is partial, up for grabs if anyone wants to continue.

Goals of the current implementation:
- Only handle single user mode.
- Potentially receive username and password from the config file(not the best idea?)
- have a "acquire (new) token" endpoint
- Check auth header compatible with [the standard HTTP way](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication) sample `Authorization: Token 9944b09199c62bcf9418ad846dd0e4bbdfc6ee4b`
- Steal implementation details from [django rest framework's token auth](https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication)

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/JavalinSetup.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/JavalinSetup.kt
@@ -56,14 +56,16 @@ object JavalinSetup {
             config.enableCorsForAllOrigins()
 
             config.accessManager { handler, ctx, _ ->
-                fun credentialsValid(): Boolean {
+                fun basicAuthCredentialsValid(): Boolean {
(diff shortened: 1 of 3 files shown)


Robonau:
> receive username and password from the config file

is usually a horrible idea to store the actual username and password in plain text at any time

AdmnJ:
Would this get rid of "Re-auth" on every screen issue on some browsers? Just curious.

Syer10:
It wouldnt

AriaMoradi:
> Would this get rid of "Re-auth" on every screen issue on some browsers? Just curious.

For ipadOS and IOS browsers and any based on safari this would solve the issue.

WantToLearnJapanese:
How's it going?

PR #371, opened 2022-07-01
Run workflow jobs toghether

mahor1221:
* Making release bundles and packages in parallel which reduces workflow build time and makes debugging each separate job easier.
* Added checksum generation to publish.yml

Test run: https://github.com/mahor1221/Tachidesk-Server/actions/runs/2596937668

### DIFF
--- a/.github/workflows/build_push.yml
+++ b/.github/workflows/build_push.yml
@@ -84,16 +84,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [
-          linux-x64,
-          linux-all,
-          macOS-x64,
-          macOS-arm64,
-          # takes about 6m to build deb package
-          #debian-all,
-          windows-x64,
-          windows-x86,
-        ]
+        os:
+          - linux-x64
+          - linux-all
+          - macOS-x64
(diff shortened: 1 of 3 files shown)


.github/workflows/build_push.yml
```diff
@@ -59,41 +57,139 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [
+          linux-x64,
+          linux-all,
+          macOS-x64,
+          macOS-arm64,
+          # takes about 6m to build deb package
+          #debian-all,
```
Syer10: I think it would be fine if this was enabled for previews, it doesnt really matter how long a workflow takes

Robonau: do u mean enabled for releases? previews being fast is kinda useful sometimes

Syer10: Tachidesk isnt making builds like no tomorrow, so a little delay in previews isnt really important enough that we have to disable the creation of deb packages.

mahor1221: You are right, someone can just comment out unwanted parts on their fork.

PR #379, opened 2022-08-04
Add workflow dispatch

mahor1221:
Enables running workflow on different branches manually
I made minor changes to represent this


AriaMoradi:
The workflow is restricted to the master branch and the clone directory is fine as well.

PR #394, opened 2022-08-25
Fix jre folder path

voltrare:
currently after running this line (with the `mkdir -p "$RELEASE_NAME/jre/`), 
the path of jre will be like this `$RELEASE_NAME/jre/jdk8u302-b08-jre/`.

to fix this redacting the line will make the jre files in the right path (`$RELEASE_NAME/jre/`)
also the `$RELEASE_NAME` folder already created at line 37

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -9,4 +9,5 @@ jobs:
       - uses: vedantmgoyal2009/winget-releaser@latest
         with:
           identifier: Suwayomi.Tachidesk-Server
+          installers-regex: '.*x64.msi$'
           token: ${{ secrets.WINGET_PUBLISH_PAT }}


AriaMoradi:
this doesn't fix the issue? And it might break bundles for other platforms

voltrare:
`mv` command will just move the folder inside the `$RELEASE_NAME/` so i think it will work
the code is just moving the extracted jre to the release folder

i dont think it will break in other platform as this is build in github actions ubuntu

and also because only jre for windows is packaged as .zip

voltrare:
the other fix i can try is rename the folder then move it
`mv "$jre_dir" "jre"`
`mv "jre" "$RELEASE_NAME/`

for some reason directly doing it
will just move to the `$jre_dir` to `jre/` folder without ""merging"" it
`mv "$jre_dir" "$RELEASE_NAME/jre"`

voltrare:
if unzip has the same `--strip-components=1` like tar then it can make the job easier

found that bsdtar supports zip files tho
[`bsdtar xvf app.zip --strip-components=1 -C /opt/some-app`](https://superuser.com/a/1665099/1703051)

voltrare:
i have checked the release https://github.com/voltrare/Tachidesk-Server/actions/runs/2926460373
and the jre folder in the correct place

mahor1221:
> for some reason directly doing it will just move to the `$jre_dir` to `jre/` folder without ""merging"" it `mv "$jre_dir" "$RELEASE_NAME/jre"`

A solution could be to use the -T flag:  `mv -T "$jre_dir" "$RELEASE_NAME/jre"`

mahor1221:
> i have checked the release https://github.com/voltrare/Tachidesk-Server/actions/runs/2926460373 and the jre folder in the correct place

I tested this action's windows-x64 artifact in a VM to just to make sure I didn't break another thing after the refactoring:

voltrare:
> A solution could be to use the -T flag: `mv -T "$jre_dir" "$RELEASE_NAME/jre"`

ah this works i tried the -T flag before but got some empty jre folder inside it. But ive retested again now and all good. Tq tq

#396

PR #407, opened 2022-09-27
update bundler and launcher for macos arm64

voltrare:
Fix cant find java error because of wrong path.
As the default  zulu arm jre8 have bit different folder structure `jre/zulu-jre.jre/Contents/...` while
temurin jre8 is `jre/Contents/...`

The macos arm64 need a seperate launcher

### DIFF
--- a/scripts/bundler.sh
+++ b/scripts/bundler.sh
@@ -78,7 +78,7 @@ main() {
       download_jre_and_electron
 
       RELEASE="$RELEASE_NAME.zip"
-      make_macos_bundle
+      make_macosarm_bundle
       move_release_to_output_dir
       ;;
     windows-x86)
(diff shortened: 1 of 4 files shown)


mahor1221:
I sincerely apologize for my mistake which will probably make us release another version. I'm aware this is a volunteered project but I should pay more attention and test all affected releases by my commits before creating PRs.  

As for the fix, this is the original way @AriaMoradi handled Zulu JRE which doesn't need separate launchers:
https://github.com/Suwayomi/Tachidesk-Server/blob/bdb0ad89d4a83318b1dcb15ca1f8ec2928351719/scripts/unix-bundler.sh#L31
https://github.com/Suwayomi/Tachidesk-Server/blob/bdb0ad89d4a83318b1dcb15ca1f8ec2928351719/scripts/unix-bundler.sh#L50-L51

I think it's a better solution although this PR certainly works.

PR #411, opened 2022-10-01
Create metadata per chapter

blindedone1458:


Syer10:
For future reference. if this is implemented it would be preffered to use https://github.com/pdvrieze/xmlutil like Tachiyomi does with https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/domain/manga/model/ComicInfo.kt

PR #415, opened 2022-10-09
replace quickjs with other js engine

L0rnwizd7:
quickjs could not be used in windows, also could not be load in linux system. replace it with jdk8's js engine to excute js code

### DIFF
--- a/AndroidCompat/src/main/java/app/cash/quickjs/QuickJs.java
+++ b/AndroidCompat/src/main/java/app/cash/quickjs/QuickJs.java
@@ -27,8 +27,7 @@ public final class QuickJs implements Closeable {
             Object value = engine.eval(script);
             return translateType(value);
         } catch (Exception exception) {
-            exception.printStackTrace();
-            throw new QuickJsException(exception.getMessage());
+            throw new QuickJsException(exception.getMessage(), exception);
         }
     }
 
--- a/AndroidCompat/src/main/java/app/cash/quickjs/QuickJs.java
+++ b/AndroidCompat/src/main/java/app/cash/quickjs/QuickJs.java
@@ -45,6 +45,10 @@ public final class QuickJs implements Closeable {
             ConsString consString = (ConsString) obj;
             return consString.toString();
         }
+        if (obj instanceof Long) {
+            Long value = (Long) obj;
+            return value.intValue();
+        }
         return obj;
     }
 
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -44,7 +44,6 @@ dependencies {
     implementation("com.squareup.okhttp3:okhttp:$okhttpVersion")
     implementation("io.reactivex:rxjava:1.3.8")
     implementation("org.jsoup:jsoup:1.14.3")
-//    implementation("app.cash.quickjs:quickjs-jvm:0.9.2")
 
     // Sort
     implementation("com.github.gpanther:java-nat-sort:natural-comparator-1.1")
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -44,7 +44,7 @@ dependencies {
     implementation("com.squareup.okhttp3:okhttp:$okhttpVersion")
     implementation("io.reactivex:rxjava:1.3.8")
     implementation("org.jsoup:jsoup:1.14.3")
-    implementation("app.cash.quickjs:quickjs-jvm:0.9.2")
+//    implementation("app.cash.quickjs:quickjs-jvm:0.9.2")
 
     // Sort
     implementation("com.github.gpanther:java-nat-sort:natural-comparator-1.1")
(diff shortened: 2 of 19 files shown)


Syer10:
This is a good idea, have you tested it with ReadComicOnline and Mangahere? They are 2 sources which evaluate JS

server/src/main/java/app/cash/quickjs/QuickJs.java
```diff
@@ -0,0 +1,39 @@
...
+        } catch (Exception exception) {
+            exception.printStackTrace();
+            return "";
+        }
+    }
+
+    public Object evaluate(String script) {
+        try {
+            return engine.eval(script);
+        } catch (Exception exception) {
```
Syer10: Is there a reason to swallow the error? I think that the sources would just error anyway if it returns a empty string since its not the expected output.

server/build.gradle.kts
```diff
@@ -44,7 +44,7 @@ dependencies {
     implementation("com.squareup.okhttp3:okhttp:$okhttpVersion")
     implementation("io.reactivex:rxjava:1.3.8")
     implementation("org.jsoup:jsoup:1.14.3")
-    implementation("app.cash.quickjs:quickjs-jvm:0.9.2")
+//    implementation("app.cash.quickjs:quickjs-jvm:0.9.2")
```
AriaMoradi: Always remove unused code, we are using git, we should use its features!

AriaMoradi:
We are using Mozilla Rhino in Duktape implementation already, maybe it could also replace quickjs

Syer10:
The Duktape implementation can be removed, but we could use Rhino in the QuickJS implementation

server/src/main/java/app/cash/quickjs/QuickJs.java
L0rnwizd7: I test it with manhuagui extension and it work well. But i find Readcomiconline extension will use quickjs's compile function and i have no idea about how this function works.
image

And you are right, I should not swallow the error, may need to throw another runtime exception. I just want the function signature to be same with quickjs.

L0rnwizd7:
I tested with manhuagui, tencentcomics and readcomiconline and all work well. But mangahere will throw NullPointerException which it's quite weird....

L0rnwizd7:
I tested some mangas in mangahere, only the below manga will throw nullpointer exception. maybe it's the extension's problem.
image

AriaMoradi:
We could keep QuickJS and substitute it with Rhino whne we are missing libc++?

Syer10:
> We could keep QuickJS and substitute it with Rhino whne we are missing libc++?

That would be difficult to do, extensions instantiate QuickJS directly. We can't really do much about that unless we add a redirect like we did with SimpleDateFormat


Also I think we should move this file to the AndroidCompat module, since its meant to replace a dependency like the other things there.

server/src/main/java/app/cash/quickjs/QuickJs.java
```diff
@@ -0,0 +1,66 @@
...
+        return this.evaluate(script);
+    }
+
+    public Object evaluate(String script) {
+        try {
+            Object value = engine.eval(script);
+            return translateType(value);
+        } catch (Exception exception) {
+            exception.printStackTrace();
+            throw new QuickJsException(exception.getMessage());
```
Syer10: No need to print the error either, the server will handle that
```suggestion
        } catch (Exception exception) {
            throw new QuickJsException(exception.getMessage());
```

Syer10:
I did some testing with the QuickJs sources, here are the results
```
Comicabc(zh) java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer
Mangahere(en) Works
Readcomiconline(en) Works
QiXiManhua(zh) Works
TencentComics(zh) Works
Manhuagui(zh) Works
```
You probably need to add `Long to Int` in the translateType function

AndroidCompat/src/main/java/app/cash/quickjs/QuickJs.java
```diff
@@ -0,0 +1,65 @@
...
+            Object[] objects = new Object[(int) length];
+            for (int i = 0; i < (int) length; i++) {
+                objects[i] = translateType(array.get(i));
+            }
+            return objects;
+        }
+        if (obj instanceof ConsString) {
+            ConsString consString = (ConsString) obj;
+            return consString.toString();
+        }
```
AriaMoradi, line 47: It should return `Int` if obj is `Long` for the sake of compatibility with quickjs, we had a discussion on discord about it.

AriaMoradi, line 47: It should return `Int` if obj is `Long` for the sake of compatibility with quickjs, we had a discussion on discord about it.

AriaMoradi (changes requested):

AndroidCompat/src/main/java/app/cash/quickjs/QuickJs.java
AriaMoradi: This should fix the issue with Comicabc

PR #424, opened 2022-10-24
Fix: Error handling for popular/latest api if pageNum was supplied as zero

### LINKED ISSUE #75: [Feature Request] Requesting page 0 for popular/latest should throw an error
When requesting page 0 of popular/latest, it should throw a error

Page 0 is a invalid index in the Tachiyomi api, depending on the source, it can error or duplicate results to the next page. I feel the api should throw an error to the client if they try to request this page to allow for easier identification to what the client did wrong

meta-boy:
fixes #75

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/MangaList.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/MangaList.kt
@@ -20,7 +20,6 @@ import suwayomi.tachidesk.manga.model.dataclass.PagedMangaListDataClass
 import suwayomi.tachidesk.manga.model.dataclass.toGenreList
 import suwayomi.tachidesk.manga.model.table.MangaStatus
 import suwayomi.tachidesk.manga.model.table.MangaTable
-import java.lang.IllegalArgumentException
 
 object MangaList {
     fun proxyThumbnailUrl(mangaId: Int): String {
@@ -28,8 +27,13 @@ object MangaList {
     }
 
     suspend fun getMangaList(sourceId: Long, pageNum: Int = 1, popular: Boolean): PagedMangaListDataClass {
-        if (pageNum == 0) {
-            throw IllegalArgumentException("Page 0 is an invalid index")
+        /*
+        * Page < 0 is a invalid index in the Tachiyomi api,
+        * depending on the source, it can error or duplicate results to the next page.
+        * Hence, handling it explicitly below
+        * */
+        require(pageNum > 0) {
+            "Page 0 is an invalid index"
         }
         val source = getCatalogueSourceOrStub(sourceId)
         val mangasPage = if (popular) {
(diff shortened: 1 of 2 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/MangaList.kt
```diff
@@ -20,13 +20,17 @@ import suwayomi.tachidesk.manga.model.dataclass.PagedMangaListDataClass
...
 
 object MangaList {
     fun proxyThumbnailUrl(mangaId: Int): String {
         return "/api/v1/manga/$mangaId/thumbnail"
     }
 
     suspend fun getMangaList(sourceId: Long, pageNum: Int = 1, popular: Boolean): PagedMangaListDataClass {
+        if (pageNum == 0) {
+            throw IllegalArgumentException("Page 0 is an invalid index")
+        }
```
Syer10, line 32: I think we should use `require() {}` since its a Kotlin util meant for argument verification

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/MangaList.kt
meta-boy: Will there be a case where pageNum can be negative?

Syer10: Not any of the current clients do it, but its possible if someone starts sending random numbers to the api endpoint

meta-boy: Understood, updating...

Syer10 (approved):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/MangaList.kt
```diff
@@ -27,6 +27,14 @@ object MangaList {
     }
 
     suspend fun getMangaList(sourceId: Long, pageNum: Int = 1, popular: Boolean): PagedMangaListDataClass {
+        /*
+        * Page < 0 is a invalid index in the Tachiyomi api,
```
AriaMoradi: The comment is too verbose, and the whole reason is already explained in the below string.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/MangaList.kt
```diff
@@ -27,6 +27,14 @@ object MangaList {
...
+        * depending on the source, it can error or duplicate results to the next page.
+        * Hence, handling it explicitly below
+        * */
+        require(pageNum > 0) {
+            "Page 0 is an invalid index"
```
AriaMoradi: `pageNum < 0 is invalid` or `pageNum = $pageNum is not in valid range` or something similar

AriaMoradi (changes requested):
Great idiomatic kotlin solution. 👏

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/MangaList.kt
AriaMoradi: Taking from principles of clean code, comments tend to fall behind from the code and rot. Also don't write comments explaining the code is doing, the code already does that. Only write comments explaining why you are doing something or adding or pointing out certain not obvious context

AriaMoradi (approved):

PR #426, opened 2022-10-25
[WIP] Base for push notification system using websockets

### LINKED ISSUE #425: [Feature Request] Push notification system using websockets
https://github.com/Suwayomi/Tachidesk-Server/projects/1#card-65983015

> Events websocket
> 1- Toast
> 2- Application update available
> 3- new chapters available

meta-boy:
This abstracts the base websocket implementation to have a better base for notification system

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/event/EventDispatcher.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/event/EventDispatcher.kt
@@ -1,0 +1,41 @@
+package suwayomi.tachidesk.event
+
+import io.javalin.websocket.WsContext
+import io.javalin.websocket.WsMessageContext
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CopyOnWriteArrayList
+
+abstract class EventDispatcher<T : EventEntity> {
+    private val clients = ConcurrentHashMap<String, WsContext>()
+    private val eventQueue = CopyOnWriteArrayList<Event<T>>()
+
+    fun addClient(ctx: WsContext) {
+        clients[ctx.sessionId] = ctx
+    }
+
+    fun removeClient(ctx: WsContext) {
+        clients.remove(key = ctx.sessionId)
+    }
+
+    abstract fun notifyClient(ctx: WsContext, event: Event<T>)
+
+    abstract fun handleRequest(ctx: WsMessageContext)
+
+    fun notifyAllClients(event: Event<T>) {
+        clients.forEach {
+            notifyClient(ctx = it.value, event = event)
+        }
+    }
+
+    fun enqueue(event: Event<T>) {
+        eventQueue += event
+    }
+
+    // to be consumed in the client based on
+    // the event type in a notifications screen
+    fun queue() = eventQueue
+
+    fun dequeue(event: Event<T>) {
+        eventQueue.removeIf { it.id == event.id }
+    }
+}
--- a/server/src/main/kotlin/suwayomi/tachidesk/event/enums/EventType.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/event/enums/EventType.kt
@@ -1,0 +1,8 @@
+package suwayomi.tachidesk.event.enums
+
+enum class EventType {
+    SYSTEM,
+    INFORMATIVE,
+    STATIC,
+    DYNAMIC
+}
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -8,62 +8,30 @@ package suwayomi.tachidesk.manga.impl.download
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import io.javalin.websocket.WsContext
-import io.javalin.websocket.WsMessageContext
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.select
 import org.jetbrains.exposed.sql.transactions.transaction
+import suwayomi.tachidesk.event.Event
+import suwayomi.tachidesk.event.enums.EventType
 import suwayomi.tachidesk.manga.impl.Manga.getManga
 import suwayomi.tachidesk.manga.impl.download.model.DownloadChapter
 import suwayomi.tachidesk.manga.impl.download.model.DownloadState.Downloading
 import suwayomi.tachidesk.manga.impl.download.model.DownloadStatus
 import suwayomi.tachidesk.manga.model.table.ChapterTable
 import suwayomi.tachidesk.manga.model.table.toDataClass
-import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArrayList
 
 object DownloadManager {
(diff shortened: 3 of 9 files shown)


AriaMoradi:
The implementation should use a separate ws endpoint, and keep the downloader ws as is for now otherwise a breaking change will occur

meta-boy:
> The implementation should use a separate ws endpoint, and keep the downloader ws as is for now otherwise a breaking change will occur

Yep this ws endpoint is different itself, it's just the notifying service has been extracted and made abstract so that the new WS for toasts and other notifications can reuse the code. The current change in downloader is backwards compatible.

AriaMoradi:
@meta-boy Are you not going to finish this PR?

meta-boy:
> @meta-boy Are you not going to finish this PR?

I will, just taking some time with Okhttp, never worked with it by itself, but only with something like retrofit. So taking my time to explore and write it in the best way possible.

AriaMoradi:
> > @meta-boy Are you not going to finish this PR?
> 
> I will, just taking some time with Okhttp, never worked with it by itself, but only with something like retrofit. So taking my time to explore and write it in the best way possible.

okhttp is for making requests, how does this PR even need sending requests?

meta-boy:
> okhttp is for making requests, how does this PR even need sending requests?

Isn't checking for new release updates also a part of this feature? That would mean querying the github api to check for new releases, which requires making an http call.

AriaMoradi:
> Isn't checking for new release updates also a part of this feature? That would mean querying the github api to check for new releases, which requires making an http call.

Not really? You should instead focus on small incremental pull requests! bigger PRs take a long time to review and merge. Make numerous small PRs instead.

meta-boy:
> > Isn't checking for new release updates also a part of this feature? That would mean querying the github api to check for new releases, which requires making an http call.
> 
> Not really? You should instead focus on small incremental pull requests! bigger PRs take a long time to review and merge. Make numerous small PRs instead.

oh then this is ready

server/src/main/kotlin/suwayomi/tachidesk/event/EventDispatcher.kt
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    fun enqueue(event: Event<T>) {
+        eventQueue += event
+    }
+
+    // to be consumed in the client based on
+    // the event type in a notifications screen
+    fun queue() = eventQueue
```
martinek, line 36: I don't understand what this should be used for. Could you expand on what would be the content of "Notifications screen"?

server/src/main/kotlin/suwayomi/tachidesk/event/enums/EventType.kt
```diff
@@ -0,0 +1,8 @@
+package suwayomi.tachidesk.event.enums
+
+enum class EventType {
```
martinek, line 3: What would be some example events for these types?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -92,12 +64,16 @@ object DownloadManager {
...
+        val status = getStatus()
+        eventDispatcher.enqueue(status)
+        eventDispatcher.notifyAllClients(status)
     }
 
     fun unqueue(chapterIndex: Int, mangaId: Int) {
         downloadQueue.removeIf { it.mangaId == mangaId && it.chapterIndex == chapterIndex }
-        notifyAllClients()
+        val status = getStatus()
+        eventDispatcher.dequeue(status)
```
martinek, line 75: I don't think this would ever do anything. `getStatus()` always returns new `Event` which has new `randomUUID()`. `dequeue(status)` then look for this new id for removal.

Also, I don't think this unqueue is supposed to dequeue the status. `unqueue` is used for removing specific download job. It should notify client with new status (which will be missing the unqueued download task) which it does below. I'm not sure what this code is meant to do.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -107,11 +83,11 @@ object DownloadManager {
         }
 
         if (downloader == null) {
-            downloader = Downloader(downloadQueue) { notifyAllClients() }
+            downloader = Downloader(downloadQueue) { eventDispatcher.notifyAllClients(getStatus()) }
             downloader!!.start()
         }
 
-        notifyAllClients()
+        eventDispatcher.notifyAllClients(getStatus())
```
martinek, line 90: Could `notifyAllClients()` maybe have been left in and only its implementation changed to use `eventDispatcher`? It seems like this line is repeated a lot.

AriaMoradi (commented):
I have several problems with this kind of design:
1- EventDispatcher should be a singleton service class and just do the part of handling of clients and message to them and provide a ws endpoint as the single channel for providing notifications, Each part of the code that wants to send a notification can inject the Event service class and call it when it wants to send a new notification or it could act like some message passing channel or a system job result pipeline for jobs that can't finish quickly like backup and restore, extension install and uninstall
 2- Your Event data class couldn't serialize a notification properly it needs to change somehow
 3- The EventType has no place here, we wouldn't have any of those event types, instead we'd need types like "download finished", "application update available", "backup finished", "backup restore successful", "backup restore failed with this error", "extension installed successfully", etc.
 4- The design is tied to the downloader too tightly, it might eventually let us overhaul the downloader status system but for no means it has to replace the downloader status ws endpoint
 5- The design might need to define it's own text protocol(like http for example) or allow for some kind of back and fourth message passing between a client and the server ( i.e. for requests and services that couldn't be done in one request and response cycle like source global search which would return each search result from each source when it's available)
 
 Finally my ideas are not coherent and maybe this needs to be two ws endpoints, one for true notifications and one for multi step jobs and requests(?)

meta-boy:
> I have several problems with this kind of design: 1- EventDispatcher should be a singleton service class and just do the part of handling of clients and message to them and provide a ws endpoint as the single channel for providing notifications, Each part of the code that wants to send a notification can inject the Event service class and call it when it wants to send a new notification or it could act like some message passing channel or a system job result pipeline for jobs that can't finish quickly like backup and restore, extension install and uninstall 2- Your Event data class couldn't serialize a notification properly it needs to change somehow 3- The EventType has no place here, we wouldn't have any of those event types, instead we'd need types like "download finished", "application update available", "backup finished", "backup restore successful", "backup restore failed with this error", "extension installed successfully", etc. 4- The design is tied to the downloader too tightly, it might eventually let us overhaul the downloader status system but for no means it has to replace the downloader status ws endpoint 5- The design might need to define it's own text protocol(like http for example) or allow for some kind of back and fourth message passing between a client and the server ( i.e. for requests and services that couldn't be done in one request and response cycle like source global search which would return each search result from each source when it's available)
> 
> Finally my ideas are not coherent and maybe this needs to be two ws endpoints, one for true notifications and one for multi step jobs and requests(?)

Understood, taking all points into consideration, let me come up with a proposal on how to actually implement this rather than jumping straight away to the code, this way we can identify issues much earlier

meta-boy:
Closing as stale — no activity in a long while. Reopen if this is still needed.

PR #431, opened 2022-10-29
Add information about last time manga and chapters were fetched from source

### LINKED ISSUE #428: [Feature Request] Store and serve timestamp of last data fetch
Would it be possible for server to store and serve the timestamp of when the data was last fetched from source? Both chapter and manga would benefit from this information.

Even better would probably be for server to store timestamp, but to serve `age` information in API, so client does not have to calculate the age by itself. There might be difference between client and server time which could cause trouble.

With this information, client can be smart about when to fetch the data with `onlineFetch=true`. Currently the client does not know how old the received data is and has no other choice but to always fetch it `online` at least once, to be sure that the data is not stale.

This would allow clients like WebUI to only fetch data online when user requests it manually (ie. with press of a "refresh" button) or when client decides that the data is "stale" (ie. older then 10 minutes or whatever).

martinek:
This PR fixes https://github.com/Suwayomi/Tachidesk-Server/issues/428. It contains:
- migration to add columns on manga where data will be stored
- changes to add new columns to API
- changes to update the data when manga / chapters are fetched from source
- "virtual" properties on manga to show calculated age of data in API

I am not sure about the way age is calculated, but I could not figure out, better way to do that.


server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/MangaDataClass.kt
```diff
@@ -33,11 +34,16 @@ data class MangaDataClass(
...
+    var chaptersLastFetchedAt: Long? = 0,
 
     val freshData: Boolean = false,
     var unreadCount: Int? = null,
     var downloadCount: Int? = null,
-    var chapterCount: Int? = null
+    var chapterCount: Int? = null,
+
+    val age: Long? = if (lastFetchedAt == null) 0 else Instant.now().epochSecond.minus(lastFetchedAt),
+    val chaptersAge: Long? = if (chaptersLastFetchedAt == null) null else Instant.now().epochSecond.minus(chaptersLastFetchedAt)
```
AriaMoradi, line 46: why should the server supply these values?

AriaMoradi (changes requested):
You could merge the two migrations into one by writing a sql migraion
https://github.com/Suwayomi/exposed-migrations/blob/master/lib/src/main/kotlin/de/neonew/exposed/migrations/helpers/AddColumnMigration.kt

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/MangaDataClass.kt
martinek: Theoretically, there could be issue if server and client have different time. These values would make it so that client does not have to calculate the value itself. Realistically, it probably won't ever be an issue.

I thought there would be a better way to add these values to API, but I could not find it. I would prefer if the values were calculated later and not stored on MangaDataClass. Maybe as a getter or something. But I don't know kotlin enough and it would not allow me to add them as getters.

martinek:
> You could merge the two migrations into one by writing a sql migraion https://github.com/Suwayomi/exposed-migrations/blob/master/lib/src/main/kotlin/de/neonew/exposed/migrations/helpers/AddColumnMigration.kt

Is that better then using helper from library though? If there was helper for multiple columns, that would be better, but custom SQL for something that can be done with lib feels less safe to me.

AriaMoradi:
>  If there was helper for multiple columns, that would be better

Well, If you are uncomfortable with doing a direct SQL Migration, make a PR to the migration lib, we own it.

martinek:
I have added the custom sql migration.

AriaMoradi (approved):

Syer10 (approved):

PR #434, opened 2022-11-03
POST variant for `/{sourceId}/search` endpoint

martinek:
This change allows client to call search endpoint with filter values in request body.

Client will be able to simply load source filters and query the source directly, without needing to update filters on server before calling search. This should allow for easier state management and filtering on client.

The filters in body are merged with original source `FilterList` to prevent unexpected values from `filterListCache`. Also they are not stored to server cache, so using them have no effect on `filterListCache`.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -48,7 +48,7 @@ object MangaAPI {
             post("{sourceId}/filters", SourceController.setFilters)
 
             get("{sourceId}/search", SourceController.searchSingle)
-            post("{sourceId}/filter", SourceController.filterSingle)
+            post("{sourceId}/quick-search", SourceController.quickSearchSingle)
 //            get("all/search", SourceController.searchGlobal) // TODO
         }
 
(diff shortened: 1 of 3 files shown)


AriaMoradi (approved):

AriaMoradi:
Make the changes as discussed on discord.

- rename the endpoint and have it be a GET request
- rename change to diff or write a system for transferring whole filter state from a client

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -48,7 +48,7 @@ object MangaAPI {
             post("{sourceId}/filters", SourceController.setFilters)
 
             get("{sourceId}/search", SourceController.searchSingle)
-            post("{sourceId}/search", SourceController.searchSingle)
+            post("{sourceId}/filter", SourceController.filterSingle)
```
AriaMoradi: name it `quick-search` and `quickSearchSingle`

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
martinek: Shouldn't the url use `quickSearch`? Looks like API uses camel case for multiple words (see `recentChapters`).

AriaMoradi: It's one instance of camel case here and one kebab case over at `GlobalAPI`, usually kebab case is used in urls

PR #436, opened 2022-11-05
add batch download api

martinek:
This PR adds `POST /downloads` endpoint for creating chapter downloads.

The path is `POST /downloads` since it best matches REST scheme, since the call is creating Downloads.

For identification of chapters is used combination of `mangaId` and `chapterIndex`. I don't think it is ideal, but chapter ids are not available to client, so this combination needs to be used.

For optimizations manga and chapters are loaded upfront. I don't know how to load only the chapters required for query as there does not seem to be any apparent way to do custom `WHERE` query. Instead all chapters with `sourceIndex` present in input are loaded and then only those actually wanted are used.

If non-existing combination is sent, request does not add anything and returns 404. There is error message in server console, I'm not sure if it is handled enough.

If all chapters are already present in queue, the queue is not started / restarted. Clients are notified thorugh websockets only once after all the chapters are added.

I have also updated existing `GET /download/{mangaId}/chapter/{chapterIndex}` to use the same code.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -105,6 +105,8 @@ object MangaAPI {
             get("start", DownloadController.start)
             get("stop", DownloadController.stop)
             get("clear", DownloadController.stop)
+
+            post("", DownloadController.queueChapters)
         }
 
         path("download") {
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -105,13 +105,12 @@ object MangaAPI {
             get("start", DownloadController.start)
             get("stop", DownloadController.stop)
             get("clear", DownloadController.stop)
-
-            post("batch", DownloadController.queueChapters)
         }
 
         path("download") {
             get("{mangaId}/chapter/{chapterIndex}", DownloadController.queueChapter)
             delete("{mangaId}/chapter/{chapterIndex}", DownloadController.unqueueChapter)
+            post("batch", DownloadController.queueChapters)
         }
 
         path("update") {
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -106,7 +106,7 @@ object MangaAPI {
             get("stop", DownloadController.stop)
             get("clear", DownloadController.stop)
 
-            post("", DownloadController.queueChapters)
+            post("batch", DownloadController.queueChapters)
         }
 
         path("download") {
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
@@ -121,10 +121,10 @@ object DownloadController {
                 summary("Downloader add multiple chapters")
                 description("Queue multiple chapters for download")
             }
-            body<Array<EnqueueInput>>()
+            body<EnqueueInput>()
         },
         behaviorOf = { ctx ->
-            val inputs = json.decodeFromString<List<EnqueueInput>>(ctx.body())
+            val inputs = json.decodeFromString<EnqueueInput>(ctx.body())
             ctx.future(
                 future {
                     DownloadManager.enqueue(inputs)
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
@@ -105,7 +105,7 @@ object DownloadController {
         behaviorOf = { ctx, chapterIndex, mangaId ->
             ctx.future(
                 future {
-                    DownloadManager.enqueue(listOf(EnqueueInput(mangaId, chapterIndex)))
+                    DownloadManager.enqueueWithChapterIndex(mangaId, chapterIndex)
                 }
             )
         },
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
@@ -105,7 +105,7 @@ object DownloadController {
         behaviorOf = { ctx, chapterIndex, mangaId ->
             ctx.future(
                 future {
-                    DownloadManager.enqueueMultiple(listOf(EnqueueInput(mangaId, chapterIndex)))
+                    DownloadManager.enqueue(listOf(EnqueueInput(mangaId, chapterIndex)))
                 }
             )
         },
@@ -121,15 +121,15 @@ object DownloadController {
                 summary("Downloader add multiple chapters")
                 description("Queue multiple chapters for download")
             }
+            body<Array<EnqueueInput>>()
         },
         behaviorOf = { ctx ->
             val inputs = json.decodeFromString<List<EnqueueInput>>(ctx.body())
             ctx.future(
                 future {
-                    DownloadManager.enqueueMultiple(inputs)
+                    DownloadManager.enqueue(inputs)
                 }
             )
-            ctx.status(200)
         },
         withResults = {
             httpCode(HttpCode.OK)
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
@@ -9,13 +9,21 @@ package suwayomi.tachidesk.manga.controller
 
 import io.javalin.http.HttpCode
 import io.javalin.websocket.WsConfig
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.json.Json
+import org.kodein.di.DI
+import org.kodein.di.conf.global
+import org.kodein.di.instance
 import suwayomi.tachidesk.manga.impl.download.DownloadManager
+import suwayomi.tachidesk.manga.impl.download.DownloadManager.EnqueueInput
 import suwayomi.tachidesk.server.JavalinSetup.future
 import suwayomi.tachidesk.server.util.handler
 import suwayomi.tachidesk.server.util.pathParam
 import suwayomi.tachidesk.server.util.withOperation
 
 object DownloadController {
+    private val json by DI.global.instance<Json>()
+
     /** Download queue stats */
     fun downloadsWS(ws: WsConfig) {
         ws.onConnect { ctx ->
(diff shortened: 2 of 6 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
```diff
@@ -107,6 +115,27 @@ object DownloadController {
         }
     )
 
+    val queueChapters = handler(
+        documentWith = {
+            withOperation {
+                summary("Downloader add multiple chapters")
+                description("Queue multiple chapters for download")
+            }
```
Syer10, line 123:
```suggestion
            }
            body<Array<EnqueueInput>>()
```

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
```diff
@@ -107,6 +115,27 @@ object DownloadController {
...
+        },
+        behaviorOf = { ctx ->
+            val inputs = json.decodeFromString<List<EnqueueInput>>(ctx.body())
+            ctx.future(
+                future {
+                    DownloadManager.enqueueMultiple(inputs)
+                }
+            )
+            ctx.status(200)
```
Syer10: No need for this, the default status code is 200 if no exceptions are thrown.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -75,24 +80,77 @@ object DownloadManager {
...
+    private fun addToQueue(manga: MangaDataClass, chapter: ChapterDataClass): DownloadChapter? {
+        if (downloadQueue.none { it.mangaId == manga.id && it.chapterIndex == chapter.index }) {
+            val dc = DownloadChapter(
+                chapter.index,
+                manga.id,
+                chapter,
+                manga
+            )
+            downloadQueue.add(dc)
+            logger.debug("Added to download queue: ${manga.title} | ${chapter.index}")
```
Syer10:
```suggestion
            logger.debug { "Added to download queue: ${manga.title} | ${chapter.index}" }
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -75,24 +80,77 @@ object DownloadManager {
...
+            return dc
+        }
+        logger.debug("Chapter already present in queue: ${manga.title} | ${chapter.index}")
```
Syer10: This helps with performance when debug logging is not enabled
```suggestion
        logger.debug { "Chapter already present in queue: ${manga.title} | ${chapter.index}" }
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -75,24 +80,77 @@ object DownloadManager {
...
-        notifyAllClients()
+    }
+
+    /**
+     * Tries to add chapter to queue.
+     * If chapter is added, returns the created DownloadChapter, otherwise returns null
+     */
+    private fun addToQueue(manga: MangaDataClass, chapter: ChapterDataClass): DownloadChapter? {
+        if (downloadQueue.none { it.mangaId == manga.id && it.chapterIndex == chapter.index }) {
+            val dc = DownloadChapter(
```
Syer10: Use descriptive variable names

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -75,24 +80,77 @@ object DownloadManager {
...
-                        transaction {
-                            ChapterTable.select { (ChapterTable.manga eq mangaId) and (ChapterTable.sourceOrder eq chapterIndex) }
-                                .first()
+    @Serializable
+    data class EnqueueInput(
+        val mangaId: Int,
+        val chapterIndex: Int
+    )
+
+    fun enqueueMultiple(inputs: List<EnqueueInput>) {
```
Syer10: Just keep it named enqueue since its replacing the old function

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -75,24 +80,77 @@ object DownloadManager {
...
+
+    /**
+     * Tries to add multiple inputs to queue
+     * If any of inputs was actually added to queue, starts the queue
+     */
+    private fun addMultipleToQueue(inputs: List<Pair<MangaDataClass, ChapterDataClass>>) {
+        val addedChapters = inputs.map { addToQueue(it.first, it.second) }
+        val anyAdded = addedChapters.any { it != null }
+
+        if (anyAdded) {
```
Syer10:
```suggestion
        val addedChapters = inputs.mapNotNull { addToQueue(it.first, it.second) }
        if (addedChapters.isNotEmpty()) {
```

Syer10 (changes requested):
Some minor comments

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -75,24 +80,75 @@ object DownloadManager {
...
+            MangaTable.select { MangaTable.id inList mangaIds }
+                .map { MangaTable.toDataClass(it) }
+        }
+
+        // This list will have unwanted chapters from other manga but there is no simple way
+        // to select only wanted ones. It will be mapped to input later so it should not be problem
+        val chapters = transaction {
+            val chapterIndexes = inputs.map { it.chapterIndex }.distinct()
+            ChapterTable.select { ChapterTable.sourceOrder inList chapterIndexes }.toList()
+        }
```
Syer10: I don't like this but I'll approve it since I can't think of a better way to do this

Syer10 (approved):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
martinek: I was thinking about this a bit more. Maybe we could just make it take chapter ID instead. Since this is a new API we could add id to chapters in all the endpoints and have this new one take chapter ID. All new batch API endpoints could then use chapter ids.

If this is added and then we would want to change it, it will make things a lot harder.

martinek:
I have changed the `EnqueueInput` to use `chapterIds` instead of combination of `mangaId` + `chapterIndex`. This is more efficient and new `EnqueueInput` is future proof for cases where we will want to use the endpoint for loading chapters directly from library screen.

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -105,6 +105,8 @@ object MangaAPI {
             get("start", DownloadController.start)
             get("stop", DownloadController.stop)
             get("clear", DownloadController.stop)
+
+            post("", DownloadController.queueChapters)
```
AriaMoradi: move the path to `download/batch`

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/ChapterDataClass.kt
```diff
@@ -8,6 +8,7 @@ package suwayomi.tachidesk.manga.model.dataclass
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 data class ChapterDataClass(
+    val id: Int,
```
AriaMoradi, line 11: Why are we exposing chapter id here? 
`XxxxDataClass` is the representation of an object understandable by the api clients.
For client side work we usually use `XxxTable` and `ResultRow`

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/ChapterDataClass.kt
AriaMoradi: > Since this is a new API we could add id to chapters in all the endpoints and have this new one take chapter ID. All new batch API endpoints could then use chapter ids.

If we want to do this then we should simplify api endpoints all over and declare a breaking change. Otherwise it'll not be coherent.

Syer10: https://github.com/Suwayomi/Tachidesk-Server/pull/436#discussion_r1014744951
Its hard to grab specific chapters from the database using a list of manga ids and chapter indexes

AriaMoradi: > Its hard to grab specific chapters from the database using a list of manga ids and chapter indexes

I acknowledge that, but private methods don't have to use data class objects

martinek: @AriaMoradi It is not practical to work with combination of identifiers because Exposed can't natively perform selects with such identifiers (combined keys).

This also simplify API. Client can explicitly identify chapters it wants to execute actions on, instead of using some combinations of related object id and index number. It does not introduce any added complexity, since clients already have to fetch chapters to use theirs index for identification.

Since chapter id alone is enough to identify chapter, there could be new endpoints introduced which would not require mangaId (ie. `/api/v1/chapter/:chapterId`). Clients could migrate to these over time and then breaking change can deprecate old API. (If this is wanted at some point in the future...)

> If we want to do this then we should simplify api endpoints all over and declare a breaking change. Otherwise it'll not be coherent.

This would block this PR for a while and later break all the clients. Why do that if this can be non-breaking change?

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
```diff
@@ -107,6 +115,27 @@ object DownloadController {
         }
     )
 
+    val queueChapters = handler(
```
AriaMoradi, line 118: Missing `pathParam`

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/ChapterDataClass.kt
AriaMoradi: Alright, I accept your argument.

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/DownloadController.kt
Syer10: I don't think it needs one?

martinek: `/download/batch/` does not have any `pathParam`.

AriaMoradi: oh, my bad

AriaMoradi (approved):

PR #438, opened 2022-11-07
add category and global meta

### LINKED ISSUE #312: [Feature Request] Meta key to store data on server level
- Now, Meta key ({baseURL}/api/v1/manga/{mangaId}/meta) is supported for Manga and Chapters.
- Need this Meta key on server level (something like ({baseURL}/api/v1/meta)).
- Need this Meta key on Category level (something like ({baseURL}/api/v1/meta)).

- Can sync client settings across apps 
  - Why: To sync app settings of Soroyomi across different apps like Android, Windows, Web,
- Apply manga filter on a category level.

AriaMoradi:
closes #312

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/global/controller/GlobalMetaController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/global/controller/GlobalMetaController.kt
@@ -1,0 +1,53 @@
+package suwayomi.tachidesk.global.controller
+
+/*
+ * Copyright (C) Contributors to the Suwayomi project
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import io.javalin.http.HttpCode
+import suwayomi.tachidesk.global.impl.GlobalMeta
+import suwayomi.tachidesk.server.util.formParam
+import suwayomi.tachidesk.server.util.handler
+import suwayomi.tachidesk.server.util.withOperation
+
+object GlobalMetaController {
+    /** used to modify a category's meta parameters */
+    val getMeta = handler(
+        documentWith = {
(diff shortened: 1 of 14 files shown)


server/src/main/kotlin/suwayomi/tachidesk/global/controller/GlobalMetaController.kt
```diff
@@ -0,0 +1,53 @@
...
+        formParam<String>("key"),
+        formParam<String>("value"),
+        documentWith = {
+            withOperation {
+                summary("Add meta data to the global meta mapping")
+                description("A simple Key-Value stored at server global level, you can set values for whatever you want inside it.")
+            }
+        },
+        behaviorOf = { ctx, key, value ->
+            GlobalMeta.modifyMeta(key, value)
```
martinek, line 45: Would it be useful to be able to edit multiple keys in one request?

AriaMoradi: Maybe in a future PR

martinek: If it may be added in future, it might be good idea to make it future-proof now. As it is proposed would make it breaking change when changed later.

Also I have noticed that some places in API use JSON body and some places use `fromParam`s. Is there any system in which uses which? I find JSON more practical, since we don't use html forms anywhere.

AriaMoradi: > If it may be added in future, it might be good idea to make it future-proof now. As it is proposed would make it breaking change when changed later.

It's fine as long as we haven't made a stable release.

> Also I have noticed that some places in API use JSON body and some places use fromParams. Is there any system in which uses which? I find JSON more practical, since we don't use html forms anywhere.

I prefer pathParam and formParam while the request expects a few keys, when it takes a lot of or long parameters I use JSON.

PR #449, opened 2022-11-14
Batch editing and deleting any chapter

### LINKED ISSUE #448: [Feature Request] Need an api to change the properties of multiple chapters at once.
Need an api to change the properties of multiple chapters from different mangas at once.

we can use this api `/manga/any/chapters/batch`.

Also need an api to delete multiple chapters from different mangas at once.

martinek:
closes #448 

This PR adds new endpoint `/manga/any/chapter/batch` (notice singular `chapter`) for batch editing any chapters based on list of chapter ids.

I have also added new `change` option to batch edit input. If `delete = true` chapter contents are deleted instead.
This delete functionality is also available on `/manga/:id/chapter/batch` endpoint as it uses same controller.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -69,6 +69,7 @@ object MangaAPI {
 
             get("{mangaId}/chapters", MangaController.chapterList)
             post("{mangaId}/chapter/batch", MangaController.chapterBatch)
+            post("any/chapter/batch", MangaController.anyChapterBatch)
             get("{mangaId}/chapter/{chapterIndex}", MangaController.chapterRetrieve)
             patch("{mangaId}/chapter/{chapterIndex}", MangaController.chapterModify)
             put("{mangaId}/chapter/{chapterIndex}", MangaController.chapterModify)
@@ -77,8 +78,6 @@ object MangaAPI {
             patch("{mangaId}/chapter/{chapterIndex}/meta", MangaController.chapterMeta)
 
             get("{mangaId}/chapter/{chapterIndex}/page/{index}", MangaController.pageRetrieve)
-
-            post("any/chapters/batch", MangaController.anyChapterBatch)
         }
 
         path("category") {
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
@@ -77,6 +77,8 @@ object MangaAPI {
             patch("{mangaId}/chapter/{chapterIndex}/meta", MangaController.chapterMeta)
 
             get("{mangaId}/chapter/{chapterIndex}/page/{index}", MangaController.pageRetrieve)
+
+            post("any/chapters/batch", MangaController.anyChapterBatch)
         }
 
         path("category") {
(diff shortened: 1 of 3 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -198,29 +199,54 @@ object Chapter {
...
+
+    fun modifyChapters(input: MangaChapterBatchEditInput, mangaId: Int? = null) {
         // Make sure change is defined
         if (input.change == null) return
-        val (isRead, isBookmarked, lastPageRead) = input.change
+        val (isRead, isBookmarked, lastPageRead, delete) = input.change
+
+        // Handle deleting separately
+        if (delete == true) {
+            return deleteChapters(input, mangaId)
```
Syer10: Probably shouldn't early return if there are other tasks to do.

```suggestion
            deleteChapters(input, mangaId)
            if (listOfNotNull(isRead, isBookmarked, lastPageRead).isEmpty()) return
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -295,6 +322,42 @@ object Chapter {
         }
     }
 
+    fun deleteChapters(input: MangaChapterBatchEditInput, mangaId: Int? = null) {
```
AriaMoradi:
```suggestion
    private fun deleteChapters(input: MangaChapterBatchEditInput, mangaId: Int? = null) {
```

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/MangaController.kt
```diff
@@ -240,25 +240,49 @@ object MangaController {
...
-            val input = json.decodeFromString<Chapter.ChapterBatchEditInput>(ctx.body())
+            val input = json.decodeFromString<Chapter.MangaChapterBatchEditInput>(ctx.body())
             Chapter.modifyChapters(input, mangaId)
         },
         withResults = {
             httpCode(HttpCode.OK)
         }
     )
 
+    /** batch edit chapters */
```
AriaMoradi:
```suggestion
    /** batch edit chapters from multiple manga */
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -198,29 +199,55 @@ object Chapter {
     data class ChapterChange(
         val isRead: Boolean? = null,
         val isBookmarked: Boolean? = null,
-        val lastPageRead: Int? = null // this probably won't be very useful, but for completion's sake
+        val lastPageRead: Int? = null, // this probably won't be very useful, but for completion's sake
```
AriaMoradi:
```suggestion
        val lastPageRead: Int? = null,
```

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -69,6 +69,7 @@ object MangaAPI {
 
             get("{mangaId}/chapters", MangaController.chapterList)
             post("{mangaId}/chapter/batch", MangaController.chapterBatch)
```
AriaMoradi: please move this to `/api/v1/chapter(s)/batch` if you don't want to put it into v2

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
martinek: I'm not sure I understood what you ment. Please check the change.

PR #450, opened 2022-11-15
Download queue missing update fix

### LINKED ISSUE #446: [Bug] Downloads Web socket is not throwing Finished State for a chapter after chapter downloaded
## Device information
- Tachidesk version: v0.6.5-r1152
- Server Operating System: macOS

## Steps to reproduce
1. Start a download
2. log Web socket output
3. Wait Until it completes

### Expected behaviour
- After Downloaded WS should throw Finished state before removing it from Queue.

### Actual behaviour
- WS is removing download from Queue without throwing Finished state. 

## Other details

```
mangaId: 107, progress: 0.9230769, state: Downloading, tries: 0)], status: Started))

flutter: AsyncData<Downloads>(value: Downloads(queue: [], status: Stopped))
```

martinek:
This PR tries to fix #446

By moving websocket updates to periodically polled updater, some updates (last) might get lost since they are not polled before notifyFlow is cleared. I believe this is caused by behavior of [sample](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/sample.html) function.

@Syer10 I believe you did the downloader rework. Would something like this be correct way to fix it?

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -94,7 +94,9 @@ object DownloadManager {
 
     private fun notifyAllClients(immediate: Boolean = false) {
         if (immediate) {
-            sendStatusToAllClients()
+            scope.launch {
+                sendStatusToAllClients()
+            }
         } else {
             scope.launch {
                 notifyFlow.emit(Unit)
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -80,17 +80,25 @@ object DownloadManager {
     init {
         scope.launch {
             notifyFlow.sample(1.seconds).collect {
-                val status = getStatus()
-                clients.forEach {
-                    it.value.send(status)
-                }
+                sendStatusToAllClients()
             }
         }
     }
 
-    private fun notifyAllClients() {
-        scope.launch {
-            notifyFlow.emit(Unit)
+    private fun sendStatusToAllClients() {
+        val status = getStatus()
+        clients.forEach {
+            it.value.send(status)
+        }
+    }
+
+    private fun notifyAllClients(immediate: Boolean = false) {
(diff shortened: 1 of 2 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -80,17 +80,25 @@ object DownloadManager {
...
+    }
+
+    private fun notifyAllClients(immediate: Boolean = false) {
+        if (immediate) {
+            sendStatusToAllClients()
+        } else {
+            scope.launch {
+                notifyFlow.emit(Unit)
+            }
         }
```
Syer10, line 102: I would suggest doing it this way so that it doesnt block downloads from continueing because of a slow connection to the subscriber.
```suggestion
        if (immediate) {
            scope.launch {
                sendStatusToAllClients()
            }
        } else {
            scope.launch {
                notifyFlow.emit(Unit)
            }
        }
```

Syer10 (changes requested):
It looks fine for the most part.

I think you are using immediate in too many places though, it doesn't need to be in re-order or unqueue as well as most of the download functions. Maybe just when one is added and when its finished.

martinek:
> I think you are using immediate in too many places though, it doesn't need to be in re-order or unqueue as well as most of the download functions. Maybe just when one is added and when its finished.

I was not sure when the bug can appear.

I was thinking about another solution, which would probably be better (more robust). Though I don't know if I fully understand the bug.

It happens when `notifyFlow` is destroyed before last update is sent to clients. Is there a place where this happens (where `notifyFlow` is being destroyed)? Since the `sample(1)` is called from `init` of `DownloadManager`, maybe it happens when `DownloadManager` is being cleaned up from memory or something? I don't understand when `notifyFlow` is disposed. Could we manually call the update then? That would probably fix all the possible cases.

martinek:
@Syer10 After the discussion we had on Discord, I have removed most of the immediate updates. I have left in immediat notification:
- when downloads are added
- when download state changes to `Downloading`
- when download state changes to `Finished`

I have also reverted the change from yesterday. Lets keep the immediate notifications in same thread to make sure the data sent to clients are current ones.

Syer10 (approved):

PR #456, opened 2022-12-04
WebView based cloudflare interceptor

### LINKED ISSUE #323: [Bug] Extension WebtoonXYZ is not working
## Device information
- Tachidesk version: v0.6.2-r1074-windows-x64
- Server Operating System: Windows 10 64-bit
- Server JVM version: bundled with tachidesk win64 
- Client Operating System: Windows 10 64-bit
- Client Web Browser: bundled Electron (I also try using Firefox 98.0.2)

## Steps to reproduce
1. Install WebtoonXYZ extension
2. Go to Sources
3. Click 'Latest' or 'Browse'

### Expected behavior
It should display the list of manga.

### Actual behavior
Spinning loading circle, no manga displayed.

## Other details
Extension does work inside Tachiyomi without any problem.

AriaMoradi:
ported https://github.com/vvanglro/cf-clearance to kotlin

closes #323

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -18,8 +18,8 @@ allprojects {
     repositories {
         mavenCentral()
         google()
-        maven("https://jitpack.io")
         maven("https://github.com/Suwayomi/Tachidesk-Server/raw/android-jar/")
+        maven("https://jitpack.io")
     }
 }
 
--- a/server/build.gradle.kts
+++ b/server/build.gradle.kts
@@ -57,7 +57,7 @@ dependencies {
     implementation("com.github.junrar:junrar:7.5.3")
 
     // CloudflareInterceptor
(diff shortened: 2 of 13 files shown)


AriaMoradi:
Actually doesn't seem to be closing #323 also doesn't work on nhentai either, looks like these extensions are setting their own user agent and it conflicts with the one from playwright hence Tachidesk can't pass cf.

We need a way to suppress the behavior?

also `HttpSource.DEFAULT_USER_AGENT` should probably be set by playwright on runtime?

Syer10:
Is there a reason we can't use the extension UA in playwright?

Syer10 (approved):
Was able to get it working

Syer10:
It works with `LaunchOptions().setHeadless(true)` as well, we could have it true by default, and have a endpoint that launches it in case it fails because of a captcha?

PR #457, opened 2022-12-04
get default User Agent from WebView

AriaMoradi:
This approach potentially will cause a pause and memory spike on user's computer.

Also, PlayWright has to download a bunch of stuff on first execution, that will probably worsen the first user experience.

### DIFF
--- a/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
+++ b/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
@@ -128,6 +128,20 @@ object CFClearance {
             .build()
     }
 
+    fun getWebViewUserAgent(): String {
+        Playwright.create().use { playwright ->
+            val browser = playwright.chromium().launch(
+                LaunchOptions()
+                    .setHeadless(true)
+            )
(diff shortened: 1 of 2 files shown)


server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -128,6 +128,20 @@ object CFClearance {
...
+            val browser = playwright.chromium().launch(
+                LaunchOptions()
+                    .setHeadless(true)
+            )
+            browser.newPage().use { page ->
+                val userAgent = page.evaluate("() => {return navigator.userAgent}") as String
+                logger.debug { "WebView User-Agent is $userAgent" }
+                return userAgent
+            }
+        }
```
Syer10:
```suggestion
            playwright.chromium().launch(
                LaunchOptions()
                    .setHeadless(true)
            ).use { browser ->
                browser.newPage().use { page ->
                    val userAgent = page.evaluate("() => {return navigator.userAgent}") as String
                    logger.debug { "WebView User-Agent is $userAgent" }
                    return userAgent
                }
            }
        }
```

Syer10 (approved):
Just make sure to close the browser and then its good to me 👍

Issue #444, opened 2022-11-09
tachidesk cannot browse 18comic aka "禁漫天堂,jinman tiantang" 


ericvlog:
---

## Device information
- Tachidesk-Server Preview -v0.6.5-r1149 / Tachidesk Deskotp version
- Debian docker and windows 11
- Edge / Chrome

## Steps to reproduce
1. Open tachidesk, go to source, select 禁漫天堂.
2. Select any long comic and open 1st or last chapter 
3. Loading after all it blank.

### Expected behavior
It should be open a comic picture after.

### Actual behavior
it loading and open black blank page.

## Other details
I using debian and install it with docker-compose, and happen as above i descript, then i try download the windows version, test browser mode and electron mode also same, Just extension 18comic or 禁漫天堂 have problem, other extension working great.

I do have android version of tachiyomi, and it working great on the extension.

Hope you can fix it !!

Thanks regard.
![image](https://user-images.githubusercontent.com/64370408/200764861-69b3946f-9733-4e1b-9582-1c810fb6ba4b.png)

ericvlog:
NO more support here?

AriaMoradi:
The extension fails at decoding image files, we have to implement `android.graphics.BitmapFactory.decodeStream` in order to fix this.

```
18:30:30.623 [DefaultDispatcher-worker-6] WARN  io.javalin.Javalin - Uncaught exception
java.lang.RuntimeException: Stub!
	at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:43)
	at eu.kanade.tachiyomi.extension.zh.jinmantiantang.ScrambledImageInterceptor.decodeImage(Unknown Source)
	at eu.kanade.tachiyomi.extension.zh.jinmantiantang.ScrambledImageInterceptor.intercept(Unknown Source)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at eu.kanade.tachiyomi.network.interceptor.SpecificHostRateLimitInterceptor.intercept(SpecificHostRateLimitInterceptor.kt:45)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at eu.kanade.tachiyomi.network.interceptor.CloudflareInterceptor.intercept(CloudflareInterceptor.kt:28)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
[...]

PR #460, opened 2022-12-05
Initial implementation of android.graphics.BitmapFactory

animeavi:
Only what was needed is implemented, compression method is still untested.

This is relevant to #444 
it does not fix the issue yet, stuff from `android.graphics.Canvas` and `android.graphics.Rect` is still needed (I have not looked into how hard that would be for me, so no promises) to support extensions that manipulate images, but this at least gets the image to the extension.

### DIFF
--- a/AndroidCompat/src/main/java/android/graphics/Bitmap.java
+++ b/AndroidCompat/src/main/java/android/graphics/Bitmap.java
@@ -1,0 +1,129 @@
+package android.graphics;
+
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Iterator;
+import javax.imageio.IIOImage;
+import javax.imageio.ImageIO;
+import javax.imageio.ImageWriteParam;
+import javax.imageio.ImageWriter;
+import javax.imageio.stream.ImageOutputStream;
+
+public final class Bitmap {
+    private int width;
+    private int height;
+    private BufferedImage image;
+
+    public Bitmap(BufferedImage image) {
+        this.image = image;
+        this.width = image.getWidth();
+        this.height = image.getHeight();
+    }
+
+    public BufferedImage getImage() {
+        return image;
+    }
+
+    public int getHeight() {
+        return height;
+    }
+
+    public int getWidth() {
+        return width;
+    }
+
+    public enum CompressFormat {
+        JPEG          (0),
+        PNG           (1),
+        WEBP          (2),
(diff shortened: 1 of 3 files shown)


AriaMoradi:
You can steal a lot of source code from android itself usually

Syer10:
That probably won't help much with Bitmaps or Canvas though, a lot of the source uses native code.

animeavi:
I have tested the compression code, turns out OpenJDK does not have a native JPEG decoder so I'll default to PNG no matter what. 
Trying to replicate Canvas with Graphics2D now, not going so well.

Syer10:
Feel free to implement libraries that add extra image support. I think there are libraries that add jpeg support for ImageIO

AriaMoradi:
> Feel free to implement libraries that add extra image support. I think there are libraries that add jpeg support for ImageIO

Also if you need implementations that are available in third party packages you could add them to dependencies and use them, for example to make dateformat work we used `icu4j` from `com.ibm.icu`

animeavi:
I've added the [TwelveMonkeys](https://github.com/haraldk/TwelveMonkeys) JPEG encoder and now JPEG also works.

I got the the Canvas partially working (works with the MangaReader.to currently, but I believe some extensions use more parts of it, I will have to look into it), should I leave that for a different PR or add it to this?

AriaMoradi:
@animeavi 
Join the discord server and also let us know when the code is ready to be reviewed

animeavi:
> @animeavi Join the discord server and also let us know when the code is ready to be reviewed

This part of it is ready, I'll do a follow up PR for the Rect and Canvas stuff cause it kinda depends on this being merged first.

AndroidCompat/src/main/java/android/graphics/Bitmap.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.imageio.IIOImage;
+import javax.imageio.ImageIO;
+import javax.imageio.ImageWriteParam;
+import javax.imageio.ImageWriter;
+import javax.imageio.stream.ImageOutputStream;
+
+public final class Bitmap {
+    private int width;
+    private int height;
+    private BufferedImage img;
```
AriaMoradi: Use full name variable names

AriaMoradi (changes requested):
We probably need to regenerate android.jar after this

AndroidCompat/src/main/java/android/graphics/Bitmap.java
animeavi: Done, changed a couple others in that class too.

PR #462, opened 2022-12-06
Get Playwright working

Syer10:
ShadowJar works using a custom DriverJar implementation, it sets the `playwright.driver.impl` system property so Playwright looks there before trying to use its own.

The only actual change between this and the original file is that it catches `FileSystemAlreadyExistsException` and uses null instead


Syer10:
Hmm, this seems to break the run gradle task, Ill see what I can do about that

Syer10:
Its all good now, works with both run and shadowJar 👍

AriaMoradi:
Let's see if they merge https://github.com/microsoft/playwright-java/pull/1140 upstream, if they refuse the patch then we'll merge this.

Syer10:
It may be better to just merge it and remove the fix when Playwright merges it and does a release, since it may take a while before its merged.

PR #479, opened 2023-01-07
Improve Playwright handling

Syer10:
- In bundlers pre-install Chromium
- Only ever download chromium, other browsers are skipped

### DIFF
--- a/scripts/bundler.sh
+++ b/scripts/bundler.sh
@@ -26,6 +26,8 @@ main() {
   set -- "${POSITIONAL_ARGS[@]}"
 
   OS="$1"
+  PLAYWRIGHT_VERSION="$(cat gradle/libs.versions.toml | grep -oP "playwright = \"\K([0-9\.]*)(?=\")")"
+  PLAYWRIGHT_REVISION="$(curl --silent "https://raw.githubusercontent.com/microsoft/playwright/v$PLAYWRIGHT_VERSION/packages/playwright-core/browsers.json" 2>&1 | grep -ozP "\"name\": \"chromium\",\n *\"revision\": \"\K[0-9]*")"
   JAR="$(ls server/build/*.jar | tail -n1)"
   RELEASE_NAME="$(echo "${JAR%.*}" | xargs basename)-$OS"
   RELEASE_VERSION="$(tmp="${JAR%-*}"; echo "${tmp##*-}" | tr -d v)"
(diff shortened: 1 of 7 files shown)


scripts/bundler.sh
```diff
@@ -57,6 +59,9 @@ main() {
       ELECTRON_URL="https://github.com/electron/electron/releases/download/$electron_version/$ELECTRON"
       download_jre_and_electron
 
+      PLAYWRIGHT_PLATFORM="linux"
```
AriaMoradi, line 62: add cpu architecture to the platform

scripts/bundler.sh
```diff
@@ -70,6 +75,9 @@ main() {
       ELECTRON_URL="https://github.com/electron/electron/releases/download/$electron_version/$ELECTRON"
       download_jre_and_electron
 
+      PLAYWRIGHT_PLATFORM="mac"
```
AriaMoradi, line 78: add cpu architecture to the platform

server/src/main/java/eu/kanade/tachiyomi/network/interceptor/DriverJar.java
```diff
@@ -26,22 +26,9 @@
...
-  >       return FileSystems.newFileSystem(uri, Collections.emptyMap());
-  >     } catch (FileSystemAlreadyExistsException e) {
-  >       return null;
-  >     }
-  >   }
-  116c124
-  <     try (FileSystem fileSystem = "jar".equals(uri.getScheme()) ? FileSystems.newFileSystem(uri, Collections.emptyMap()) : null) {
-  ---
-  >     try (FileSystem fileSystem = "jar".equals(uri.getScheme()) ? initFileSystem(uri) : null) {
+/**
```
AriaMoradi, line 29: can we use be a better approach? i.e. running `Chromium.preinstall` in serverSetup
I prefer to not have a custom DriverJar for playwright

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/Chromium.kt
```diff
@@ -0,0 +1,58 @@
+package eu.kanade.tachiyomi.network.interceptor
```
AriaMoradi: looks like we should put playwright code out of this package, its interfering with the rule of one interceptor per file.

AriaMoradi (changes requested):

server/src/main/java/eu/kanade/tachiyomi/network/interceptor/DriverJar.java
Syer10: We cannot, we need a custom DriverJar so that we make it only install chromium(line 96)

scripts/bundler.sh
Syer10: It has nothing to do with cpu architecture, it only has to do with the location of the download on Playwrights CDN

AriaMoradi (approved):

PR #491, opened 2023-02-08
Fixes Page index issues for some providers

akabhirav:
Fixes Page starting index issues for certain providers.

**Background**
Tachiyomi does not standardize the starting indexes for it's extensions. This has led to some extensions starting the Page indexs from 0 and others from 1. This causes the reader in Tachidesk to crash since the first api to get Page with index 0 fails


AriaMoradi:
@Syer10 any ideas for a faster/better approach?

Syer10:
This looks fine to me, since it does the ordering and offet using a sql query it should be fast

Syer10:
Though I took a look at Tachiyomi and it seems like Tachiyomi re-indexes the pages anyway before downloading them
https://github.com/tachiyomiorg/tachiyomi/blob/34a586ce48d6c10ac9eadb697b955ba336044bc2/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt#L326

PR #493, opened 2023-02-09
Decouples Cache and Download behaviour

akabhirav:
# Decoupling Cache and Download behaviour

## Separating cache directory from Download directory
Tachidesk's caching and download logic are tightly coupled. This includes the cache directory being the same as download directory. Ideally the directory for files meant to be temporarily stored as cache should be the operating system's temp directory. The first change deals with changing the cache directory to the OS' temp directory

## Implementing a new Downloader logic
Now that cached files are stored in another place, the downloader relying on cache logic to store the page(image) files no longer works. To fix this we need to make the downloader self-reliant which only asks the ImageRespone to be in image and then stores the file to the download folder on it's own

## Technicals
All of this introduces a few new changes in the server:
1. Introduced `DownloadedFilesProvider` as an abstract class which sets up the struture for a provider that is responsible for handling downloaded files i.e. getting, putting and deleting pages.
2. Introduced `FileProvider` as the default provider implementing `DownloadedFilesProvider` that implements the default behaviour of saving each page as a separate image in the chapter folder inside downloads directory.
3. Introduced `ChapterDownloadHelper` object that works as a `DownloadedFilesProvider` factory which generates the appropriate provider object. This abstracts the complexity that each provider must implement to store it's own logic for the above mentioned tasks.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
@@ -1,0 +1,24 @@
+package suwayomi.tachidesk.manga.impl
+
+import suwayomi.tachidesk.manga.impl.download.DownloadedFilesProvider
+import suwayomi.tachidesk.manga.impl.download.FolderProvider
+import java.io.InputStream
+
+object ChapterDownloadHelper {
+    fun getImage(mangaId: Int, chapterId: Int, index: Int): Pair<InputStream, String> {
+        return provider(mangaId, chapterId).getImage(index)
+    }
+
+    fun delete(mangaId: Int, chapterId: Int): Boolean {
+        return provider(mangaId, chapterId).delete()
+    }
+
+    fun putImage(mangaId: Int, chapterId: Int, index: Int, image: InputStream): Boolean {
+        return provider(mangaId, chapterId).putImage(index, image)
+    }
+
+    // return the appropriate provider based on how the download was saved. For the logic is simple but will evolve when new types of downloads are available
+    private fun provider(mangaId: Int, chapterId: Int): DownloadedFilesProvider {
+        return FolderProvider(mangaId, chapterId)
+    }
+}
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
@@ -5,7 +5,6 @@ import suwayomi.tachidesk.manga.impl.download.DownloadedFilesProvider
 import suwayomi.tachidesk.manga.impl.download.FolderProvider
 import suwayomi.tachidesk.manga.impl.download.model.DownloadChapter
 import java.io.InputStream
-import kotlin.reflect.KSuspendFunction2
 
 object ChapterDownloadHelper {
     fun getImage(mangaId: Int, chapterId: Int, index: Int): Pair<InputStream, String> {
@@ -21,7 +20,7 @@ object ChapterDownloadHelper {
         chapterId: Int,
         download: DownloadChapter,
         scope: CoroutineScope,
-        step: KSuspendFunction2<DownloadChapter?, Boolean, Unit>
+        step: suspend (DownloadChapter?, Boolean) -> Unit
     ): Boolean {
         return provider(mangaId, chapterId).download(download, scope, step)
     }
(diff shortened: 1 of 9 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
```diff
@@ -0,0 +1,24 @@
...
+object ChapterDownloadHelper {
+    fun getImage(mangaId: Int, chapterId: Int, index: Int): Pair<InputStream, String> {
+        return provider(mangaId, chapterId).getImage(index)
+    }
+
+    fun delete(mangaId: Int, chapterId: Int): Boolean {
+        return provider(mangaId, chapterId).delete()
+    }
+
+    fun putImage(mangaId: Int, chapterId: Int, index: Int, image: InputStream): Boolean {
```
akabhirav: going to change this to download since we don't need a single put image here, we can just move the entire downlod logic here. Also, building an ArchiveProvider is proving difficult since appending to zip is not working for some reason

akabhirav: One more point that I was thinking about is implementing a CacheProvider, seeing if this could help us completely move out any storage logic out of ImageResponse

akabhirav: Made this change in the latest commit

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
```diff
@@ -0,0 +1,33 @@
...
+    fun delete(mangaId: Int, chapterId: Int): Boolean {
+        return provider(mangaId, chapterId).delete()
+    }
+
+    suspend fun download(
+        mangaId: Int,
+        chapterId: Int,
+        download: DownloadChapter,
+        scope: CoroutineScope,
+        step: KSuspendFunction2<DownloadChapter?, Boolean, Unit>
```
Syer10:
```suggestion
        step: suspend (DownloadChapter?, Boolean) -> Unit
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadedFilesProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+/*
+* Base class for downloaded chapter files provider, example: Folder, Archive
+* */
+abstract class DownloadedFilesProvider(val mangaId: Int, val chapterId: Int) {
+    abstract fun getImage(index: Int): Pair<InputStream, String>
+
+    abstract suspend fun download(
+        download: DownloadChapter,
+        scope: CoroutineScope,
+        step: KSuspendFunction2<DownloadChapter?, Boolean, Unit>
```
Syer10:
```suggestion
        step: suspend (DownloadChapter?, Boolean) -> Unit
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/FolderProvider.kt
```diff
@@ -0,0 +1,93 @@
...
+        val file = folder.listFiles()?.get(index)
+        val fileType = file!!.name.substringAfterLast(".")
+        return Pair(FileInputStream(file).buffered(), "image/$fileType")
+    }
+
+    @OptIn(FlowPreview::class)
+    override suspend fun download(
+        download: DownloadChapter,
+        scope: CoroutineScope,
+        step: KSuspendFunction2<DownloadChapter?, Boolean, Unit>
```
Syer10:
```suggestion
        step: suspend (DownloadChapter?, Boolean) -> Unit
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/FolderProvider.kt
```diff
@@ -0,0 +1,93 @@
...
+                            step(null, false) // don't throw on canceled download here since we can't do anything
+                        }
+                        .launchIn(scope)
+                }.first
+
+                val filePath = "$chapterDir/$fileName"
+                ImageResponse.saveImage(filePath, image)
+                withContext(Dispatchers.IO) {
+                    image.close()
+                }
```
Syer10:
```suggestion
                Page.getPageImage(
                    mangaId = download.mangaId,
                    chapterIndex = download.chapterIndex,
                    index = pageNum
                ) { flow ->
                    pageProgressJob = flow
                        .sample(100)
                        .distinctUntilChanged()
                        .onEach {
                            download.progress = (pageNum.toFloat() + (it.toFloat() * 0.01f)) / pageCount
                            step(null, false) // don't throw on canceled download here since we can't do anything
                        }
                        .launchIn(scope)
                }.first.use { image ->
                    val filePath = "$chapterDir/$fileName"
                    ImageResponse.saveImage(filePath, image)
                }
```

Syer10 (changes requested):

akabhirav:
@Syer10 made the changes you've requested till now

PR #497, opened 2023-02-12
allow disabling playwright

AriaMoradi:


Syer10:
I think it would be better to just disable it on releases for now, either that or remove it from the .conf files. Since its planned to change later on, if people install Tachidesk this update they will always have webview disabled, even when we enable it later on.

Syer10 (commented):
I think it would be better to just disable it on releases for now, either that or remove it from the .conf files. Since its planned to change later on, if people install Tachidesk this update they will always have webview disabled, even when we enable it later on.

PR #508, opened 2023-02-20
Fix random page sent when book is downloaded

akabhirav:
# Fixes random pages being sent when a chapter is downloaded

## Steps to reproduce
1. In the master branch or release, download any chapter from any book while keeping download as cbz disabled
2. Go to read that downloaded chapter

### Expected behaviour
The pages should render in correct order

### Current behaviour
The pages are getting rendered randomly

The problem was the files were not being sorted before being read and sent. Now that's fixed


schroda:
fixes #528

AriaMoradi:
More like fix-ed we won't close that one as we'll won't have a future stable release of Tachidesk in a few months

PR #511, opened 2023-03-01
Feature/library update from source grouped by source

### LINKED ISSUE #383: [Bug] Taking too much time to update manga titles
## Device information
- Tachidesk version: Tachidesk-Server-v0.6.3-r1100-windows-x64
- Server Operating System: Windows 10
- Server JVM version: OpenJDK 8.0.3020.8
- Client Operating System: Windows 10
- Client Web Browser: Brave 1.42.86, also Electron

## Steps to reproduce
1. Click on "update"
2. Wait for manga to be updated

### Expected behavior
Update of 119 different titles should occur in a reasonable time. (Tachiyomi on my OnePlus Nord 2 5G takes around 2 minutes)

### Actual behavior
Running the debug version, I can see that Tachidesk waits over 5-7 seconds after polling each title to see if it has been updated, ends up taking around 20-25 minutes. This is not related to my internet speed or my PC's speed.

## Other details
None that I can think of.

schroda:
Update is currently done synchronously which can take a while.

With this change it's now done in parallel per source.
In case a library consists of mangas from many different sources, the max. parallel update requests are limited.

should fix #383 or at least improve it

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
@@ -14,9 +14,12 @@ import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Semaphore
+import kotlinx.coroutines.sync.withPermit
 import mu.KotlinLogging
 import suwayomi.tachidesk.manga.impl.Chapter
 import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
+import suwayomi.tachidesk.server.serverConfig
 import java.util.concurrent.ConcurrentHashMap
 
 class Updater : IUpdater {
@@ -29,6 +32,8 @@ class Updater : IUpdater {
     private val tracker = ConcurrentHashMap<Int, UpdateJob>()
     private val updateChannels = ConcurrentHashMap<String, Channel<UpdateJob>>()
 
+    private val semaphore = Semaphore(serverConfig.maxParallelUpdateRequests)
+
     private fun getOrCreateUpdateChannelFor(source: String): Channel<UpdateJob> {
         return updateChannels.getOrPut(source) {
             logger.debug { "getOrCreateUpdateChannelFor: created channel for $source - channels: ${updateChannels.size + 1}" }
@@ -40,12 +45,14 @@ class Updater : IUpdater {
         val channel = Channel<UpdateJob>(Channel.UNLIMITED)
         channel.consumeAsFlow()
             .onEach { job ->
-                _status.value = UpdateStatus(
-                    process(job),
-                    tracker.any { (_, job) ->
-                        job.status == JobStatus.PENDING || job.status == JobStatus.RUNNING
-                    }
-                )
+                semaphore.withPermit {
+                    _status.value = UpdateStatus(
+                        process(job),
+                        tracker.any { (_, job) ->
+                            job.status == JobStatus.PENDING || job.status == JobStatus.RUNNING
+                        }
+                    )
+                }
             }
             .catch { logger.error(it) { "Error during updates" } }
             .launchIn(scope)
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
@@ -27,7 +27,14 @@ class Updater : IUpdater {
     override val status = _status.asStateFlow()
 
     private val tracker = ConcurrentHashMap<Int, UpdateJob>()
-    private var updateChannel = createUpdateChannel()
+    private val updateChannels = ConcurrentHashMap<String, Channel<UpdateJob>>()
+
+    private fun getOrCreateUpdateChannelFor(source: String): Channel<UpdateJob> {
+        return updateChannels.getOrPut(source) {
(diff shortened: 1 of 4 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -14,31 +14,50 @@ import kotlinx.coroutines.flow.launchIn
...
+
+    private val semaphore = Semaphore(MAX_PARALLEL_UPDATE_REQUESTS)
+
+    private fun getOrCreateUpdateChannelFor(source: String): Channel<UpdateJob> {
+        if (!updateChannels.containsKey(source)) {
+            updateChannels[source] = createUpdateChannel()
+            logger.debug { "getOrCreateUpdateChannelFor: created channel for $source - channels: ${updateChannels.size}" }
+        }
+
+        return updateChannels[source]!!
```
Syer10:
```suggestion
        return updateChannels.getOrPut(source) {
            logger.debug { "getOrCreateUpdateChannelFor: created channel for $source - channels: ${updateChannels.size}" }
            createUpdateChannel()
        }
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -14,31 +14,50 @@ import kotlinx.coroutines.flow.launchIn
...
     private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
 
     private val _status = MutableStateFlow(UpdateStatus())
     override val status = _status.asStateFlow()
 
     private val tracker = ConcurrentHashMap<Int, UpdateJob>()
-    private var updateChannel = createUpdateChannel()
+    private val updateChannels = mutableMapOf<String, Channel<UpdateJob>>()
+
+    private val semaphore = Semaphore(MAX_PARALLEL_UPDATE_REQUESTS)
```
Syer10: Just put the value here, annotate it with a comment if you like. Magic numbers are not something we are worried about 😄

Syer10 (commented):
Small changes, but otherwise looks good

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -14,31 +14,50 @@ import kotlinx.coroutines.flow.launchIn
...
+
     private val logger = KotlinLogging.logger {}
     private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
 
     private val _status = MutableStateFlow(UpdateStatus())
     override val status = _status.asStateFlow()
 
     private val tracker = ConcurrentHashMap<Int, UpdateJob>()
-    private var updateChannel = createUpdateChannel()
+    private val updateChannels = mutableMapOf<String, Channel<UpdateJob>>()
```
Syer10: Should also probably use a ConcurrentMap so that we don't run into issues with multiple channel creation
```suggestion
    private val updateChannels = ConcurrentHashMap<String, Channel<UpdateJob>>()
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
schroda: done

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
schroda: removed, but it does feel wrong 😆

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
schroda: done

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -29,6 +31,9 @@ class Updater : IUpdater {
     private val tracker = ConcurrentHashMap<Int, UpdateJob>()
     private val updateChannels = ConcurrentHashMap<String, Channel<UpdateJob>>()
 
+    // max parallel update requests
+    private val semaphore = Semaphore(10)
```
Trung0246: Is it viable for this constant to be available for the config file?

schroda: I don't see why not, good point

schroda: done

server/src/main/resources/server-reference.conf
```diff
@@ -18,6 +18,9 @@ server.electronPath = ""
 server.downloadAsCbz = false
 server.downloadsPath = ""
 
+# updater
+server.maxParallelUpdateRequests = 10 # sets how many sources can be updated in parallel. updates are grouped by source and all mangas of a source are updated synchronously
```
AriaMoradi, line 22: Tachiyomi sets this to 5, what can a good default be? Why 10?

AriaMoradi (changes requested):

server/src/main/resources/server-reference.conf
schroda: I saw that as well, my guess was that it's because their running on smartphones.
I thought that this was not the case for the Tachidesk-Server so I increased it a bit, so there is no real reason for the 10, just that I thought 5 was a little too low.

I also don't really now what the average use case is, meaning how many different sources you have in your library.
E.g. I have only 5 different sources atm so the max parallel update requests would be 5 anyway.

But since it is now included in the config, I don't really care what the default value is, so if you or someone else prefers something else it can be changed

PR #520, opened 2023-03-12
Add logic to only update specific categories

schroda:
Makes it possible to only update specific categories.

In case a manga is in an excluded category it will be excluded even if it is also in an included category.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
@@ -13,10 +13,7 @@ import suwayomi.tachidesk.manga.impl.Chapter
 import suwayomi.tachidesk.manga.impl.update.IUpdater
 import suwayomi.tachidesk.manga.impl.update.UpdateStatus
 import suwayomi.tachidesk.manga.impl.update.UpdaterSocket
-import suwayomi.tachidesk.manga.model.dataclass.CategoryDataClass
-import suwayomi.tachidesk.manga.model.dataclass.MangaChapterDataClass
-import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
-import suwayomi.tachidesk.manga.model.dataclass.PaginatedList
+import suwayomi.tachidesk.manga.model.dataclass.*
 import suwayomi.tachidesk.server.JavalinSetup.future
 import suwayomi.tachidesk.server.util.formParam
 import suwayomi.tachidesk.server.util.handler
@@ -94,11 +91,21 @@ object UpdateController {
             updater.reset()
         }
 
-        val mangasToUpdate = categories
-                .flatMap { CategoryManga.getCategoryMangaList(it.id) }
-                .distinctBy { it.id }
-                .sortedWith(compareBy(String.CASE_INSENSITIVE_ORDER, MangaDataClass::title))
-                .filter { it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE }
+        val includeInUpdateStatusToCategoryMap = categories.groupBy { it.includeInUpdate }
+        val excludedCategories = includeInUpdateStatusToCategoryMap[IncludeInUpdate.EXCLUDE].orEmpty()
+        val includedCategories = includeInUpdateStatusToCategoryMap[IncludeInUpdate.INCLUDE].orEmpty()
+        val unsetCategories = includeInUpdateStatusToCategoryMap[IncludeInUpdate.UNSET].orEmpty()
+        val categoriesToUpdate = includedCategories.ifEmpty { unsetCategories }
+
+        logger.debug { "Updating categories: '${categoriesToUpdate.joinToString("', '") { it.name }}'" }
+
+        val categoriesToUpdateMangas = categoriesToUpdate
+            .flatMap { CategoryManga.getCategoryMangaList(it.id) }
+            .distinctBy { it.id }
+        val mangasToCategoriesMap = CategoryManga.getMangasCategories(categoriesToUpdateMangas.map { it.id })
+        val mangasToUpdate = categoriesToUpdateMangas
+            .filter { it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE }
+            .filter { !excludedCategories.any { category -> mangasToCategoriesMap[it.id]?.contains(category) == true } }
 
         // In case no manga gets updated and no update job was running before, the client would never receive an info about its update request
         if (mangasToUpdate.isEmpty()) {
@@ -106,7 +113,10 @@ object UpdateController {
             return
         }
 
-        updater.addMangasToQueue(mangasToUpdate)
+        updater.addMangasToQueue(
+            mangasToUpdate
+                .sortedWith(compareBy(String.CASE_INSENSITIVE_ORDER, MangaDataClass::title)),
+        )
     }
 
     fun categoryUpdateWS(ws: WsConfig) {
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/CategoryManga.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/CategoryManga.kt
@@ -116,4 +116,20 @@ object CategoryManga {
             }
         }
     }
+
+    fun getMangasCategories(mangaIDs: List<Int>): Map<Int, List<CategoryDataClass>> {
+        return buildMap {
+            transaction {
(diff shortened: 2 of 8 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
```diff
@@ -93,11 +93,14 @@ object UpdateController {
...
         }
+
+        val categoriesToExclude = categories.filter { !it.includeInUpdate }
         categories
+            .filter { it.includeInUpdate }
             .flatMap { CategoryManga.getCategoryMangaList(it.id) }
             .distinctBy { it.id }
             .sortedWith(compareBy(String.CASE_INSENSITIVE_ORDER, MangaDataClass::title))
-            .filter { it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE }
+            .filter { it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE && !categoriesToExclude.any { category -> CategoryManga.getMangaCategories(it.id).contains(category) } }
```
Syer10: This will create a massive amount of queries to the database, there would probably be a better way.

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
schroda: good point, didn't think about that

changed it to get the categories for all mangas of the categories that should get updated in one query (not sure if that can be optimized, my sql knowledge is rotting away and it took me way too long to get it working with exposed 😭)

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
```diff
@@ -93,11 +93,20 @@ object UpdateController {
...
+
+        val categoriesToExclude = categories.filter { !it.includeInUpdate }
+        val unfilteredMangasToUpdate = categories
+            .filter { it.includeInUpdate }
             .flatMap { CategoryManga.getCategoryMangaList(it.id) }
             .distinctBy { it.id }
+        val mangasToCategoriesMap = CategoryManga.getMangasCategories(unfilteredMangasToUpdate.map { it.id })
+        val mangasToUpdate = unfilteredMangasToUpdate.filter {
+            it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE &&
+                !categoriesToExclude.any { category -> mangasToCategoriesMap[it.id]!!.contains(category) }
```
Syer10: No need to enforce not null
```suggestion
                !categoriesToExclude.any { category -> mangasToCategoriesMap[it.id]?.contains(category) == true }
```

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
```diff
@@ -93,11 +90,29 @@ object UpdateController {
...
             .flatMap { CategoryManga.getCategoryMangaList(it.id) }
             .distinctBy { it.id }
+        val mangasToCategoriesMap = CategoryManga.getMangasCategories(categoriesToUpdateMangas.map { it.id })
+        val mangasToUpdate = categoriesToUpdateMangas
+            .filter {
+                val alwaysUpdate = it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE
+                val isMangaInExcludedCategory = excludedCategories.any { category -> mangasToCategoriesMap[it.id]?.contains(category) == true } ?: false
+
+                alwaysUpdate && !isMangaInExcludedCategory
+            }
```
Syer10:
```suggestion
        val mangasToUpdate = categoriesToUpdateMangas
            .filter { it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE }
            .filter {
                !excludedCategories.any { category -> mangasToCategoriesMap[it.id]?.contains(category) == true }
            }
```

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/CategoryDataClass.kt
```diff
@@ -1,16 +1,27 @@
...
+
 /*
  * Copyright (C) Contributors to the Suwayomi project
  *
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+enum class IncludeInUpdate(@JsonValue val value: Int) {
+    EXCLUDE(0), INCLUDE(1), UNSET(2);
```
Syer10: Why is UNSET 2 and not 0? I image 0 as a sane default

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/CategoryManga.kt
```diff
@@ -116,4 +116,21 @@ object CategoryManga {
...
+                .forEach {
+                    val mangaId = it.key.value
+                    val categories = it.value
+
+                    mangasToCategoriesMap[mangaId] = categories.map { category -> CategoryTable.toDataClass(category) }
+                }
+        }
+
+        return mangasToCategoriesMap
+    }
```
Syer10:
```suggestion
    fun getMangasCategories(mangaIDs: List<Int>): Map<Int, List<CategoryDataClass>> {
        return buildMap {
            transaction {
                CategoryMangaTable.innerJoin(CategoryTable)
                    .select { CategoryMangaTable.manga inList mangaIDs }
                    .groupBy { it[CategoryMangaTable.manga] }
                    .forEach {
                        val mangaId = it.key.value
                        val categories = it.value

                        set(mangaId, categories.map { category -> CategoryTable.toDataClass(category) })
                    }
            }
        }
    }
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/CategoryDataClass.kt
schroda: no real reason, I just went with boolean (0 = false/exclude, 1 = true/include) which left 2 for UNSET

AriaMoradi:
Fix merge conflicts

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/CategoryDataClass.kt
AriaMoradi: well if we are mapping to numbers, -1 would be better for UNSET as it's not a natural number.

AriaMoradi: But all things considered, as this is not c/c++ it doesn't matter which number we are mapping to.

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
```diff
@@ -93,11 +90,25 @@ object UpdateController {
...
         }
-        categories
+
+        val includeInUpdateStatusToCategoryMap = categories.groupBy { it.includeInUpdate }
+        val excludedCategories = includeInUpdateStatusToCategoryMap[IncludeInUpdate.EXCLUDE].orEmpty()
+        val includedCategories = includeInUpdateStatusToCategoryMap[IncludeInUpdate.INCLUDE].orEmpty()
+        val unsetCategories = includeInUpdateStatusToCategoryMap[IncludeInUpdate.UNSET].orEmpty()
+        val categoriesToUpdate = includedCategories.ifEmpty { unsetCategories }
+
+        logger.info { "Updating categories: '${categoriesToUpdate.joinToString("', '") { it.name }}'" }
```
AriaMoradi: make this logger.debug

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/CategoryDataClass.kt
schroda: I also thought about -1, I prefer that over having UNSET as 0

schroda: I changed it to -1

Syer10 (approved):

AriaMoradi:
Fix merge conflicts

schroda:
> Fix merge conflicts

done

PR #537, opened 2023-04-09
Feature/global update exclude mangas with state

schroda:
why a specific endpoint for global update settings:
 - settings are known unlike metadata set by the client
 - documented in swagger, otherwise, they would be "hidden" unless code gets checked
 - easier to use on client side

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/UpdateController.kt
@@ -18,7 +18,9 @@ import suwayomi.tachidesk.manga.model.dataclass.IncludeInUpdate
 import suwayomi.tachidesk.manga.model.dataclass.MangaChapterDataClass
 import suwayomi.tachidesk.manga.model.dataclass.MangaDataClass
 import suwayomi.tachidesk.manga.model.dataclass.PaginatedList
+import suwayomi.tachidesk.manga.model.table.MangaStatus
 import suwayomi.tachidesk.server.JavalinSetup.future
+import suwayomi.tachidesk.server.serverConfig
 import suwayomi.tachidesk.server.util.formParam
 import suwayomi.tachidesk.server.util.handler
 import suwayomi.tachidesk.server.util.pathParam
@@ -108,8 +110,13 @@ object UpdateController {
             .distinctBy { it.id }
         val mangasToCategoriesMap = CategoryManga.getMangasCategories(categoriesToUpdateMangas.map { it.id })
         val mangasToUpdate = categoriesToUpdateMangas
+            .asSequence()
             .filter { it.updateStrategy == UpdateStrategy.ALWAYS_UPDATE }
+            .filter { if (serverConfig.excludeUnreadChapters) { (it.unreadCount ?: 0L) == 0L } else true }
+            .filter { if (serverConfig.excludeNotStarted) { it.lastReadAt != null } else true }
+            .filter { if (serverConfig.excludeCompleted) { it.status != MangaStatus.COMPLETED.name } else true }
             .filter { !excludedCategories.any { category -> mangasToCategoriesMap[it.id]?.contains(category) == true } }
+            .toList()
 
         // In case no manga gets updated and no update job was running before, the client would never receive an info about its update request
         if (mangasToUpdate.isEmpty()) {
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -35,6 +35,9 @@ class ServerConfig(getConfig: () -> Config, moduleName: String = MODULE_NAME) : 
 
     // updater
     var maxParallelUpdateRequests: Int by overridableConfig
+    var excludeUnreadChapters: Boolean by overridableConfig
+    var excludeNotStarted: Boolean by overridableConfig
+    var excludeCompleted: Boolean by overridableConfig
 
     // Authentication
     var basicAuthEnabled: Boolean by overridableConfig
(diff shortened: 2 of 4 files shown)


server/src/main/kotlin/suwayomi/tachidesk/global/GlobalAPI.kt
```diff
@@ -22,6 +23,10 @@ object GlobalAPI {
         path("settings") {
             get("about", SettingsController.about)
             get("check-update", SettingsController.checkUpdate)
+            path("global-update") {
+                get("", GlobalUpdateController.getSettings)
+                patch("", GlobalUpdateController.modifySetting)
+            }
```
AriaMoradi: this is not global API, global API namespace was made to "general" app API that would not belong to either manga or anime. this used to make more sense when we had three namespaces
`suwayomi.tachidesk.global`, `suwayomi.tachidesk.manga`, `suwayomi.tachidesk.anime`

move this and friends to manga.

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0027_GlobalUpdateSettings.kt
```diff
@@ -0,0 +1,28 @@
...
+    val table = "GLOBALMETA"
+
+    override val sql = """
+        ALTER TABLE $table ADD COLUMN type varchar(256);
+        
+        UPDATE $table SET type = '${SettingType.META}';
+        
+        INSERT INTO $table (key, type, value)
+        VALUES
+            ${GlobalUpdateSettingsDataClass::class.memberProperties.joinToString(", ") { "('${it.name}', '${SettingType.GLOBAL_UPDATE}', 'true')" }};
```
AriaMoradi: this is confusing to read, hardcode the values or calculate them in a more verbose way

server/src/main/kotlin/suwayomi/tachidesk/global/impl/GlobalMeta.kt
```diff
@@ -14,7 +12,36 @@ import suwayomi.tachidesk.global.model.table.GlobalMetaTable
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-object GlobalMeta {
+enum class SettingType {
+    META,
+    GLOBAL_UPDATE,
+}
```
AriaMoradi: this file is intended for general meta where settings that would belong to the app itself or shared between anime and manga, move the new stuff to a new API

server/src/main/kotlin/suwayomi/tachidesk/global/model/table/GlobalMetaTable.kt
```diff
@@ -8,11 +8,27 @@ package suwayomi.tachidesk.global.model.table
...
+    when (DataClass::class) {
+        GlobalUpdateSettingsDataClass::class -> GlobalUpdateSettingsDataClass(
+            getSettingValue("excludeUnreadChapters").toBoolean(),
+            getSettingValue("excludeNotStarted").toBoolean(),
+            getSettingValue("excludeCompleted").toBoolean(),
+        ) as DataClass
+
+        else -> throw IllegalStateException("Unknown DataClass Type")
+    }
+}
```
AriaMoradi: these don't belong in this table as discussed earlier.
You can make a MangaMeta prefix for the new classes.

AriaMoradi (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/global/model/table/GlobalMetaTable.kt
schroda: overall in the GlobalMeta table (the database table) or not adding the toDataClass function to the table class?

AriaMoradi: In this class?
To tell the truth, the handling of settings is a mess. maybe we should introduce a settings manipulation API before this PR.

schroda: what do you have in mind?

server/src/main/kotlin/suwayomi/tachidesk/global/GlobalAPI.kt
schroda: moved to update endpoint (`.../update/settings`)

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0027_GlobalUpdateSettings.kt
schroda: changed

server/src/main/kotlin/suwayomi/tachidesk/global/impl/GlobalMeta.kt
schroda: moved

server/src/main/kotlin/suwayomi/tachidesk/global/impl/GlobalMeta.kt
```diff
@@ -1,11 +1,10 @@
 package suwayomi.tachidesk.global.impl
 
-import org.jetbrains.exposed.sql.insert
-import org.jetbrains.exposed.sql.select
-import org.jetbrains.exposed.sql.selectAll
+import org.jetbrains.exposed.sql.*
```
AriaMoradi: no wildcard imports

AriaMoradi (changes requested):

schroda:
moved settings to the server config.
with #545 merged, settings can now be modified during runtime (some require a restart).
the thing that is missing is to add a endpoint to modify the config (and handle settings that require a restart)

AriaMoradi (approved):

PR #542, opened 2023-04-30
Feature/move server settings to database

schroda:


server/src/main/kotlin/suwayomi/tachidesk/global/GlobalAPI.kt
```diff
@@ -20,6 +20,8 @@ object GlobalAPI {
             patch("", GlobalMetaController.modifyMeta)
         }
         path("settings") {
+            get("", SettingsController.serverSettings)
+            patch("", SettingsController.modifyServerSettings)
```
AriaMoradi, line 24: the URL seems to make sense as is but I'm thinking that it might become confusing to API users?

server/src/main/kotlin/suwayomi/tachidesk/server/SettingsTable.kt
```diff
@@ -0,0 +1,22 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import org.jetbrains.exposed.dao.id.IntIdTable
+import org.jetbrains.exposed.sql.ResultRow
+
+object SettingsTable : IntIdTable() {
+    val key = varchar("key", 256)
+    val value = varchar("value", 4096)
+    val requiresRestart = bool("requires_restart").default(false)
+}
```
AriaMoradi, line 17: It might be a better idea to use `java.util.prefs.Preferences` as storage backend instead?

- Using a database storage allows to easily make a portable Tachidesk installation (like on a usb stick)
- Using database will make you lose settings when the database gets corrupted or data is deleted

server/src/main/kotlin/suwayomi/tachidesk/server/Settings.kt
```diff
@@ -0,0 +1,233 @@
...
+            webUIInterface = PartialSetting(value = config.webUIInterface),
+            electronPath = PartialSetting(value = config.electronPath),
+            downloadAsCbz = PartialSetting(value = config.downloadAsCbz),
+            downloadsPath = PartialSetting(value = config.downloadsPath),
+            maxParallelUpdateRequests = PartialSetting(value = config.maxParallelUpdateRequests),
+            basicAuthEnabled = PartialSetting(value = config.basicAuthEnabled),
+            basicAuthUsername = PartialSetting(value = config.basicAuthUsername),
+            basicAuthPassword = PartialSetting(value = config.basicAuthPassword),
+            debugLogsEnabled = PartialSetting(value = config.debugLogsEnabled),
+            systemTrayEnabled = PartialSetting(value = config.systemTrayEnabled)
```
AriaMoradi, line 230: This list of values are repeated 8 times across the code base (2 times before + 6 times now), this is too much duplication.

### Some concerns
- We are cluttering the current package too much, put all settings stuff into a new sub-package
- the new setting module(s) seem too big and complected, I would split it apart heavily 
- We should completely eradicate all `com.typesafe.config` based code and only use a HOCON parser to read the config file (also there is an android config somewhere that needs to be dealt with) as we are building our own config engine now
- With this design we have multiple sources of truth: conf file and the database conf storage, this looks like an anti-pattern to me. No its not actually one source of truth and then an override, from user's view the conf file is the source of truth and from app's view it's the database. To fix this we might need to perform surgery and remove the conf file entirely or better than that make it the actual settings storage hence the source of truth which would fix the issue completely(also changing the format to toml or something along the way to make writing the conf file from software easy and accessible) 
  - As more rant about this I believe that the settings storage should be in plain text and accessible to the user, this is why we like technologies like csv, LaTeX, markdown more than their alternative binary forms. So we should either keep HOCON or switch to some other human readable format.

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -137,6 +137,12 @@ fun applicationSetup() {
 
     databaseUp()
 
+    Settings.init()
+
+    Settings.setup()
+
+    Settings.update(Settings.fromConfig(serverConfig))
```
AriaMoradi, line 144: This seems too complicated, we need one entry point, the steps here are a blackbox to `ServerSetup`

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfigDb.kt
```diff
@@ -0,0 +1,80 @@
...
+class DatabaseDelegator {
+    operator fun <R, SettingType, SETTING : Setting<SettingType>> setValue(
+        thisRef: R,
+        property: KProperty<*>,
+        setting: SETTING
+    ) {
+        Settings.put(property.name, setting)
+    }
+
+    inline operator fun <R, reified SETTING> getValue(thisRef: R, property: KProperty<*>): SETTING {
```
AriaMoradi, line 26: Not everyone is familiar with KProperty, maybe some comments with reference to external docs would be helpful

server/src/main/kotlin/suwayomi/tachidesk/global/GlobalAPI.kt
schroda: probably not possible since it would change an existing endpoint (on the other hand the frontend could quickly be updated), but would it make sense to rename "settings" to "server" and then add a "settings" sub endpoint?
at least I wouldn't consider "about" and "check-update" to be settings

e.g.:
 - server
   - settings
   - about
   - check-update

server/src/main/kotlin/suwayomi/tachidesk/server/Settings.kt
schroda: the idea would be to completely remove the old config file and only use the settings from the database.
the settings would then only be changeable via the UI (and via a launcher, in case that is still planned?)

regarding the code duplication, this is so that there is a partial data class of the settings, so that via the endpoint a single setting can be changed, without having to send all settings
not sure if this can be done in a better way, haven't found anything besides this

so regarding the backend of the settings, I understand you want to keep a config file?
if java.util.prefs.Preferences should be used, if I understand it correctly, we have to implement that ourself, since by default it will be saved in the registry

server/src/main/kotlin/suwayomi/tachidesk/global/GlobalAPI.kt
AriaMoradi: Not a good idea, unnecessary breaking changes are pointless. There might be other private Tachidesk forks and clients we don't know about.

server/src/main/kotlin/suwayomi/tachidesk/server/Settings.kt
AriaMoradi: > the idea would be to completely remove the old config file and only use the settings from the database.
the settings would then only be changeable via the UI (and via a launcher, in case that is still planned?)

I don't really like that idea because we have a lot of headless (docker, NAS, etc) users that prefer the sys admin conventional way, we also service home users, we really want to make things easier and more simple stupid for them (this has been the driving force behind most of my decisions) but we shouldn't turn our back to those lovely power users. 

> regarding the code duplication, this is so that there is a partial data class of the settings, so that via the endpoint a single setting can be changed, without having to send all settings
not sure if this can be done in a better way, haven't found anything besides this

@Syer10 any ideas?

I could try to come up with something when I have the time.

> so regarding the backend of the settings, I understand you want to keep a config file?

Yes, as explained before.

> if java.util.prefs.Preferences should be used, if I understand it correctly, we have to implement that ourself, since by default it will be saved in the registry

About that one I was just throwing the an idea, now that I think about it it might not be the best idea? 
- Java prefs are something that is kept hidden away and it's not an industry standard (besides in android) 
- Keeping the config in cleartext and accessible is almost a must, even though one of our regulars on discord likes to play with the database file but that would be annoying to anyone else.

schroda:
new approach https://github.com/Suwayomi/Tachidesk-Server/pull/545

PR #546, opened 2023-05-07
feat: add dockerfile using linuxserver apline iamge with option to change PUID and PGID

taos15:
- Make multi stage dockerfile that build the application and then create a container using an alpine image.

- Use linuxserver base alpine imagine with S6 overlays

- You can change the PUID and PGID that the aplication will run as

- It update the permissions of the data folder to the PUID and PGID passed with the docker run command


AriaMoradi:
Thanks for the PR but as we won't accept any docker stuff in this repo, I redirect you to https://github.com/Suwayomi/docker-tachidesk 

You can either rework your PR there or ask for a new repo on the Suwayomi org if there is enough justification.

PR #552, opened 2023-05-24
Category mangas

Syer10:


AriaMoradi:
Just to clarify this again, my reasoning for this is using mangas where manga should be plural just for extra verbosity and consistency with the rest of the codebase.

PR #573, opened 2023-06-09
Feature/update default category handling

schroda:
**should be merged after #574 and has to be updated before merge due to database migration with the same ID**

 - always return the "default" category
   - to be able to change the order
   - to be able to change settings (without additional logic needed on UI side)
     - "isDefault" flag
     - "includeInUpdate" flag
 - handle "default" category mapping in "CategoryManga" table
   - simplifies "default" category handling
   - makes it possible to have mangas in "default" and other categories simultaneously


Syer10:
I feel a lot of these features shouldn't be added. The Default category is meant to be a fallback category, not a configurable category. If you want a configurable category you should make one yourself, as the default category only shows when there are manga not in other categories.

I am fine with these features:
> to be able to change settings (without additional logic needed on UI side)
>     "isDefault" flag
>     "includeInUpdate" flag

But these should be removed:
> always return the "default" category 
> to be able to change the order
> handle "default" category mapping in "CategoryManga" table
>     simplifies "default" category handling
>     makes it possible to have mangas in "default" and other categories simultaneously

schroda:
lets pretend like this pr never existed 👀

PR #574, opened 2023-06-09
Do not count mangas as part of categories that aren't in the library

schroda:
Otherwise, the "size" calculation of categories is wrong, since it includes mangas that aren't in the library anymore and won't be returned in the list of mangas of a category


AriaMoradi:
This is tachiyomi behavior that we are reproducing. We should try working around this, unless there is a good reason to not to.
(I believe this behavior was originally intended to keep your categories for when adding manga to library again)

cc: @Syer10

Syer10:
Yes this is intenional behavior, I would suggest modifying the size calculation to only include in-library manga instead.

schroda:
changed

but what is the use case of this behaviour - that the categories are "restored" when adding it back to the library?

Syer10:
Yes, instead of using the default categories, it should use the previously selected categories

PR #581, opened 2023-06-19
Feature/decouple thumbnail downloads and cache

schroda:
Currently all thumbnails are downloaded permanently.
This not only includes thumbnails of mangas that are part of the library but also when browsing a source.
Unless you manually cleanup these files, they will never get deleted.

This PR changes thumbnails to be tmp files, which get saved in the systems tmp folder.
Only thumbnails of mangas, that are part of the library, will be saved permanently.
Once a manga gets added/removed from the library, the permanent thumbnail will get downloaded/deleted.

The logic for this reuses the existing logic of the chapter pages.
There had to be slight adaptions to the interface and classes since the functions for the thumbnails and pages have different number of arguments.
Since they are still classes which have to provide the same behaviour with the same functions (e.g. get image, download image, delete image) I wanted to use the same interface

In case the tmp files should not get saved in the systems tmp folder, this can be changed in a later/separate PR which implements a custom cache (which IMO then should also be used for the pages of chapters)

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/MangaMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/MangaMutation.kt
@@ -8,6 +8,7 @@ import org.jetbrains.exposed.sql.transactions.transaction
 import org.jetbrains.exposed.sql.update
 import suwayomi.tachidesk.graphql.types.MangaMetaType
 import suwayomi.tachidesk.graphql.types.MangaType
+import suwayomi.tachidesk.manga.impl.Library
 import suwayomi.tachidesk.manga.impl.Manga
 import suwayomi.tachidesk.manga.model.table.MangaMetaTable
 import suwayomi.tachidesk.manga.model.table.MangaTable
@@ -44,7 +45,7 @@ class MangaMutation {
         val patch: UpdateMangaPatch
     )
 
-    private fun updateMangas(ids: List<Int>, patch: UpdateMangaPatch) {
+    private suspend fun updateMangas(ids: List<Int>, patch: UpdateMangaPatch) {
         transaction {
             if (patch.inLibrary != null) {
                 MangaTable.update({ MangaTable.id inList ids }) { update ->
@@ -53,37 +54,47 @@ class MangaMutation {
                     }
                 }
             }
+        }.apply {
+            if (patch.inLibrary != null) {
+                ids.forEach {
+                    Library.handleMangaThumbnail(it, patch.inLibrary)
+                }
+            }
         }
     }
 
-    fun updateManga(input: UpdateMangaInput): UpdateMangaPayload {
+    fun updateManga(input: UpdateMangaInput): CompletableFuture<UpdateMangaPayload> {
         val (clientMutationId, id, patch) = input
 
-        updateMangas(listOf(id), patch)
+        return future {
+            updateMangas(listOf(id), patch)
+        }.thenApply {
+            val manga = transaction {
+                MangaType(MangaTable.select { MangaTable.id eq id }.first())
+            }
 
-        val manga = transaction {
-            MangaType(MangaTable.select { MangaTable.id eq id }.first())
+            UpdateMangaPayload(
+                clientMutationId = clientMutationId,
+                manga = manga
+            )
         }
-
-        return UpdateMangaPayload(
-            clientMutationId = clientMutationId,
-            manga = manga
-        )
     }
 
-    fun updateMangas(input: UpdateMangasInput): UpdateMangasPayload {
+    fun updateMangas(input: UpdateMangasInput): CompletableFuture<UpdateMangasPayload> {
         val (clientMutationId, ids, patch) = input
 
-        updateMangas(ids, patch)
+        return future {
+            updateMangas(ids, patch)
+        }.thenApply {
+            val mangas = transaction {
+                MangaTable.select { MangaTable.id inList ids }.map { MangaType(it) }
+            }
 
-        val mangas = transaction {
-            MangaTable.select { MangaTable.id inList ids }.map { MangaType(it) }
+            UpdateMangasPayload(
+                clientMutationId = clientMutationId,
+                mangas = mangas
+            )
         }
-
-        return UpdateMangasPayload(
-            clientMutationId = clientMutationId,
-            mangas = mangas
-        )
     }
 
     data class FetchMangaInput(
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
@@ -235,7 +235,7 @@ object Manga {
     private val network: NetworkHelper by injectLazy()
 
     suspend fun fetchMangaThumbnail(mangaId: Int): Pair<InputStream, String> {
-        val cacheSaveDir = applicationDirs.thumbnailsRoot
+        val cacheSaveDir = applicationDirs.tempThumbnailCacheRoot
         val fileName = mangaId.toString()
 
         val mangaEntry = transaction { MangaTable.select { MangaTable.id eq mangaId }.first() }
@@ -302,7 +302,7 @@ object Manga {
     }
 
     private fun clearMangaThumbnailCache(mangaId: Int) {
-        val saveDir = applicationDirs.thumbnailsRoot
+        val saveDir = applicationDirs.tempThumbnailCacheRoot
         val fileName = mangaId.toString()
 
         clearCachedImage(saveDir, fileName)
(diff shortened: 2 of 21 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/storage/ImageResponse.kt
```diff
@@ -48,13 +56,9 @@ object ImageResponse {
         val cachedFile = findFileNameStartingWith(saveDir, fileName)
         val filePath = "$saveDir/$fileName"
 
-        // in case the cached file is a ".tmp" file something went wrong with the previous download, and it has to be downloaded again
+// in case the cached file is a ".tmp" file something went wrong with the previous download, and it has to be downloaded again
```
AriaMoradi: fix indentation

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -331,13 +331,24 @@ object Chapter {
         }
     }
 
+    private fun deleteFilesOf(chapterIds: List<Int>, mangaId: Int) {
```
AriaMoradi: choose a better name for this function

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/IDownloadedFilesProvider.kt
```diff
@@ -0,0 +1,5 @@
+package suwayomi.tachidesk.manga.impl.download.fileProvider
+
+interface IDownloadedFilesProvider : IFileDownloader, IFileRetriever {
```
AriaMoradi: there is no need to include "I" in front of interface names

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/IFileDownloader.kt
```diff
@@ -0,0 +1,29 @@
+package suwayomi.tachidesk.manga.impl.download.fileProvider
+
+@FunctionalInterface
+interface IFileDownload {
+    suspend fun executeDownload(vararg args: Any): Boolean
+}
+
+@FunctionalInterface
+interface IFileDownload1<A> : IFileDownload {
```
AriaMoradi: make up better names instead of numbering

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/DirName.kt
```diff
@@ -44,8 +44,12 @@ private fun getChapterDir(mangaId: Int, chapterId: Int): String {
     return getMangaDir(mangaId) + "/$chapterDir"
 }
 
+fun getThumbnailDownloadPath(): String {
+    return applicationDirs.downloadsRoot + "/thumbnails"
+}
+
 fun getChapterDownloadPath(mangaId: Int, chapterId: Int): String {
-    return applicationDirs.mangaDownloadsRoot + "/" + getChapterDir(mangaId, chapterId)
+    return applicationDirs.downloadsRoot + "/mangas/" + getChapterDir(mangaId, chapterId)
```
AriaMoradi: bad abstraction here

AriaMoradi (dismissed):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
schroda: renamed it to `deleteDownloadsFor`

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/IFileDownloader.kt
schroda: tbh, no clue how to name this, I added "Args" as an suffix
I'm open for better names

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/DirName.kt
schroda: added `mangaDownloadsRoot` to app dirs

AriaMoradi: also we should use a sensible path joiner function instead of hand concatenating them

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/MangaMutation.kt
```diff
@@ -71,7 +82,7 @@ class MangaMutation {
         )
     }
 
-    fun updateMangas(input: UpdateMangasInput): UpdateMangasPayload {
+    suspend fun updateMangas(input: UpdateMangasInput): UpdateMangasPayload {
```
Syer10: Public functions in the `mutations` and `queries` packages cannot be suspend, but they can return a `CompleteableFuture` created by the `future {}` function

schroda: fixed

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/MangaMutation.kt
```diff
@@ -44,46 +45,60 @@ class MangaMutation {
         val patch: UpdateMangaPatch
     )
 
-    private fun updateMangas(ids: List<Int>, patch: UpdateMangaPatch) {
+    private suspend fun updateMangas(ids: List<Int>, patch: UpdateMangaPatch) {
```
Syer10, line 48: Can you revert most of the changes here? I had it like this so that we could add more user fields with minimal changes, just do another check for inLibrary when choosing what to do with the thumbnail

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/storage/SafePath.kt
```diff
@@ -9,6 +11,10 @@ package suwayomi.tachidesk.manga.impl.util.storage
 
 // adopted from: https://github.com/tachiyomiorg/tachiyomi/blob/4cefbce7c34e724b409b6ba127f3c6c5c346ad8d/app/src/main/java/eu/kanade/tachiyomi/util/storage/DiskUtil.kt
 object SafePath {
+    fun joinPaths(vararg paths: String): String {
```
Syer10: I'm not too sure about this function, I would prefer just doing something like this instead of using a function

```kotlin
val path = Path(System.getProperty("java.io.tmpdir")) / "Tachidesk"
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/storage/SafePath.kt
schroda: so, just leave it as it was? https://github.com/Suwayomi/Tachidesk-Server/pull/581#discussion_r1240983810

Syer10: Path.div is an operator function for Path objects

schroda: going a step back, does this actually matter (regarding the sensible path joiner)? won't it be correct either way once the path is used to create a File or Path?

Syer10: Nope, it doesn't matter. I just found the joinPaths function weird, if you want to revert those changes and just use / that's fine, since / is supported by every platform

schroda: I reverted it

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/MangaMutation.kt
schroda: done

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
```diff
@@ -284,8 +286,23 @@ object Manga {
...
+                ThumbnailDownloadHelper.getImage(mangaId)
+            }
+        }
+
+        return fetchMangaThumbnail(mangaId)
+    }
+
     private fun clearMangaThumbnailCache(mangaId: Int) {
-        val saveDir = applicationDirs.thumbnailsRoot
+        val saveDir = applicationDirs.tempThumbnailCacheRoot
```
Syer10: We call this when the thumbnail url was changed, so we need to download the new cover, is this properly handled for library manga?

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
schroda: it was not - updated

Syer10 (approved):

PR #591, opened 2023-07-01
Feature/move server frontend mapping to the frontend

schroda:
Currently the webUI requires a server release.
This PR enables the webUI to do its own releases by adding the functionality to the server to download the latest webUI version.
The webUI defines which webUI version is compatible with which server version.
Thus, the server can download the latest compatible version based on its version.

The mapping of webUI version to server version is a file inside the webUI repo.
This file gets updated each time a new webUI version gets released which requires a newer server version than the previous release.

example:

 - ui: PREVIEW; server: x
 - ui: v5; server: v3
 - ui: v2; server: v2
 
In this example
 - the current preview version requires a newer server version than v3.
 - each webUI version from ("[": =>; "]": <=; "(": >; ")": <)
   - [v5, PREVIEW) is compatible with server versions [v3, x)
   - [v2, v5) is compatible with server versions [v2, v3)
  
This does not change the current behaviour of the server release bundling a webUI version.
This version is still used as a fallback in case everything else fails.

How the server will setup the webUI:
 - local webUI is available?
   - available: local version is valid?
     - valid: check for update
       - update available?
         - available: download latest version if auto update is enabled
         - up-to-date: -
      - invalid: _download latest version_
        - _download result handling_
   - missing: _download latest version_
     - _download result handling_

_download result handling_:
- download failed?
  - yes: is default webUI
    - yes: fallback to bundled version
    - no: fallback to default webUI and retry _download latest version_
  
Support for different webUIs is also added.
As long as they fullfill the required conditions, e.g. version mapping file in correct format, available version downloads, they can be defined in the server to be provided as an optional webUI that can be used.
In case the server runs into issues with one of these webUIs (e.g. download not possible) it will fallback to the default webUI

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/util/WebInterfaceManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/util/WebInterfaceManager.kt
@@ -17,12 +17,12 @@ import org.json.JSONArray
 import org.kodein.di.DI
 import org.kodein.di.conf.global
 import org.kodein.di.instance
-import suwayomi.tachidesk.manga.impl.update.Updater
 import suwayomi.tachidesk.server.ApplicationDirs
 import suwayomi.tachidesk.server.BuildConfig
 import suwayomi.tachidesk.server.serverConfig
 import suwayomi.tachidesk.util.HAScheduler
 import java.io.File
+import java.io.InputStream
 import java.net.HttpURLConnection
 import java.net.URL
 import java.nio.charset.StandardCharsets
@@ -37,7 +37,7 @@ private val tmpDir = System.getProperty("java.io.tmpdir")
 private fun ByteArray.toHex(): String = joinToString(separator = "") { eachByte -> "%02x".format(eachByte) }
 
 enum class WebUIChannel {
-    BUNDLED, // the version bundled with the server release
+    BUNDLED, // the default webUI version bundled with the server release
     STABLE,
     PREVIEW;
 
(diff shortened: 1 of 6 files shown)


server/src/main/kotlin/suwayomi/tachidesk/server/util/WebInterfaceManager.kt
```diff
@@ -7,130 +7,371 @@ package suwayomi.tachidesk.server.util
...
+        val isDownloadValid = isLocalWebUIValid(tempUnzippedWebUIFolderPath)
+
+        File(tempUnzippedWebUIFolderPath).deleteRecursively()
+
+        return isDownloadValid
+    }
+
+    private fun extractDownload(zipFilePath: String, targetPath: String) {
+        File(targetPath).mkdirs()
+        ZipFile(zipFilePath).extractAll(targetPath)
```
Syer10: I think this is a Closable, so I believe the code should be
```suggestion
        ZipFile(zipFilePath).use { it.extractAll(targetPath) }
```

Syer10 (changes requested):

Syer10 (approved):
@schroda Is this ready to be merged? WebUI has the mapping?

schroda:
> @schroda Is this ready to be merged? WebUI has the mapping?

not yet, I'll ping you when it's ready

---

I updated the automated webUI update check to use the "HAScheduler"

schroda:
I added a function to get the latest version of the webUI preview, instead of having to change the naming pattern of the bundled webUI zip file from `Tachidesk-WebUI-<revision>.zip` to `Tachidesk-WebUI.zip`

schroda:
webUI repo is up-to-date

PR #593, opened 2023-07-02
Feature/global update trigger automatically

schroda:


Syer10 (changes requested):
The update interval needs to have a reasonable minimum, preferably similar to Tachi's 12 hours

schroda:
only every 12 hours? 
64d1xe


what about 6? - you could/can trigger it manually without any limitations anyway

Syer10 (approved):

PR #596, opened 2023-07-10
Feature/automatically download new chapters

schroda:


server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt
```diff
@@ -120,19 +122,23 @@ class ChapterMutation {
         input: FetchChaptersInput
     ): CompletableFuture<FetchChaptersPayload> {
         val (clientMutationId, mangaId) = input
+        var currentChapters: List<ChapterDataClass> = emptyList()
 
         return future {
+            currentChapters = Chapter.getChapterList(input.mangaId)
             Chapter.fetchChapterList(mangaId)
         }.thenApply {
```
Syer10:
```suggestion

        return future {
            val currentChapters = Chapter.getChapterList(input.mangaId)
            Chapter.fetchChapterList(mangaId)
            currentChapters
        }.thenApply { currentChapters ->
```

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt
```diff
@@ -120,19 +122,23 @@ class ChapterMutation {
...
             val chapters = transaction {
                 ChapterTable.select { ChapterTable.manga eq mangaId }
                     .orderBy(ChapterTable.sourceOrder)
-                    .map { ChapterType(it) }
             }
 
+            // download new chapters if settings flag is enabled
+            Chapter.downloadNewChapters(mangaId, currentChapters.size, chapters.map { ChapterTable.toDataClass(it) })
```
Syer10: I don't like that we are converting it to a ChapterDataClass, is there a way to simplify it?

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt
schroda: can't think of anything, maybe you have an idea?

if I want to pass the chapter list, it has to get mapped to something, so either ChapterDataClass or ChapterType, meaning that either here or for the old api the the list has to get mapped to the other structure.
or a complete new structure just for this function which only contains the id, read and download property, but that seems to be a overhead to me?

I guess instead of passing the list the size of it could be passed, but I decided against it to avoid another read from the database, especially for large lists, since the list is already read from the database by the time the function gets called

Syer10: Could we pass the ResultRow directly? Either that or map them to a common class that only contains the data needed for the function. I want to maximize the performance as much as possible

schroda: I think passing ResultRow would be the best solution

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt
```diff
@@ -122,17 +124,21 @@ class ChapterMutation {
         val (clientMutationId, mangaId) = input
 
         return future {
+            val currentChapters = Chapter.getChapterList(input.mangaId)
```
Syer10: I think you could change this for a database query that just gets the chapter list size now that I think about it

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -55,6 +62,15 @@ object Chapter {
         }
     }
 
+    fun getCountOfMangaChapters(mangaId: Int): Int {
+        return transaction {
+            ChapterTable.select { ChapterTable.manga eq mangaId }
+                .orderBy(ChapterTable.sourceOrder to SortOrder.DESC)
+                .count()
+                .toInt()
```
Syer10:
```suggestion
            ChapterTable.select { ChapterTable.manga eq mangaId }.count().toInt()
```

schroda: just noticed it myself, changed

Syer10 (approved):

PR #598, opened 2023-07-12
Filter out mangas of category that are not in the library

schroda:


Syer10:
I'm 50/50 on this, I don't want to interfere directly with the relational data, but I can see how this would be useful.

schroda:
> I'm 50/50 on this, I don't want to interfere directly with the relational data, but I can see how this would be useful.

Ideally it would be possible to apply condtions/filters/pagination onto "child" nodes as it's possible for the "base" node (no clue how that is actually called)

But while this is not possible (if at all?) I'd say it should be aligned to the old api which only returned library mangas.
At least in the webUI there is an option to display the total number of mangas in the library and categories, which would currently show wrong numbers, unless all mangas of all categories get requested and then filtered locally

PR #599, opened 2023-07-13
Preserve download queue through server restarts

schroda:


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -47,6 +54,42 @@ object DownloadManager {
...
+            }
+
+            queueStorageFile.readText().split(",").filter { it.isNotEmpty() }.map { it.toInt() }
+        } catch (e: IOException) {
+            e.printStackTrace()
+            emptyList()
+        }
+    }
+
+    private fun saveDownloadQueue() {
```
Syer10, line 65: I think it would be better if we stored it in a SharedPreferences object like we store cookies

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
schroda: I actually saved it at first in preferences but the size of the element go too big and threw an error.
This was still before I reduced the data down to the chapterIds so it should not be a problem anymore.

I still decided to use a file (database before but the required logic would get more complex) instead since the queue has no size limit, thus, I'm not sure if it's possible to get too big?
According to the documentation the max size of a preference element is 8192 chars, so in case this gets exceeded, logic would be needed to split the queue into multiple preference elements

If we assume the average char length of a chapterId is 5 and add the necessary comma for separation, that would make it ~6 chars per chapter.
So it would be possible to store up to ~1365 chapters in a single preference element

at the end this would only be an issue for people that download a lot but still has to be considered and handled

Syer10: I was thinking of storing it using the StringSet preference, the StringSet is configured so that it serializes each item as a separate preference, which should work with this,

schroda: updated

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -47,6 +54,38 @@ object DownloadManager {
     private val downloadQueue = CopyOnWriteArrayList<DownloadChapter>()
     private val downloaders = ConcurrentHashMap<Long, Downloader>()
 
+    private const val downloadQueueKey = "downloadQueueKey"
+    private val sharedPreferences =
+        Injekt.get<Application>().getSharedPreferences(DownloadManager::class.jvmName, Context.MODE_PRIVATE)
+
+    private fun loadDownloadQueue(): List<Int> {
+        return try {
+            sharedPreferences.getStringSet(downloadQueueKey, emptySet())?.map { it.toInt() } ?: emptyList()
```
Syer10:
```suggestion
            sharedPreferences.getStringSet(downloadQueueKey, emptySet())?.mapNotNull { it.toIntOrNull() } ?: emptyList()
```
Just to be safe.
Also I don't think you need the try catch statements anymore

schroda: oh yeah, forgot to remove them

Syer10 (approved):

PR #609, opened 2023-07-23
Library Update Queries and Mutations

Syer10:


server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/UpdateMutation.kt
```diff
@@ -0,0 +1,69 @@
...
+import suwayomi.tachidesk.graphql.types.UpdateStatus
+import suwayomi.tachidesk.manga.impl.Category
+import suwayomi.tachidesk.manga.impl.update.IUpdater
+import suwayomi.tachidesk.manga.model.table.CategoryTable
+import suwayomi.tachidesk.manga.model.table.toDataClass
+
+class UpdateMutation {
+    private val updater by DI.global.instance<IUpdater>()
+
+    data class UpdateGlobalMangaInput(
```
schroda: what about `UpdateLibraryManga...`?

schroda: would it also be better to use `mangas` instead of `manga` to make it clear that it's plural?

Syer10: Thanks

Syer10: Manga rolls off the tongue better, its not like we need to be specific here

schroda (approved):

PR #610, opened 2023-07-23
Downloader Queries and Mutations

Syer10:


server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
```diff
@@ -25,7 +25,7 @@ class DownloadMutation {
         val chapters: List<ChapterType>
     )
 
-    fun deleteChapters(input: DeleteChapterDownloadsInput): DeleteChapterDownloadsPayload {
+    fun deleteChapterDownloads(input: DeleteChapterDownloadsInput): DeleteChapterDownloadsPayload {
```
schroda: what about `deleteDownloadedChapter(s)`, `DeleteDownloadedChapter(s)Input` and `DeleteDownloadedChapter(s)Payload`?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
```diff
@@ -0,0 +1,260 @@
...
+            EnqueueChapterDownloadPayload(
+                clientMutationId = clientMutationId,
+                downloadStatus = withTimeout(30.seconds) {
+                    DownloadStatus(DownloadManager.status.first { it.queue.any { it.chapter.id == chapter } })
+                }
+            )
+        }
+    }
+
+    data class UnqueueChapterDownloadsInput(
```
schroda: I saw this in the DownloadManager as well, but wouldn't it be `dequeue` instead of `unqueue`?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
Syer10: Good idea

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
Syer10: Thanks

schroda (approved):

PR #617, opened 2023-07-28
Feature/listen to server config value changes

schroda:


Syer10:
Nice to see the Tachidesk-Launcher code being used here 😄

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -8,58 +8,124 @@ package suwayomi.tachidesk.server
...
+    val localSourcePath: MutableStateFlow<String> by OverrideConfigValue(StringConfigAdapter)
+
+    fun <T> subscribeTo(flow: Flow<T>, onChange: suspend (value: T) -> Unit, ignoreInitialValue: Boolean = true) {
+        val actualFlow = if (ignoreInitialValue) {
+            flow.drop(1)
+        } else {
+            flow
+        }
+
+        actualFlow.distinctUntilChanged().onEach { onChange(it) }.conflate().launchIn(mutableConfigValueScope)
```
Syer10: I don't think you need to Conflate here

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -161,25 +171,40 @@ fun applicationSetup() {
...
     // create system tray
-    if (serverConfig.systemTrayEnabled) {
+    serverConfig.subscribeTo(serverConfig.systemTrayEnabled, { systemTrayEnabled ->
         try {
-            systemTrayInstance
+            if (systemTrayEnabled) {
+                systemTrayInstance
+            } else {
+                systemTrayInstance?.remove()
+            }
```
Syer10, line 180: I believe this will break if we disable system tray and then try to enable it again.

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -161,25 +171,40 @@ fun applicationSetup() {
...
+            serverConfig.socksProxyPort
+        ) { proxyEnabled, proxyHost, proxyPort ->
+            Triple(proxyEnabled, proxyHost, proxyPort)
+        },
+        { (proxyEnabled, proxyHost, proxyPort) ->
+            if (proxyEnabled) {
+                System.getProperties()["socksProxyHost"] = proxyHost
+                System.getProperties()["socksProxyPort"] = proxyPort
+                logger.info("Socks Proxy is enabled to $proxyHost:$proxyPort")
+            }
```
Syer10, line 208: Does this actually work?

server/src/test/kotlin/suwayomi/tachidesk/test/ApplicationTest.kt
```diff
@@ -138,12 +139,22 @@ open class ApplicationTest {
...
+                    Triple(proxyEnabled, proxyHost, proxyPort)
+                },
+                { (proxyEnabled, proxyHost, proxyPort) ->
+                    if (proxyEnabled) {
+                        System.getProperties()["socksProxyHost"] = proxyHost
+                        System.getProperties()["socksProxyPort"] = proxyPort
+                        logger.info("Socks Proxy is enabled to $proxyHost:$proxyPort")
+                    }
+                }
+            )
```
Syer10: Don't think this is needed in the tests.

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
schroda: I blame ChatGPT for giving me an incorrect solution (and definitely not me for not correctly checking if that is working 😅)

the idea is to have the latest value wait until the current handling of the last emitted value to be finished and omit all values in between, so basically this https://github.com/Suwayomi/Tachidesk-Server/pull/639/files#diff-65dbed49da02019a15cea5b7d68041ebd4b65a0b5bee5daf91fba2e47b07401bR146

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
schroda: yeah... testing it correctly would be helpful
it can be removed but trying to show it again does nothing, but that just because the setup has to be done again and just writing the instance here does nothing ofc..., it has to be recreated then it works

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
schroda: I forgot about that, I wasn't able to test that (and wasn't able to find where this is actually accessed), so some else would need to check if that is working

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
schroda: fixed

server/src/test/kotlin/suwayomi/tachidesk/test/ApplicationTest.kt
schroda: reverted

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
schroda: fixed (using MutableSharedFlow)

Syer10: Hmm now that I think about it, couldn't you just use a Mutex with Mutex.tryLock instead of withLock?

Syer10: Actually ignore that, it may break something

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
schroda: yes, works

schroda:
the only setting that would require a restart is `webUIEnabled`, I wasn't able to find a way to remove served files during runtime for the used javalin version.
from what I've seen, in later versions the config exposes the `staticFiles` property, which might make it possible to remove served files during runtime

for now this setting is just omitted from the settings that can be changed via the api and can only be changed via the launcher or by editing the config file.

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -8,58 +8,127 @@ package suwayomi.tachidesk.server
...
+
+    fun <T> subscribeTo(flow: Flow<T>, onChange: suspend (value: T) -> Unit, ignoreInitialValue: Boolean = true) {
+        val actualFlow = if (ignoreInitialValue) {
+            flow.drop(1)
+        } else {
+            flow
+        }
+
+        val sharedFlow = MutableSharedFlow<T>(extraBufferCapacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
+        flow.distinctUntilChanged().onEach { sharedFlow.emit(it) }.launchIn(mutableConfigValueScope)
```
Syer10: Is this supposed to be actualFlow?

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
schroda: yes it is, fixed

Syer10 (approved):

Syer10:
@schroda needs to be rebased/merged due do the thumbnail decouple changes

schroda:
rebased onto master

PR #623, opened 2023-07-29
User Accounts

### LINKED ISSUE #298: [Feature Request]  Add Users Accounts on web login
add the feature of user accounts

it would be great to have users accounts, to login on the acces to the web server

my son and i are manga readers and both have the tachiyomi app in each tablet, but sometimes we read on our iphones, (Komga server) it would be great to have the manga reading centralized on one app 

thanks

Syer10:
# Multi-User Accounts (`ui_login`)

Adds a full multi-user mode to Suwayomi-Server: each person on the server gets their own account, library, downloads, tracking, and settings. Gated by the `ui_login` authentication mode; in all other modes there is a single admin account and the per-user features are inert.

Docs: `docs/Multi-User-Authentication.md`

## What's in this PR

**Accounts & auth**
- `ui_login` auth mode: login via the `login` GraphQL mutation (JWT access + refresh tokens); requests authenticate with `Authorization: Bearer <token>` (cookie / query-param fallbacks), including WebSockets.
- Accounts stored in new `USERACCOUNT` / `USERROLES` / `USERPERMISSIONS` tables (migration `M0065`); built-in admin is user `1`, mirroring `server.authUsername` / `server.authPassword` (change the default immediately if the server is network-reachable).
- `setPassword` requires the current password; user `1` is not modifiable via the API.
- **Session invalidation**: per-user `SESSION_VERSION` embedded in every token as `token_version`; any password change (self `setPassword` or recovery redemption) bumps it, logging the account out everywhere.

**Recovery & registration codes** (one-time, admin-issued, bcrypt-hashed at rest, returned in plaintext exactly once; uniform "Invalid or expired code" on all failures)
- `createRecoveryCode` / `redeemRecoveryCode` — 24h TTL, one active code per user; the user sets their own new password (the admin never sees it).
- `createRegistrationCode` / `redeemRegistrationCode` — 7d TTL, controlled self-signup; creates the account (default permissions + `USER` role) and returns a JWT.
- `userCodes` / `revokeUserCode` — list outstanding codes and revoke any.

**Permissions**
- `@RequirePermissions` GQL directive (fail-closed, Admin bypass) + 10 permissions: `INSTALL_EXTENSIONS`, `INSTALL_EXTERNAL_EXTENSIONS`, `UNINSTALL_EXTENSIONS`, `DOWNLOAD_CHAPTERS`, `MANAGE_SETTINGS`, `MANAGE_USERS`, `MANAGE_EXTENSION_STORES`, `MANAGE_SOURCE_PREFERENCES`, `MANAGE_CACHE`.
- User-management API: `user` / `users` (Node/NodeList) and `updateUser` (replace permissions/role; user `1` immutable).
- **Security posture:** new accounts default to `INSTALL_EXTENSIONS`, `INSTALL_EXTERNAL_EXTENSIONS`, `UNINSTALL_EXTENSIONS`, `DOWNLOAD_CHAPTERS` — i.e. self-registered users can install untrusted `.jar` extensions. Deliberate (treat accounts as trusted); trim per account with `updateUser`.

**Per-user data**
- Library, chapters (read/bookmark/lastPageRead), categories, category↔manga, all meta (global/manga/chapter/category/source), and tracking records are per-user (per-user tables; M0063 backfills existing data to user `1`).
- **Downloads are reference-counted and shared**: per-user download intent, global file state — a file is kept while any user has intent and removed when the last one deletes.
- Library updates, SyncYomi sync, Koreader sync, and automated backups are per-user (per-user config, state, scheduling, and one backup file per user).
- Chapter-URL-change migrations carry per-user state (and per-user download state when the shared file is preserved) for all users.

**Per-user settings**
- 31 settings moved per-user (library-update filters, auto-download, OPDS, Koreader behavior, `serveConversions`, SyncYomi) with a typed `userSettings` / `setUserSettings` / `resetUserSettings` GQL API; the global counterparts are deprecated.
- Users without `MANAGE_SETTINGS` get a masked global-settings view (defaults + their own per-user-moved values) so pre-user-account clients keep working.
- One-time backfill pins the legacy single-user server values into user `1`'s user settings on upgrade.

**Backups**
- Exports include the user's settings (new proto field, gated by `autoBackupIncludeUserSettings`); global `serverSettings` are only exported/restored for users with `MANAGE_SETTINGS`.
- Restoring an old backup maps its legacy global values to the importing user's per-user settings.

**Tests & docs**
- ~100 new tests: full GQL end-to-end suite (auth seam, permissions, user codes, per-user settings, per-user backups), plus unit coverage for session invalidation, the settings store, the user-1 backfill, and per-user sync.
- `docs/Multi-User-Authentication.md` + `autoBackupIncludeUserSettings` in the config docs.

## Known limitations / out of scope

- WebUI needs UI for issuing/listing/redeeming codes and user management — this PR exposes the GraphQL API only.
- Shared download queue control (`start`/`stop`/`clear`/`reorder`/`dequeue`) is not per-user gated; `updateStop` cancels all users' library-update jobs; library-update progress is shared.
- ConfigurableSource preferences remain global.
- No user deletion yet; no email/SMS code delivery (out-of-band handoff is the model); no configurability of code TTLs.
- Permission/role changes via `updateUser` take effect at the target's next access-token refresh (claims are embedded in the JWT).


schroda:
what has to be considered as well are user related settings that are currently in the server config

schroda:
> Decide how to update user's manga, update all manga at once or separate them by user?

I'd say everything should be per user.

that should also include downloads.
or how should that be handled? what if two users try to download from the same source, wouldn't then the downloads of one user block the ones from the other?

I guess this would still be a problem even with separated downloads, unless every user uses their own source instance.
but also overall, if every user has a own source instance, couldn't that, in theory, lead to basically spamming sources?

everything that is currently stored on the file system would also have to be per user right? e.g. downloads and thumbnails
I guess it could also be kept global by only deleting files if no user requires them anymore, e.g. for downloads, if a user "deletes" a chapter of manga X and other users still have that chapter downloaded, nothing would get deleted.
but then how would you handle normal or cbz downloads which could be different per user

Syer10:
Library update could be per user, for this I think we would need a task that runs every 6 hours, grabs manga from users who's interval has passed(could be days for some users, 6 at the minimum), and then update them.

Downloader is more special, right now my idea is allow anyone to download, and the download queue will grab the user data based on the subscriber, this would allow unique user data to go to each user. This only works on gql, so the old websocket url won't have user data. Only admins would be able to delete files. Or we could make a userDeleted field in the chapter user data, and if all users who have that manga favorited have the userDeleted field true, then actually delete the file.

I think normal vs cbz downloads would be configured by the system administrator, not per user.

Everything on the filesystem doesn't need to be per user, since manga and chapter ids won't change between users, we dont need to separate anything.

Thats right, if we separate sources per user, rate-limits won't properly apply between http clients. We may have to make source settings global that only the admin can change hmm.

schroda:
> Library update could be per user, for this I think we would need a task that runs every 6 hours, grabs manga from users who's interval has passed(could be days for some users, 6 at the minimum), and then update them.

I guess you would have to decrease the task to 1h since in case someone has it to update between every e.g. 6h and 12h then it would only get triggered every 12h

would this be one global updater or one per user?
in case it's one global updater - what about cases were one user has a huge library while another has a small one and gets "blocked" by the other user or would this be works as expected, and you just have to wait?
e.g. user A updates 50 mangas of source A and user B updates 10 mangas of source A, but they're at the end of the queue

same for the downloader I guess?

> Or we could make a userDeleted field in the chapter user data, and if all users who have that manga favorited have the userDeleted field true, then actually delete the file.

I think that would be a better experience, since you don't have to manually clean up things that aren't needed anymore

> We may have to make source settings global that only the admin can change hmm.

that are the preferences right? I think doing it like that would be the easiest way, but I guess there are also settings which would make sense to be per user instead of global
e.g. going by MangaDex:

- "Data saver" would be something that has to be decided by the one that runs the server
- "Alternative titles in description" could be per user?
- "Default content rating" also per user?

Syer10:
There are technical limitations to source settings, since we need to have 1 source instance to preserve the rate-limit, we can't separate the sources per user, and since sources expect their settings to be in the source, we have to limit them to either be editable by everyone or admin only.

There are drawbacks to library update and downloader sharing, but its the best option we have. We could always sort manga update order according to the the ones that are used the most or updating the smaller libraries first. I think 6 hours is a fine interval, since we can always just bump the user to the next interval. If a user sets 24 hours, they will always get the update at the same time, or if a user selects 6 hours, they will always get updates.

schroda:
I think the limitations should be made quite clear in the documentation once that is in to prevent any wrong expectations

If the "global" update task runs every 6h then I think the possible update interval has to be changed to be only multiples of 6, since everything between them will be bumped to the next higher multiple

Syer10:
Hmm, right I was thinking in Tachiyomi library update terms, I forgot that users can set weird update times like every 7 hours. I think every hour would be fine then

dhvcc:
Hi, any progress on that? Would love to see that feature

L00maca:
Any chance of this ever getting merged? 
I'd love the feature. 

Having to run three different instances is a pain an doesn't help with throttle limiting (plus downloading/keeping multiple copies of the same series)

Syer10:
Its a very low priority feature since none of the devs need it. If I have time I may continue working on it, I am also open to others contributing and completing any of the TODOs.

L00maca:
Took a look at the source, Kotolin is not something I can help with unfortunately :/

GonzaCass:
Hi @Syer10,

I came across your work on the User Accounts feature branch and noticed it's marked as WIP. Since some time has passed, I wanted to ask:

1. Is there any updated version or roadmap for this feature that I could reference?
2. Are there specific outstanding tasks where community help would be valuable?. 
3. Would you be open to collaboration if I wanted to continue development based on your existing work?

I'm particularly interested in helping bring this feature to completion and would appreciate any guidance on:

- Current state of the branch
- Known challenges or blockers
- Architectural decisions I should be aware of

Thanks for your work on this.

Syer10:
@ov3rsp1der I would love some contributions on this!

To answer your questions:
- 1 and 2. Everything in the Todo block in the PR message still needs to be done, I've also littered quite a few Todo's around the codebase of things that need to be fixed or added for user support.
- 3: Yes! I personally don't have a use for this feature so its in the backburner for me. I would defenitly be open to collaborate with people so that this can be implemented, I've done a lot of the groundwork needed for how I would like this feature to work on the server-side, just needs more implementating



- ~~For the branch I'll take a look at merging master into this within the next few days so you have a more recent base to work on.~~ Already done, is now recent enough to work on
- I can't think of any blockers at the top of my head, most of it is ready to be implemented, some design choices need to be taken for different features, but that can be done later once we have more of the groundwork done. If you run into a issue with a place requiring a userId and you don't have one, feel free to default to 0 and add a `// todo: User Accounts`


For Archeticual decisions, it should hopefully be pretty recognizable in the code. There will only be GQL endpoints for user handling(as the REST API is deprecated). We use the `.getAttribute(Attribute.TachideskUser).requireUser()` with the context or dataFetchingEnviornment to get the user, a similar system will be needed for permissions. We join tables to to get data + user data wherever we can. The admin/default user is id `1`, and thats the only user when not in Multi-User mode. There is probably more, feel free to talk to me about this in the Discord whenever you start looking at the code.

GonzaCass:
Thank you! let's see what I can do

zeedif:
With the recent merge of the basic JWT login, I've gotten interested in helping out with multi-user support. Before diving in, I'd like to discuss how this would interact with the existing authentication modes.

For example, once an instance has multiple users, how would one revert back to a 'no-auth' mode, or switch to a single-user basic auth mode? I'm guessing the idea might be that each user gets their own <user-id>/library and the decision to protect it is handled on a per-user basis? Or maybe the server would be toggled between a single-user and multi-user mode entirely?

Syer10:
Each user does get thier own library, the default user(user id 1) is the only user by default and is what will be used when multiuser is disabled.

Enabling multiuser would allow access to the server by other users. All data will be stored in the same way regardless. 

If multiuser is disabled, all data will be kept but only the default user will be accessible

Syer10:
I have merged the branch with the latest server.

zeedif:
> Each user does get thier own library, the default user(user id 1) is the only user by default and is what will be used when multiuser is disabled.
> 
> Enabling multiuser would allow access to the server by other users. All data will be stored in the same way regardless.
> 
> If multiuser is disabled, all data will be kept but only the default user will be accessible

I understand the point about starting with or reverting to the traditional single-user mode. However, it presents a challenge when you consider that users can be deleted and that at least one administrator role must exist.

This leads to scenarios where, for instance, in a two-user setup where both are admins (e.g., ID 1 and ID 2), the second admin could revoke the admin role from the default user (ID 1). In that case, how would the system revert to single-user mode?

I suppose the best approach would be one of two options: either the administrator who makes the change to single-user mode becomes the sole active account, or the administrator can select which account will become the active one in single-user mode.

Syer10:
The way I am thinking of is that there is the admin(id 1), and they can manage permissions for all other users. Manage permissions could also be a permission, but no permissions are can actually be taken away by user 1, since they have all of them.

zeedif:
I'm concerned that tying a permanent, irrevocable admin role to user ID 1 isn't a secure strategy due to potential security risks.

While I know Suwayomi is self-hosted and mostly used on LANs, there's always a chance an instance could be exposed online. If the credentials for user ID 1 were compromised, there would be no way for a second administrator to revoke their permissions.

It's a basic and common practice in any multi-user system for administrators to be able to manage each other's permissions to handle such security threats. I can't think of any examples of systems that use the model you're proposing.

Syer10:
The closest application that sorta models the way I am thinking of taking Suwayomi is Plex. There is 1 central user that manages the app, and everyone else is given access and is secondary.

zeedif:
Hmm, I'm not very familiar with Plex. I tried it briefly but switched to Jellyfin precisely because it wasn't possible for two users to administer the same server.

That said, one concern I have is how OIDC support would be added later under this strategy. I suppose it could be handled like Kavita does, where OIDC accounts are not the same as the accounts generated from within Suwayomi itself.

Perhaps it would be a good idea to do some kind of poll to see how the community would prefer to handle multiple users? I might not be the only one who is not entirely comfortable with the vision of making it similar to Plex's single-admin management.

dhvcc:
> Hmm, I'm not very familiar with Plex. I tried it briefly but switched to Jellyfin precisely because it wasn't possible for two users to administer the same server.
> 
> That said, one concern I have is how OIDC support would be added later under this strategy. I suppose it could be handled like Kavita does, where OIDC accounts are not the same as the accounts generated from within Suwayomi itself.
> 
> Perhaps it would be a good idea to do some kind of poll to see how the community would prefer to handle multiple users? I might not be the only one who is not entirely comfortable with the vision of making it similar to Plex's single-admin management.

I agree with @Syer10 
I think having 1 root admin access and an ability to escalate other user's permission to admin (or add more permissions, whatever) is pretty normal. This way you can have 2 admins, while id=1 account will be the one to be left if you switch off the multi tenancy. AWS also has root account and administrator ones alongside ([ref](https://docs.aws.amazon.com/streams/latest/dev/setting-up.html#:~:text=When%20you%20sign%20up%20for%20an%20AWS%20account%2C%20an%20AWS,that%20require%20root%20user%20access.))

I think if you expose root admin credentials it's on you really

cpiber:
> If the credentials for user ID 1 were compromised, there would be no way for a second administrator to revoke their permissions.

Maybe I'm not seeing it, but why not just change the password? As long as an admin can reset or set the password, that's effectively as good as disabling the account and creating a new one. If we allow changing user names additionally (since everything is tied to IDs anyways), there's no functional difference at all.
There would need to be a mechanism to revoke refresh tokens of course to terminate the compromised session, but I don't see a need to revoke the entire user. Easiest would be to just nuke the JWT secret and restart the server, that effectively terminates all sessions immediately.

zeedif:
Regarding dhvcc's point about the AWS root account model, I agree that's a good example. My main concern was a Plex-like, single-admin strategy, which I personally find too restrictive as it prevents delegating full administrative permissions. The AWS approach addresses my primary worry, as it fits Suwayomi's current structure well and would avoid significant refactoring. However, my question is how OIDC integration would work. We would need to consider whether to link external accounts to standard users—who could then be promoted in their role—while keeping the root user separate, or to simply ignore role customizations for any OIDC account that matches the root user.

Syer10:
I do not have any actual expierence with OIDC, but my vision would be that admin roles would be copied over to Suwayomi where possible, but the root user would never be able to have thier roles or permissions changed.

I think if they want thier accounts to be controlled by OIDC, that they should leave the root user unattached to an OIDC account.

OIDC users would be a regular user, maybe with a OIDC table row that references the specific user its attached to, or just a column on the user table. Depending on which is easiest and makes the most sense to implement.

docs/Multi-User-Authentication.md
```diff
@@ -0,0 +1,156 @@
...
+
+| Permission | Gates |
+|---|---|
+| `INSTALL_EXTENSIONS` | Installing extensions from the store |
+| `INSTALL_EXTERNAL_EXTENSIONS` | Installing untrusted extensions (`installExternalExtension`) |
+| `UNINSTALL_EXTENSIONS` | Uninstalling extensions |
+| `DOWNLOAD_CHAPTERS` | Enqueueing chapter downloads |
+| `ACCESS_NSFW` | Viewing/fetching NSFW sources and extensions |
+| `MANAGE_SETTINGS` | Reading (real values) / writing global server settings; including server settings in backup exports and applying them on backup restore |
+| `MANAGE_USERS` | Registering users, listing users, granting/revoking permissions, setting roles, issuing/redeeming user codes |
```
cpiber:
```suggestion
| `MANAGE_USERS` | Registering users, listing users, granting/revoking permissions, setting roles, issuing user codes |
```

docs/Multi-User-Authentication.md
```diff
@@ -0,0 +1,156 @@
...
+| `MANAGE_CACHE` | Clearing cached images and webview cookies/cache |
+
+### Default permissions for new accounts
+
+**Important security posture:** a newly created account (via `register` or a registration code) is granted, by default:
+
+```
+INSTALL_EXTENSIONS
+INSTALL_EXTERNAL_EXTENSIONS
+UNINSTALL_EXTENSIONS
```
cpiber, line 91: Should `UNINSTALL_EXTENSIONS` really be part of the default permissions? I can get behind `INSTALL_EXTENSIONS`, to allow other users to expand the available content, but uninstalling extensions can break other users' library

docs/Multi-User-Authentication.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Settings API for users without MANAGE_SETTINGS
+
+Pre-user-account clients still call the global `settings` query and the `setSettings` / `resetSettings` mutations. For users **without** `MANAGE_SETTINGS` these do not return an error; instead they operate on a masked view:
+
+- `settings` returns all settings at their **default values**, except the deprecated settings that were moved to per-user storage — those show the caller's **own** effective user setting, so legacy clients keep working for the options they know.
+- `setSettings` does **not** change the global config. Only the per-user-moved settings take effect, applied to the caller's own user settings.
+- `resetSettings` does **not** reset the global config; it resets the caller's per-user-moved user settings.
+
+Users **with** `MANAGE_SETTINGS` (including the `ADMIN` role) see and write the ServerConfig as before.
```
cpiber, line 142: Does this include the deprecated settings? Would this not break legacy clients as well?

Edit: Checking below, I see that deprecated settings are applied to user settings always, but for admins it is also synced to the global settings. I believe this should be documented.

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -616,78 +641,81 @@ class ServerConfig(
...
+                    it.value.headers?.map { header ->
+                        BackupSettingsDownloadConversionHeaderType(
+                            header.key,
+                            header.value,
+                        )
+                    },
+                )
+            }
+        },
+    )
```
cpiber: Share this with UserConfig.kt?

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
```diff
@@ -83,29 +89,31 @@ class ChapterFlagCountForMangaDataLoader : KotlinDataLoader<Int, MangaChapterSta
...
+                            .When(ChapterUserTable.isDownloaded eq true, intLiteral(1))
                             .Else(intLiteral(0))
                             .sum()
 
                     val bookmarkCount =
                         Case()
-                            .When(ChapterTable.isBookmarked eq true, intLiteral(1))
+                            .When(ChapterUserTable.isBookmarked eq true, intLiteral(1))
                             .Else(intLiteral(0))
                             .sum()
```
cpiber, line 112: Do I not understand something, or does this count for all users, not just the current user? The `statsByMangaId` is per-user, so this seems inconsistent

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
```diff
@@ -153,24 +182,32 @@ class DownloadMutation {
     )
 
     @RequireAuth
-    fun dequeueChapterDownloads(input: DequeueChapterDownloadsInput): CompletableFuture<DequeueChapterDownloadsPayload?> {
+    fun dequeueChapterDownloads(
```
cpiber, line 186: I feel like this should also have `@RequirePermissions(UserPermission.DOWNLOAD_CHAPTERS)`, because this allows a nonprivileged user to deny new downloads (denial of service)

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ExtensionMutation.kt
```diff
@@ -79,10 +87,29 @@ class ExtensionMutation {
         }
     }
 
+    private fun checkUpdatePermissions(
```
cpiber, line 87: What about `update`? Should this not be considered by `UserPermission.INSTALL_EXTENSIONS`?

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/UserQuery.kt
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    @RequireAuth
+    @RequirePermissions(UserPermission.MANAGE_USERS)
+    fun users(
+        condition: UserCondition? = null,
+        filter: UserFilter? = null,
+        order: List<UserOrder>? = null,
+        before: Cursor? = null,
+        after: Cursor? = null,
```
cpiber, line 207: How many users are we expecting to make it this complicated? Just a regular query should be fine, no?

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/UserCodeType.kt
```diff
@@ -0,0 +1,43 @@
...
+    constructor(code: UserCodeService.OutstandingCode) : this(
+        code.id,
+        code.purpose,
+        code.createdAt,
+        code.expiresAt,
+        code.user,
+        code.createdBy,
+    )
+
+    fun user(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<out UserType?>? {
```
cpiber: You always return a future, so why is the return type nullable? Only the `UserType` needs to be nullable

server/src/main/kotlin/suwayomi/tachidesk/manga/model/table/ChapterUserTable.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    val isRead = bool("read").default(false)
+    val isBookmarked = bool("bookmark").default(false)
+    val lastPageRead = integer("last_page_read").default(0)
+    val lastReadAt = long("last_read_at").default(0)
+
+    val isDownloaded = bool("is_downloaded").default(false)
+    val isDownloadRequested = bool("is_download_requested").default(false)
+
+    val koreaderHash = varchar("koreader_hash", 32).nullable()
```
cpiber, line 28: Does this need to be here? Should be the same for all users as I understand it

server/src/main/kotlin/suwayomi/tachidesk/server/user/UserCodeService.kt
```diff
@@ -0,0 +1,377 @@
...
+        val user: Int?,
+        val createdBy: Int,
+        val createdAt: Long,
+        val expiresAt: Long,
+    )
+
+    private fun now(): Instant = Clock.System.now()
+
+    /**
+     * Generates a 128-bit random code encoded as 26 Crockford base32 characters.
```
cpiber: Why not just hex or base64 encode?

server/src/main/kotlin/suwayomi/tachidesk/server/user/UserCodeService.kt
```diff
@@ -0,0 +1,377 @@
...
+            transaction {
+                UserAccountTable
+                    .selectAll()
+                    .where { UserAccountTable.username.lowerCase() eq username.lowercase() }
+                    .isNotEmpty()
+            }
+
+        if (userExists) {
+            throw Exception("Username already exists")
+        }
```
cpiber: This invalidates the code in the case where the user already exists, even though it wasn't actually used, checks should be reversed

cpiber (commented):
Some things from looking over the code, will play around with it over the next few days

cpiber:
One major thing I noticed: The legacy REST API does not handle the new permissions. We briefly discussed on discord to remove deprecated settings before the next release, does this include the REST API? If so, I am fine with ignoring it here, otherwise that should be handled as well (at the least by stubbing like the chapter download API).

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/UserQuery.kt
Syer10: This is just following our NodeList template, it should be fine

server/src/main/kotlin/suwayomi/tachidesk/manga/model/table/ChapterUserTable.kt
Syer10: Each user has their own `koreaderSyncChecksumMethod` thats used to generate this.

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/UserCodeType.kt
Syer10: Fixed

server/src/main/kotlin/suwayomi/tachidesk/server/user/UserCodeService.kt
Syer10: Changed

server/src/main/kotlin/suwayomi/tachidesk/server/user/UserCodeService.kt
Syer10: Fixed

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
Syer10: This should only show stats for the current user, as these are applied to the ChapterTable joined to ChapterUserTable with a where filter for the current userId.

cpiber: Right, I misread it, I interpreted the `unreadCount` etc. as individual transactions, when they're only a single expression of the overall statement, my bad

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ExtensionMutation.kt
Syer10: Updating is not really destructive, and the user may be the only one using the extension and may need the updates.

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
Syer10: Added

docs/Multi-User-Authentication.md
Syer10: Can you expand on this? What global settings vs user settings?

cpiber: Actually, just ignore this, I overlooked that `SettingsType` proxies to the `UserSettingsRegistry` for per-user-moved settings. I thought that multiple admins would conflict with each other, but since the deprecated settings are not written nor read, it doesn't actually matter.

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
Syer10: Added

docs/Multi-User-Authentication.md
Syer10: I feel like Suwayomi should be default trust, and permissions can be taken away if abused

Syer10:
> One major thing I noticed: The legacy REST API does not handle the new permissions. We briefly discussed on discord to remove deprecated settings before the next release, does this include the REST API? If so, I am fine with ignoring it here, otherwise that should be handled as well (at the least by stubbing like the chapter download API).

Lets remove REST after this is merged then. We've kept it long enough.

schroda:
> NSFW is enforced on the source/extension surface and fetchSourceManga, not on content-fetch operations (fetchManga / fetchChapters / fetchChapterPages); legacy page-image and CBZ REST endpoints serve content without an ownership/NSFW check.

This permission is not enforced properly. You can easily get nsfw sources and extensions through nested fields

Syer10:
> > NSFW is enforced on the source/extension surface and fetchSourceManga, not on content-fetch operations (fetchManga / fetchChapters / fetchChapterPages); legacy page-image and CBZ REST endpoints serve content without an ownership/NSFW check.
> 
> This permission is not enforced properly. You can easily get nsfw sources and extensions through nested fields

ATM the NSFW check is only surface-level(as documented). More thorough checks can be a future feature in my opinion.

schroda:
I guess the question is what you mean by `"extension/source surface"`. The doc says `"Viewing/fetching NSFW sources and extensions"`
When I read that I assume it is not possible to get a source/extension that is NSFW at all, but that is not the case. The server will only enforce it on the source/extension queries/mutations directly, but there are x other ways to get it via the api.

So the way I see it, is that the permission doesn't enforce what it says it does

Syer10:
The `extension(s)`, `source(s)`, `fetchSourceManga`, `fetchExtension` queries and mutations is what is currently filtered/blocked. The main ones which are used by clients, priority 1. Yes if someone without the NSFW permission goes to the API they can get pass parts with little difficulty, this is planned to be expanded in the future.

schroda:
Who is going to get that permission denied? I'd say only children. So if a parent reads that permission in the docs, they won't know that it won't actually prevent them from being able to access it if they really want to.

Which means there are only two ways to prevent nsfw content from being accessible.
The first one is to not have any nsfw extensions installed in the first place on a shared suwayomi instance with users without the permission.
The second is to have a separate instance for users that are not supposed to be able to access nsfw content.
Both make this permission pointless.

I get that in the future this might be expanded, but as it is, I just don't see how it makes sense to add it in that state. It's just misleading

Syer10:
Removed the NSFW permission for now.

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/ChapterType.kt
```diff
@@ -100,6 +92,39 @@ class ChapterType(
 
     fun meta(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<List<ChapterMetaType>> =
         dataFetchingEnvironment.getValueFromDataLoader<Int, List<ChapterMetaType>>("ChapterMetaDataLoader", id)
+
+    fun user(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<ChapterUserType?> =
+        dataFetchingEnvironment.getValueFromDataLoader<Int, ChapterUserType?>("ChapterUserForChapterDataLoader", id)
+
+    @GraphQLDeprecated("Use user.isRead instead")
```
schroda: Why were these fields moved under `user`, is it necessary? From a client perspective it's quite clear that these fields are for the current user. Having to access them via `chapter.user.x` feels unnatural

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/MangaType.kt
```diff
@@ -178,6 +178,21 @@ class MangaType(
 
     fun trackRecords(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<TrackRecordNodeList> =
         dataFetchingEnvironment.getValueFromDataLoader<Int, TrackRecordNodeList>("TrackRecordsForMangaIdDataLoader", id)
+
+    fun user(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<MangaUserType?> =
+        dataFetchingEnvironment.getValueFromDataLoader<Int, MangaUserType?>("MangaUserForMangaDataLoader", id)
+
+    @GraphQLDeprecated("Use user.inLibrary instead")
+    fun inLibrary(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<Boolean> =
```
schroda: Same as [here](https://github.com/Suwayomi/Suwayomi-Server/pull/623/changes#r3945880609)

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/ChapterType.kt
schroda: Why were these fields moved under `user`, is it necessary? From a client perspective it's quite clear that these fields are for the current user. Having to access them via `chapter.user.x` feels unnatural

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/MangaType.kt
```diff
@@ -178,6 +178,21 @@ class MangaType(
 
     fun trackRecords(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<TrackRecordNodeList> =
         dataFetchingEnvironment.getValueFromDataLoader<Int, TrackRecordNodeList>("TrackRecordsForMangaIdDataLoader", id)
+
+    fun user(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<MangaUserType?> =
+        dataFetchingEnvironment.getValueFromDataLoader<Int, MangaUserType?>("MangaUserForMangaDataLoader", id)
+
+    @GraphQLDeprecated("Use user.inLibrary instead")
```
schroda: Same as [here](https://github.com/Suwayomi/Suwayomi-Server/pull/623/changes#r3945880609)

server/src/main/kotlin/suwayomi/tachidesk/server/settings/UserSettings.kt
```diff
@@ -0,0 +1,200 @@
...
+    @Suppress("UNCHECKED_CAST")
+    private fun <T : Any> entry(
+        userId: Int,
+        setting: UserSetting<T>,
+    ): Entry<T> =
+        cache.getOrPut(cacheKey(userId, setting.key)) {
+            val e = Entry<T>()
+            val stored = readStored(userId, setting.key)
+            val default = setting.defaultValue
+            e.flow = MutableStateFlow(if (stored != null) decode(stored, setting) else default)
```
schroda: Was the logic to ensure that read values are valid or getting converted to valid ones intentionally removed?

When the settings got streamlined, [inline value normalization](https://github.com/Suwayomi/Suwayomi-Server/pull/1614/changes#diff-b482d3903208d5a54731a75eb5531a40a8024f176f3973d0973812448d95336dL25) got removed because it was moved to the `serverConfig` itself. Which made sure that the initially read value is valid as well as preventing that a invalid value was written.

The user settings now only prevent writing invalid values.

Due to having to edit the database manually it's harder for a user to mess this up. Though, it was also meant as a safeguard so that users couldn't circumvent intentional limitations - e.g., min/max values - which would now be possible

server/src/main/kotlin/suwayomi/tachidesk/global/impl/sync/SyncYomiSyncService.kt
```diff
@@ -125,31 +128,32 @@ object SyncYomiSyncService {
...
+        userId: Int,
         syncData: SyncData,
         full: Boolean,
         startDate: Instant,
         setSyncState: (SyncManager.SyncState) -> Unit,
     ): SyncResult {
         val backup = syncData.backup ?: return SyncResult(null, changed = false, protocolV2 = true)
-        val host = serverConfig.syncYomiHost.value
-        val apiKey = serverConfig.syncYomiApiKey.value
+        val host = userSettings.value(userId, userConfig.syncYomiHost)
```
schroda: This feels verbose, isn't it possible to keep it similar to the original syntax, e.g., something like `userConfig(userId).syncYomiApiKey.value`?

docs/Multi-User-Authentication.md
```diff
@@ -0,0 +1,153 @@
+# Multi-User Authentication
+
+Suwayomi supports a multi-user mode in which each person on the server has their own account, their own library, their own downloads, and their own per-user settings. This page explains how to enable it, how accounts work, and what the security posture is.
+
+Multi-user mode is gated by the `ui_login` authentication mode. In all other modes (`none`, `basic_auth`, `simple_login`) there is effectively a single admin account and the per-user features are inert.
```
schroda: I'd say e.g. `inactive` is easier to understand than `inert`

docs/Multi-User-Authentication.md
schroda: How about changing it so that the mutation to create the registration code requires one to pass initial permissions?

Even if the default permissions are kept, I'd say it should be possible to pass permissions that overwrite the defaults, so that one doesn't have to wait for the registration code to be used to be able to adjust them

docs/Multi-User-Authentication.md
```diff
@@ -0,0 +1,153 @@
...
+- `permissions` (non-null) replaces the account's full permission set.
+- `role` (non-null) replaces the account's single role (`USER` or `ADMIN`). The `VISITOR` role cannot be granted, and user `1` cannot be modified.
+
+Note: permission and role changes take effect at the target account's next access-token refresh (up to `jwtTokenExpiry`), because the claims are embedded in the JWT.
+
+## Session invalidation
+
+Accounts use stateless JWTs (HMAC, no token store). To let the server revoke a session (e.g. after a password change), each account has a `SESSION_VERSION` counter that is embedded in every access and refresh token as the `token_version` claim.
+
+- Changing a password — either the account's own `setPassword` or redeeming a recovery code — bumps the `SESSION_VERSION`.
```
schroda, line 108: When using the refresh token to get a new access token, we should also bump the version, so that the old refresh token is invalid.

I.e., bump version, create new tokens and return them

server/src/main/kotlin/suwayomi/tachidesk/global/impl/sync/SyncManager.kt
```diff
@@ -66,162 +76,283 @@ object SyncManager {
...
+
+    /**
+     * Idempotently subscribe to a user's (enabled, interval) config so their scheduled sync task is kept in sync.
+     */
+    private fun subscribeToUserSyncConfig(userId: Int) {
+        if (!subscribedUsers.add(userId)) {
+            return
+        }
+
+        val enabledFlow = userSettings.flow(userId, userConfig.syncYomiEnabled)
```
schroda: This and the value reads feels verbose, isn't it possible to keep it similar to the original syntax, e.g., something like `userSettings/userConfig(userId).syncYomiEnabled` and `userSettings/userConfig(userId).syncYomiEnabled.value`?

server/src/main/kotlin/suwayomi/tachidesk/global/impl/sync/SyncManager.kt
```diff
@@ -66,162 +76,283 @@ object SyncManager {
...
-                            name = "sync",
-                        )
+    private fun rescheduleUserSync(
+        userId: Int,
+        enabled: Boolean,
+        interval: Duration,
+    ) {
+        val shouldSchedule = enabled && interval > 0.seconds
+        val existing = scheduledSyncTasks[userId]
+
```
schroda, line 183:
```suggestion
  if (!shouldSchedule) {
    ...
    return
  }
```

would be easier to read then if else imo

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/BackupMutation.kt
```diff
@@ -76,25 +84,40 @@ class BackupMutation {
...
+            } else {
+                BackupFlags(
+                    includeManga = BackupFlags.DEFAULT.includeManga,
+                    includeCategories = input?.includeCategories ?: BackupFlags.DEFAULT.includeCategories,
+                    includeChapters = input?.includeChapters ?: BackupFlags.DEFAULT.includeChapters,
+                    includeTracking = input?.includeTracking ?: BackupFlags.DEFAULT.includeTracking,
+                    includeHistory = input?.includeHistory ?: BackupFlags.DEFAULT.includeHistory,
+                    includeClientData = input?.includeClientData ?: BackupFlags.DEFAULT.includeClientData,
+                    includeServerSettings = input?.includeServerSettings ?: BackupFlags.DEFAULT.includeServerSettings,
+                    includeUserSettings = BackupFlags.DEFAULT.includeUserSettings,
```
schroda, line 108:
```suggestion
                    includeUserSettings = input?.includeUserSettings ?: BackupFlags.DEFAULT.includeUserSettings,
```

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/UserSettingsMutation.kt
```diff
@@ -0,0 +1,82 @@
...
+    @GraphQLIgnore
+    fun updateSettings(
+        userId: Int,
+        settings: PartialUserSettingsType,
+    ) {
+        // Ensure all UserSetting descriptors are registered in UserSettingsRegistry before looking them up;
+        // otherwise an uninitialized registry would silently drop every field
+        userConfig
+
+        settings
```
schroda, line 53: Don't the settings need to be validated before setting them?
In case a setting value is invalid, the mutation will fail, potentially having updates settings partially which the client will never get informed about

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/ChapterType.kt
schroda, line 99: Why were these fields moved under user, is it necessary? From a client perspective it's quite clear that these fields are for the current user. Having to access them via `chapter.user.x` feels unnatural

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/MangaType.kt
schroda, line 185: Same as [here](https://github.com/Suwayomi/Suwayomi-Server/pull/623/changes#r3945880609)

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/Downloader.kt
```diff
@@ -178,10 +182,28 @@ class Downloader(
...
+                            it[isDownloaded] = true
+                        }
+
+                        ChapterUserTable.update(
+                            {
+                                (ChapterUserTable.chapter eq download.chapterId) and
+                                    (ChapterUserTable.user inList requestingUserIds)
+                            },
+                        ) {
+                            it[ChapterUserTable.isDownloaded] = true
```
schroda, line 205: Doesn't `isDownloadRequested` need to be set to `false` here?

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0065_AddUsers.kt
```diff
@@ -0,0 +1,608 @@
...
+            return """
+            ALTER TABLE $table
+                DROP CONSTRAINT UC_$table;
+
+            ALTER TABLE $table
+                ADD CONSTRAINT UC_$table UNIQUE ($groupBy);
+
+            """.also {
+                println(it)
+            }
```
schroda:
```suggestion
            """
```

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0065_AddUsers.kt
```diff
@@ -0,0 +1,608 @@
...
+import org.jetbrains.exposed.v1.jdbc.SchemaUtils
+import org.jetbrains.exposed.v1.jdbc.transactions.TransactionManager
+import org.jetbrains.exposed.v1.jdbc.vendors.currentDialectMetadata
+import suwayomi.tachidesk.global.impl.util.Bcrypt
+import suwayomi.tachidesk.graphql.types.DatabaseType
+import suwayomi.tachidesk.server.database.migration.helpers.toSqlName
+import suwayomi.tachidesk.server.serverConfig
+
+@Suppress("ClassName", "unused")
+class M0065_AddUsers : Migration() {
```
schroda, line 24: There are a bunch of indexes that should be added. 
E.g., foreign keys aren't creating one by default.

server/src/main/kotlin/suwayomi/tachidesk/server/settings/UserSettings.kt
schroda: Was the logic to ensure that read values are valid or getting converted to valid ones intentionally removed?

When the settings got streamlined, [inline value normalizations](https://github.com/Suwayomi/Suwayomi-Server/pull/1614/changes#diff-b482d3903208d5a54731a75eb5531a40a8024f176f3973d0973812448d95336dL25) got removed because it was moved to the `serverConfig` itself. Which made sure that the initially read value is valid as well as preventing that a invalid value was written.

The user settings now only prevent writing invalid values.

Due to having to edit the database manually it's harder for a user to mess this up. Though, it was also meant as a safeguard so that users couldn't circumvent intentional limitations - e.g., min/max values - which would now be possible

server/src/main/kotlin/suwayomi/tachidesk/server/user/UserSettingsBackfill.kt
```diff
@@ -0,0 +1,117 @@
...
+            val decoded = userSettings.decode(value.unwrapped().toString(), setting as UserSetting<Any>)
+            if (decoded != setting.defaultValue) {
+                userSettings.setAny(1, setting, decoded)
+            }
+            applied++
+        } catch (e: Exception) {
+            logger.error(e) { "Failed to apply backfilled value for setting $key; skipping" }
+        }
+    }
+
```
schroda, line 120: Is the file intentionally not getting deleted?

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -471,4 +481,29 @@ fun applicationSetup() {
         },
         ignoreInitialValue = false,
     )
+
+    serverConfig.subscribeTo(
```
schroda, line 488: What happens if the credentials are changed via the settings mutation and this db update fails?

docs/Multi-User-Authentication.md
Syer10: What do you mean?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/BackupMutation.kt
Syer10: There is no `input.includeUserSettings`, `input.includeServerSettings` and such are deprecated and are to be removed.

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/ChapterType.kt
Syer10: To more clearly define the boundary between user data and server data. It doesn't make much of a different on the clients end anyway.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/Downloader.kt
Syer10: No we use isDownloadRequested as the reference counter, isDownloaded is the status of the download to the user.

docs/Multi-User-Authentication.md
Syer10: Applied

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/UserSettingsMutation.kt
Syer10: Fixed

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0065_AddUsers.kt
Syer10: Fixed

server/src/main/kotlin/suwayomi/tachidesk/server/user/UserSettingsBackfill.kt
Syer10: Fixed, now deleted after migration is finished

server/src/main/kotlin/suwayomi/tachidesk/global/impl/sync/SyncManager.kt
Syer10: Found a good way of doing it

server/src/main/kotlin/suwayomi/tachidesk/global/impl/sync/SyncManager.kt
Syer10: Applied

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0065_AddUsers.kt
Syer10: Added a bunch, let me know if you have any further suggestions

server/src/main/kotlin/suwayomi/tachidesk/server/settings/UserSettings.kt
Syer10: Added validation to decoded values

server/src/main/kotlin/suwayomi/tachidesk/server/user/SessionVersion.kt
```diff
@@ -0,0 +1,51 @@
...
+                    .where { UserAccountTable.id eq userId }
+                    .first()[UserAccountTable.sessionVersion]
+            }
+        }
+
+    fun bump(userId: Int) {
+        transaction {
+            UserAccountTable
+                .update({ UserAccountTable.id eq userId }) {
+                    it[UserAccountTable.sessionVersion] = UserAccountTable.sessionVersion + 1
```
cpiber: overflow?

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
Syer10: Added error handling and fallbacks

Syer10:
@schroda the the review comments have been handled or responded to.

docs/Multi-User-Authentication.md
schroda: When using the `refreshToken` mutation to get a new `accessToken`, we just create and return a new `accessToken`.
What I meant is that the `refreshToken` should be rotated when calling that mutation (invalidate the one used to create the new `accessToken` and return both a new `refreshToken` and `accessToken`).

Though, you can just ignore me, because I ignored that with the session implementation that would just lead to all devices of a user to get logged out.

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/BackupMutation.kt
schroda: Nervermind me, didn't read properly

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/ChapterType.kt
schroda: It does create a breaking change which requires the code to be updated (for quite a few places since chapter/manga data is the core information). For a differentiation in client only data which the client just doesn't care about. 
A specific user is logged in, the client requests data, which means the returned data is for that user. The server side implementation of how the user data is handled is of no interest of the client.

Besides this, if the `user` field is kept, I'd say that the user specific data should be considered a requirement for the "basic chapter/manga type", since a client will basically always request this data. Which currently would mean two sequential data loader calls.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/Downloader.kt
schroda: It's used as the reference counter for if it is in the queue for that user. Once the download succeeded, it is not in the queue anymore. Which currently does not reset the in-queue-reference counter of the users. Or am I mistaken?

Syer10: By reference counter I mean the reference that the user is one of the users who want that chapter.

The chapter is kept downloaded even if one of the users who wants that chapter clicks delete. It is only deleted if all users who requested that download clicks delete.

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/ChapterType.kt
Syer10: I've kept the user field, but added user data to the basic construction of MangaType and ChapterType.

aaronbamblett:
Three server issues I encountered while testing against `25cba596`:

**Built-in admin username changes on startup**

In `ServerSetup.kt`, the username collision query includes user ID `1`. If the built-in admin already has the configured username, the query finds that same account, treats its name as taken, and writes the username plus random digits back to it.

Starting the server with the configured username already assigned to the built-in admin should preserve that name. Instead, login using the configured username can fail after startup. The collision check needs to exclude the account being updated.

**Download requests cleared before completion**

In `refreshChapterPageList()`, fetching the page list clears `ChapterUserTable.isDownloadRequested` for every user associated with the chapter.

The downloader subsequently uses `isDownloadRequested` to determine which users should have `isDownloaded` set on completion. For a download that needs its page list fetched, this can erase the requesting users before completion is recorded. The files can download successfully while the requesting account still reports the chapter as not downloaded.

Refreshing page metadata should preserve pending download requests.

**Browsing existing manga throws a database exception**

In `MangaList.insertOrUpdate()`, the existing-manga filter reads `coalesceDefault(MangaUserTable.inLibrary, false)` from the result row, but the query did not select that expression.

When browse results include manga already stored in the database, this path can throw `COALESCE(MANGAUSER.IN_LIBRARY, FALSE) is not in record set` instead of returning the results.

The filter needs to read the selected nullable `inLibrary` column and treat a missing value as false, or explicitly select the expression it reads.

I’ll follow up with the fixes I made during testing for you to review.

aaronbamblett:
I’ve opened three fixes against your branch:

- [Admin username changing on startup](https://github.com/Syer10/Tachidesk/pull/1)
- [Download requests cleared during page refresh](https://github.com/Syer10/Tachidesk/pull/2)
- [Browsing existing manga throwing an exception](https://github.com/Syer10/Tachidesk/pull/3)

cpiber (approved):

aaronbamblett:
Opened a fix against your branch: https://github.com/Syer10/Tachidesk/pull/4

Once one user has read a chapter, any other user's progress on it is accepted and discarded, so their library stays unread. Found it moving my wife from Mihon.

PR #627, opened 2023-07-30
Info Queries

Syer10:


schroda:
should we include webUI related info (version, channel, udpate check, triggering update (mutation)) here as well or in a separate query?

Syer10:
I think it would be fine in this Query

PR #629, opened 2023-08-01
Feature/graphql server settings

schroda:
is rebased onto #617

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
@@ -1,8 +1,12 @@
 package suwayomi.tachidesk.graphql.mutations
 
 import suwayomi.tachidesk.graphql.types.PartialSettingsType
+import suwayomi.tachidesk.graphql.types.Settings
 import suwayomi.tachidesk.graphql.types.SettingsType
+import suwayomi.tachidesk.server.SERVER_CONFIG_MODULE_NAME
+import suwayomi.tachidesk.server.ServerConfig
 import suwayomi.tachidesk.server.serverConfig
+import xyz.nulldev.ts.config.GlobalConfigManager
 
 class SettingsMutation {
     data class SetSettingsInput(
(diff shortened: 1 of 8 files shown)


server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SettingsType.kt
```diff
@@ -0,0 +1,208 @@
...
+    val backupPath: String?
+    val backupTime: String?
+    val backupInterval: Int?
+    val backupTTL: Int?
+
+    // local source
+    val localSourcePath: String?
+}
+
+data class PartialSettingsType(
```
Syer10, line 69: I see that these are nullable, but they don't have a default value, does this allow partial editing or does the user have to set the values as null manually?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
```diff
@@ -0,0 +1,89 @@
...
+    data class ResetSettingsPayload(
+        val clientMutationId: String?,
+        val settings: SettingsType
+    )
+
+    fun resetSettings(input: ResetSettingsInput): ResetSettingsPayload {
+        val (clientMutationId) = input
+
+        GlobalConfigManager.resetUserConfig()
+        val defaultServerConfig = ServerConfig({ GlobalConfigManager.config.getConfig(SERVER_CONFIG_MODULE_NAME) })
```
Syer10, line 82: This resets the running server? All the MutableStateFlows update properly?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    fun setSettings(input: SetSettingsInput): SetSettingsPayload {
+        val (clientMutationId, settings) = input
+
+        updateSettings(settings)
+
+        return SetSettingsPayload(clientMutationId, SettingsType())
+    }
+
+    data class ResetSettingsInput(val clientMutationId: String?)
```
Syer10: Add default null

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SettingsType.kt
schroda: it does, I guess because it's a property of a required object parameter?

Syer10: Then its good 👍

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
schroda: yes, this is what happens:

 - `GlobalConfigManager.resetUserConfig()` resets the config file
 - `val defaultServerConfig = ...` creates a temp ServerConfig instance to access the variables without having to manually retrieve them from the config
 - `updateSettings(settings)` updates the flows

and I have to add a guard to prevent updating the config file in case the value that should be set is the same as the current value in the config
since currently the file gets reset and then every flow value change updates the file again...

- [x] add guard to config value updates

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
schroda: done

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
schroda: added guard

Syer10:
Now this needs to be rebased 😄

schroda:
> Now this needs to be rebased 😄

done

Syer10 (approved):

PR #639, opened 2023-08-05
Fix/downloader manager persisting queue

schroda:
is rebased onto #638

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -20,7 +20,9 @@ import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.launchIn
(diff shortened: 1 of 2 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -124,6 +138,31 @@ object DownloadManager {
...
+        saveQueueFlow.flatMapLatest {
+            saveQueue.acquire()
+            flow { emit(it) }
+        }.onEach {
+            mutex.withLock {
+                saveDownloadQueue()
+                saveQueue.release()
+            }
+        }.launchIn(scope)
+    }
```
Syer10: I'm pretty sure this can just be replaced by using

```
val saveQueueFlow = MutableSharedFlow(extraBufferCapacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
```
instead of this magic

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
schroda: but everyone needs a little magic in their lives - updated
your-a-wizard-harry

Syer10 (approved):

PR #656, opened 2023-08-12
Correctly emit changed values

schroda:
"SharedFlow::emit" blocked the flow due not being called in a new coroutine.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -20,6 +20,7 @@ import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.launchIn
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -108,7 +109,8 @@ class ServerConfig(getConfig: () -> Config, val moduleName: String = SERVER_CONF
         }
 
         val sharedFlow = MutableSharedFlow<T>(extraBufferCapacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
-        actualFlow.distinctUntilChanged().onEach { sharedFlow.emit(it) }.launchIn(mutableConfigValueScope)
+        actualFlow.distinctUntilChanged().onEach { mutableConfigValueScope.launch { sharedFlow.emit(it) } }
```
Syer10: I think changing it to mapLatest would be a better solution

schroda: done

Syer10 (approved):

PR #657, opened 2023-08-12
Feature/updater provide more info about update

schroda:


server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/UpdateQuery.kt
```diff
@@ -4,21 +4,18 @@ import org.kodein.di.DI
...
-            isRunning = status.running,
-            pendingJobs = UpdateStatusType(status.statusMap[JobStatus.PENDING]?.map { it.id }.orEmpty()),
-            runningJobs = UpdateStatusType(status.statusMap[JobStatus.RUNNING]?.map { it.id }.orEmpty()),
-            completeJobs = UpdateStatusType(status.statusMap[JobStatus.COMPLETE]?.map { it.id }.orEmpty()),
-            failedJobs = UpdateStatusType(status.statusMap[JobStatus.FAILED]?.map { it.id }.orEmpty())
-        )
+        return UpdateStatus(updater.status.value)
+    }
+
+    data class LastUpdateTimestampPayload(val timestamp: Long)
```
Syer10, line 16: Eventually I will want to move this to a specific timestamp scalar, but since we are using Long and Int already for manga and chapters its fine.

Syer10 (approved):

PR #661, opened 2023-08-20
Fix some settings not being applied properly

Syer10:


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/backup/proto/ProtoBackupExport.kt
```diff
@@ -62,7 +62,8 @@ object ProtoBackupExport : ProtoBackupBase() {
                     timeOfDay
                 )
             },
-            ::scheduleAutomatedBackupTask
+            ::scheduleAutomatedBackupTask,
+            ignoreInitialValue = false
```
schroda: does this work?
I think I started the interval in the ServerSetup because the ProtoBackupExport never gets loaded unless a backup gets created

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -79,7 +79,7 @@ fun applicationSetup() {
         } else {
             setLogLevel(Level.INFO)
         }
-    })
+    }, ignoreInitialValue = false)
```
schroda, line 82: is this 
```
        // set log level early
        if (debugLogsEnabled(config)) {
            setLogLevel(Level.DEBUG)
        }
```
still needed in the ConfigManger with this change?

I believe the issue with the debug logs is because I forgot to set the default log level, which was INFO before in the logback config file, when I added the file logging

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -53,7 +53,7 @@ class Updater : IUpdater {
     private var currentUpdateTaskId = ""
 
     init {
-        serverConfig.subscribeTo(serverConfig.globalUpdateInterval, ::scheduleUpdateTask)
+        serverConfig.subscribeTo(serverConfig.globalUpdateInterval, ::scheduleUpdateTask, ignoreInitialValue = false)
```
schroda: The interval also gets started in the ServerConfig
I think I did it that way for the same reason as for the backup interval

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/backup/proto/ProtoBackupExport.kt
Syer10: It should, objects are statically initialized

Syer10: Though I see now that they are initialized in ServerSetup, so I'll revert these ones

schroda: I think I confused it with the Updater, because that only gets initialized lazily, so I just didn't change it like this when I added the server setting subscription

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
Syer10: I think it still is, configs would get initialized and there are some logs in the config manager that should be outputted

schroda (approved):

PR #666, opened 2023-08-23
Throw error instead of returning null

schroda:
In case e.g. no manga exists for the passed id, the query returned null. This makes it harder to have a "streamlined" error handling in the client, since these types of queries need a special handling to detect that the query returned an error.

Could be further improved to return custom errors, but for now, having a default error response is good enough


Syer10 (commented):
What error does this throw? I'm not too sure if just changing the generic is enough

schroda:
it will return an error that it is definend as non nullable but returns null

Syer10 (approved):

PR #671, opened 2023-08-28
Always return "ArchiveProvider" in case "downloadAsCbz" is enabled

### LINKED ISSUE #669: [Bug] First chapter not converted to CBZ format.
## Device information
- Tachidesk version: v0.7.0-r1355
- Server Operating System: Windows 10
- Server JVM version: java version "17.0.7" 2023-04-18 LTS
- Client Operating System: Windows 10 22H2
- Client Web Browser: Opera One (version : 101.0.4843.43)

## Steps to reproduce
1. In server.conf, enable download as cbz
2. Start tachidesk
3. Download a manga
4. While it downloads closes tachidesk
5. Restart tachidesk-server

### Expected behavior
Restarting tachidesk should retrieve the manga being downloaded and then put it in cbz at the end. 

### Actual behavior
It retrieves the manga being downloaded, but at the end does not put it as cbz and leaves the file as is.

## Other details
The bug only works on the chapter that was cut during download, on the next ones if there are they will be put in cbz.

schroda:
Due to reusing the "FolderProvider" to download a chapter as a cbz file, a normal chapter download folder was created.
In case the download was aborted before the cbz file got created and the folder deleted, the next time the chapter got downloaded, the wrong "FileProvider" was selected, causing the chapter not to get downloaded as a cbz file.

fixes #669

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
@@ -1,7 +1,16 @@
 package suwayomi.tachidesk.manga.impl.download.fileProvider
 
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.sample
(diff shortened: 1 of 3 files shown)


Syer10:
I'm not too sure about this, I would like to keep support for old folder downloads, and only use cbz when creating new downloads and using cbz archives. If this is merged we wouldn't be able to use old folder downloads, and people would have to re-download everything. Is it possible to download to a temporary folder instead and move the folder/compress it when its finished?

schroda:
makes sense, did not think about that use case

the chapter now gets downloaded in a temp folder and depending on the file provider it gets either saved as a normal folder or cbz file in the actual download location

Syer10 (approved):

PR #681, opened 2023-09-10
Feature/auto download ahead

schroda:
Adds the logic to "download ahead".
By using the "autoDownloadAheadLimit" setting, this logic downloads the latest unread and not downloaded chapters until the set limit is reached.

The logic does not get triggered automatically by the server, but instead, the client has to call the gql mutation.

To make this work, additional logic was added to the downloader and chapters.
Each chapter now has a "downloadReason" property, which indicates, what caused that chapter to get downloaded (e.g. user, download ahead).
This is also attached to the "DownloadChapter".
This makes it possible to only delete chapters that were added with a specific cause.

A different use case for this flag, besides the "download ahead" logic, could be to provide the option to automatically delete chapters after marking them as read.
While this should (only?) delete automatically downloaded chapters, it would also delete manually downloaded chapters.
To prevent this, the newly added flag can be used, to prevent automatically deleting manually downloaded chapters.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/DownloadMutation.kt
@@ -259,14 +259,18 @@ class DownloadMutation {
         }
     }
 
-    data class DownloadAheadInput(val clientMutationId: String? = null, val mangaIds: List<Int>)
+    data class DownloadAheadInput(
+        val clientMutationId: String? = null,
+        val mangaIds: List<Int> = emptyList(),
+        val latestReadChapterIds: List<Int>? = null
+    )
 
     data class DownloadAheadPayload(val clientMutationId: String?)
 
     fun downloadAhead(input: DownloadAheadInput): DownloadAheadPayload {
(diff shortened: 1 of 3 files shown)


Syer10:
I think that download ahead is something that the client should implement, not the server. This has a high chance to get messy when someone just wants to mark a bunch of chapters as read

schroda:
you're thinking of only triggering it when the chapters get marked as read through the reader?
because otherwise, I think it would just move the issue to the client, which would be not as good at handling it as the server would?
in theory the same issue would still be possible when doing it like this if the chapters are being skipped through in the reader, just less likely?

and at least I would want/expect it to trigger it everytime a chapter gets marked as read, no matter from where in the app that happens

I would prefer to keep it in the server and just optimize it
it could be updated to also remove chapters from the queue in case they get marked as read, but only in case they were added via the "download ahead" logic (some flag or type in the download chapter indicating what enqueued the download - e.g. user or "download ahead" or something else)
and to prevent it from being triggered too often, it could be "debounced" (collecting all manga ids of all calls) and only once some time has passed, trigger the logic

Syer10 (approved):

PR #684, opened 2023-09-16
Send every download status change to the subscriber

schroda:
Flow::stateIn has "Strong equality-based conflation" (see documentation). Thus, it omits every value in case it's equal to the previous one. Since the DownloadManger::getStatus function returns a status with a queue, that contains all current "DownloadChapters" by reference, the equality check was always true. Thus, progress changes of downloads were never sent to subscribers. Subscriber were only notified about finished downloads (size of queue changed) or downloader status changes


Syer10 (approved):
I would prefer to work with a StateFlow, but I see that DownloadChapter isn't a data class like I thought it was. We would need to rewrite it quite a bit for it to work like how I like it.

PR #688, opened 2023-09-22
remove @Synchronized in CloudflareInterceptor.kt for performance

ghost:
CloudflareInterceptor is a singleton. The intercept method of this interceptor is annotated with @Synchronized, which causes all network requests made using the cloudflareClient to be executed sequentially. Currently, CloudflareInterceptor only throws an exception when it encounters a Cloudflare block, so it is safe to remove the @Synchronized annotation.


Syer10 (approved):
Dam, no wonder performance takes a hit when using a bunch of network queries, thanks for catching this.

PR #712, opened 2023-10-14
Query for mangas in specific categories

schroda:


Syer10:
I feel like there would be more benefits if it was in the Filter instead of the Condition

schroda:
I thought I already added that

Syer10 (approved):

PR #720, opened 2023-10-16
add trackers support

### LINKED ISSUE #80: [Feature Request] Syncing Tachidesk reading status with Tracking services like MAL, Anilist etc
## Device information
- Tachidesk version: Docker [arbuild/tachidesk-docker:latest/slim](https://hub.docker.com/r/arbuild/tachidesk-docker)
- Server Operating System: Win 10 Home
- Server Desktop Environment: WSL2 Ubuntu 20.04
- Server JVM version: bundled with docker image 
- Client Web Browser: Google Chrome Version latest.

Currently we lack syncing with any tracking services. It would be good to have syncing your reading status to Mal, Anilist etc services. 

Features to have:

- Offline list with tags like reading/ dropped/ complete on hold etc and fields like rating, last read chapter, last read volume, read start date, completion date etc
- Compare and synchronise local list with tracking sites list
- Edit each services tracking links from manga page.

We can look over the source code of [trackma](https://github.com/z411/trackma) for guidance.

tachimanga:
Note: the Anilist & MyAnimeList clientId should be replaced with official clientId, and set callback url to suwayomi://oauth/anilist and suwayomi://oauth/myanimelist

MyAnimeListApi.kt 
AnilistApi.kt

close https://github.com/Suwayomi/Tachidesk-Server/issues/80


server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -132,5 +133,14 @@ object MangaAPI {
...
         }
+
+        path("track") {
+            get("list", TrackController.list)
+            post("login", TrackController.login)
+            post("logout", TrackController.logout)
+            post("search", TrackController.search)
+            post("bind", TrackController.bind)
+            post("update", TrackController.update)
+        }
```
Syer10: Do you think you would be able to add GraphQL endpoints for these?

tachimanga: Hi @Syer10, I'm sorry that I have very limited knowledge about GraphQL.

Syer10: I'll try to take a look at implementing it when I have time, I'll push to this branch so that it can be included in the PR.
We have deprecated the REST api and all future development is to take place in the GraphQL api(under the suwayomi.tachidesk.graphql package). So we need GraphQL endpoints for these before merging.

tachimanga: BTW, the related PR to Sorayomi can be found here: https://github.com/Suwayomi/Tachidesk-Sorayomi/pull/236

Syer10:
@tachimanga I did a bunch of cleanup as well as add GraphQL support. Can you verify that everything still works as intended?

One thing to note is that I changed the track/bind endpoint to add a `mangaId` query parameter since I removed it and other unused parameters from the TrackSearch model. So this will have to be accounted for in Sorayomi as well as other minor model changes I did.

tachimanga:
@Syer10  I am a Kotlin beginner, and my code often shows some Java flavor. I have learned a lot from your modifications, and I will make the necessary changes to Sorayomi to handle the endpoint changes. I will reply once I have finished the modifications.

tachimanga:
@Syer10 Everything is working great! It was a good idea to remove mangaId from the MangaTrackerDataClass. 
I have made the corresponding changes to the Sorayomi code. You can check the modifications in this commit: https://github.com/Suwayomi/Tachidesk-Sorayomi/pull/236/commits/2b227ab2a5705b3a9df12f6ebfa743fc3e10b2e7

Syer10:
Great! I'll look into the client ids in the next few days, we should be able to merge it sometime soon.

tachimanga:
@Syer10 Some notes about the callback URL.  
In the configuration for MyAnimeList and AniList, the callback URL must be a fixed value. Currently, the scheme for the callback URL is set as suwayomi://, which makes it easy for us to complete the OAuth process on the client-side. 
The flow is as follows: `MyAnimeList/Anilist -> Oauth Handler(client-side).`
However, if we need to implement the "add trackers" feature on the web UI, the scheme must be http(s):// to complete the process. Of course, we can use different client IDs for different front-ends. The client-side can still use suwayomi://, but the callback URL for the web UI cannot be set as a fixed value, such as http://127.0.0.1:4567/oauth/myanimelist, because the IP might not always be 127.0.0.1. Therefore, we need an intermediate page where users can enter the local IP to complete the OAuth process. 
The flow would be as follows: `MyAnimeList/Anilist -> Proxy Page (on the internet) -> Callback Page (on the local network).`

Syer10:
Hmm, I see.
@schroda do you think this would be possible?

AriaMoradi:
> The flow would be as follows: `MyAnimeList/Anilist -> Proxy Page (on the internet) -> Callback Page (on the local network).`

It might be possible to set it up via github pages? Does their oauth request accept additional arguments so we could pass it the real redirect url?

Alternatively this could be the first electron only feature in WebUI.

tachimanga:
@AriaMoradi  Github Pages is OK, and when the OAuth process is completed, it will jump back to a URL such as `suwayomi://oauth/myanimelist?code=def50200c056...` with no additional arguments.

If we implement the tracker setting in the web UI, we can set up a github page with the URL as `https://suwayomi.org/oauthproxy-anilist.html`. Then, when they jump back, they will be redirected to `https://suwayomi.org/oauthproxy-anilist.html?code=def50200c056....` on the proxy page, then we can redirect to `http://{local_ip_user_input}:4567/webui/oauth/anilist?code=def50200c056` to complete OAuth process.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/track/tracker/TrackerPreferences.kt
```diff
@@ -0,0 +1,69 @@
...
+    fun getTrackUsername(sync: Tracker) = preferenceStore.getString(trackUsername(sync.id), "")
+
+    fun getTrackPassword(sync: Tracker) = preferenceStore.getString(trackPassword(sync.id), "")
+
+    fun setTrackCredentials(
+        sync: Tracker,
+        username: String,
+        password: String,
+    ) {
+        logger.debug { "setTrackCredentials: id=${sync.id} username=$username password=$password" }
```
schroda: password gets logged in cleartext

Syer10:
I've decided to just merge it, and disable the endpoints until we have a working auth solution

renjfk:
There's also Auth Pin option from [AniList](https://anilist.gitbook.io/anilist-apiv2-docs/overview/oauth/getting-started) at least:

> If you are absolutely unable to have a Http or application URI scheme redirection you can use the Auth Pin redirect URL. After the user has authorized your application they will be redirected to a page requesting them to manually copy & paste the code/token into your application.
To enable this, on your AniList v2 client page, set your Client Redirect Uri to `https://anilist.co/api/v2/oauth/pin`.

renjfk:
Now that I think about it, alternatively the Suwayomi-Server could be transparent regarding OAuth client and the actual client e.g. WebUI should determine OAuth client therefore you could have multiple URI scheme redirections. 

So basically the original initiator gets the token if it's supported by the UI type, stores it to server and the server makes use of it amongsts different UIs.

You can think of it like, Suwayomi-Server will act like transparent proxy while fetching the token and then grab it.

taos15:
That is how it works, kind o. after a tracker is authenticated, the token get served in the server, and any client (even if they dont have a login implementations) can use it and make authenticated called to the tracker. A Pr I was working on give you another OATH2 implementation (authentication code), which let you set the redirect uri instead of using the hardcoded one, then the client can pick/implement what redirect uri will be use for authenticating. Since tracking has just been implemented in the webUI, I have put move the PR to low priority, but I am thinking on finish it in the future. The logic for the implementation can be reused to add oath2 authentication to secure suwayomi.

PR #729, opened 2023-10-19
Fetch mangas during the update

schroda:
The update only fetched the chapter list of a manga but never the manga itself. Thus, e.g. unless the manga got online fetched via the ui, it would never get recognized if it is completed or not. This would e.g. prevent the update setting, to not update completed mangas, from working as intended


Syer10:
Can you make this an option/parameter of the endpoint? In Tachiomi this is a optional setting that is off by default.

schroda:
added a setting

Syer10 (approved):

Syer10:
Ill merge this after #738 is merged and there is a graphql setting for it.

schroda:
done

PR #731, opened 2023-10-24
Fix/updater update stuck in running status after failure

schroda:
In case the channel failed due to an exception, the update for the source failed completely.
This however was never handled and the pending updates for the source were never set to failed.
Due to this, the global updates running state was always true

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
@@ -123,13 +123,18 @@ class Updater : IUpdater {
      */
     private fun updateStatus(
         jobs: List<UpdateJob>,
-        running: Boolean,
+        running: Boolean? = null,
         categories: Map<CategoryUpdateStatus, List<CategoryDataClass>>? = null,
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/update/Updater.kt
```diff
@@ -123,54 +124,85 @@ class Updater : IUpdater {
...
 
-    private suspend fun process(job: UpdateJob): List<UpdateJob> {
+    private fun handleChannelUpdateFailure(source: String) {
+        val isFailedSourceUpdate = { job: UpdateJob ->
+            val isForSource = job.manga.sourceId == source
+            val hasFailed = job.status == JobStatus.FAILED
+
+            isForSource && hasFailed
+        }
+        val failAllUpdatesForSource = {
```
Syer10: I don't think this Lamba is needed, can you remove it and just add a comment?

schroda: removed

Syer10 (approved):

PR #746, opened 2023-10-30
Fix/backup import failure not resetting status

schroda:


Syer10 (commented):
Is there a better way to do this? This is a breaking change to the backup models in the scheme

schroda:
not sure tbh

only having an  `Idle` status does not work for detecting if it failed or succeded.
and once the `Idle` status gets replaced with `Failure` and `Success` you can't reset to `Idle` anymore, since you have to pull for the restore status, and thus, setting it to `Idle` could lead to never getting the `Failure` or `Success` status

Syer10 (approved):

PR #750, opened 2023-11-02
Fix path to Preference file if it contains a invalid path character

Syer10:


AndroidCompat/src/main/java/xyz/nulldev/androidcompat/io/sharedprefs/JavaSharedPreferences.kt
```diff
@@ -34,7 +35,12 @@ class JavaSharedPreferences(key: String) : SharedPreferences {
         private val logger = KotlinLogging.logger {}
     }
 
-    private val file = Path(ApplicationRootDir, "settings", "$key.xml")
+    private val file =
+        Path(
+            ApplicationRootDir,
+            "settings",
+            "${SafePath.buildValidFilename(key)}.xml",
```
schroda, line 42: could we get key names that are too long and after taking the max allowed number of chars, we don't have a unique key anymore?
in case it is, what about splitting the path name at every "_" instead of only taking the max number of allowed chars?

e.g. for `some_random_key_name_that_is_too_long_..._the_end`  it would be `some/random/key/name/that/is/too/long/../the/end.xml` instead of `some_random_key_name_that_is_too.xml`

Syer10: Thats very unlikely. Android has the same limit as other OS's in regards to filename length, which effects the shared preferences filenames. I don't want to complicate it with a whole bunch of folders as well.

PR #751, opened 2023-11-03
Support all Tachiyomi downloaded chapter names

Syer10:


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/DirName.kt
```diff
@@ -31,21 +31,23 @@ private fun getMangaDir(mangaId: Int): String {
...
-        SafePath.buildValidFilename(
-            when {
-                chapterEntry[ChapterTable.scanlator] != null -> "${chapterEntry[ChapterTable.scanlator]}_${chapterEntry[ChapterTable.name]}"
-                else -> chapterEntry[ChapterTable.name]
-            },
-        )
-
-    return getMangaDir(mangaId) + "/$chapterDir"
+    return buildList {
+        if (!chapterEntry[ChapterTable.scanlator].isNullOrEmpty()) {
```
schroda, line 41: is there a difference in this and line 45?

Syer10: Hmm, previously I wanted to read downloads where there was an empty scanlator, but I fixed that in another PR. I'll remove it

PR #754, opened 2023-11-04
Chapter Fetch Improvements

Syer10:
- Only 1 fetch can happen at a time per manga
- Read and Bookmark statuses are ported over if a chapter gets a new url
- Updates list is not spammed with re-added chapters
- Manga title is removed from the chapter name automatically
- DateUpload is set to now if the source does not provide one

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
@@ -243,6 +243,7 @@ object Chapter {
                             this[ChapterTable.realUrl] = chapter.realUrl
                             this[ChapterTable.isRead] = false
                             this[ChapterTable.isBookmarked] = false
+                            this[ChapterTable.isDownloaded] = false
 
                             // is recognized chapter number
                             if (chapter.chapterNumber >= 0f && chapter.chapterNumber in deletedChapterNumbers) {
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
@@ -7,17 +7,21 @@ package suwayomi.tachidesk.manga.impl
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import com.google.common.cache.Cache
+import com.google.common.cache.CacheBuilder
 import eu.kanade.tachiyomi.source.model.SChapter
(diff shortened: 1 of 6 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -109,133 +117,170 @@ object Chapter {
...
-                    this[ChapterTable.fetchedAt] = it.fetchedAt
-                    this[ChapterTable.manga] = it.mangaId
-                    this[ChapterTable.realUrl] = it.realUrl
+                    if (chapterEntry == null) {
+                        val newChapterData =
+                            if (chapterData.uploadDate == 0L) {
+                                val altDateUpload = if (maxSeenUploadDate == 0L) now else maxSeenUploadDate
+                                chapterData.copy(uploadDate = altDateUpload)
+                            } else {
+                                maxSeenUploadDate = max(maxSeenUploadDate, chapterData.uploadDate)
```
schroda, line 186: not needed due to line 189

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -109,133 +117,170 @@ object Chapter {
...
+
+                // clear any orphaned/duplicate chapters that are in the db but not in `chapterList`
+                val dbChapterCount = chaptersInDb.count()
+                if (dbChapterCount > chapterList.size) { // we got some clean up due
+                    val chapterUrls = chapterList.map { it.url }.toSet()
+
+                    val chaptersIdsToDelete =
+                        chaptersInDb.mapNotNull { dbChapter ->
+                            if (!chapterUrls.contains(dbChapter.url)) {
+                                if (dbChapter.read) deletedReadChapterNumbers.add(dbChapter.chapterNumber)
```
schroda, line 215: are downloads not possible to be preserved?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
Syer10: Thanks, the one on 189 isn't supposed to be there.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
Syer10: Added this, I am a bit hesitant due to chapter downloads relying on chapter names, and if they are renamed it won't work. But since we don't handle anything like that yet I added it as is.

schroda (approved):

PR #775, opened 2023-11-19
Add mutation to clear the cached images

schroda:


server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ImageMutation.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.kodein.di.DI
+import org.kodein.di.conf.global
+import org.kodein.di.instance
+import suwayomi.tachidesk.manga.impl.util.storage.ImageResponse
+import suwayomi.tachidesk.server.ApplicationDirs
+
+private val applicationDirs by DI.global.instance<ApplicationDirs>()
+
+class ImageMutation {
+    data class ClearCachedImagesInput(
```
Syer10, line 12: Needs clientMutationId

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ImageMutation.kt
schroda: fixed

Syer10 (approved):

PR #781, opened 2023-11-26
Fix/gql about webui query same response type as webui update info

schroda:


server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/InfoMutation.kt
```diff
@@ -32,6 +33,8 @@ class InfoMutation {
                 val (version, updateAvailable) = WebInterfaceManager.isUpdateAvailable()
 
                 if (!updateAvailable) {
+                    val didUpdateCheckFail = version === ""
```
Syer10:
```suggestion
                    val didUpdateCheckFail = version.isEmpty()
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/InfoMutation.kt
schroda: oh yeah, wrong language 😅

Syer10 (approved):

PR #784, opened 2023-12-01
Changed format for downloaded CBZ files. Added manga title to help scrapers…

benticlaskaris:
Changed format for downloaded CBZ files. Added manga title to help scrapers and manga organizer like komga,kavita,etc...

Best would be to offer a format configuration but even adding the manga title to the file works correctly.


schroda:
existing downloads need to be updated

Syer10:
Thanks for the suggestion, but I am going to deny this. We plan to keep the downloads using the same scheme as Tachiyomi. Tachidesk is not meant to be a downloader that you use to copy to other places, its mainly meant to be an all-in-one application, so the naming scheme is based on our needs.

If you want to copy them elsewhere, your best chance is to create a script that uses the api, that matches the chapters to the downloaded files, and then copies them elsewhere based on your naming scheme.

PR #809, opened 2024-01-09
Fix/update tachiyomi extension repo

schroda:


schroda:
will get removed completely

PR #829, opened 2024-01-15
Add mutex to "updateExtensionDatabase"

schroda:
If called in quick succession it is possible that duplicated extensions get inserted to the database, because it has not yet been updated by the first call


Syer10 (approved):

chancez:
Ah this explains some weird stuff I noticed recently during development. 👍

PR #832, opened 2024-01-17
Add support for configuring which categories are downloaded automatically

chancez:
Implementation is similar to the global update categories. This allows you to configure which manga you want to auto download based on their categories. The default behavior remains the same. If no categories are configured, then automatic downloads will behave the same as prior to this change, which is to say, all manga in your library are considered for automatic downloads.

The idea is that you might have `server.excludeEntryWithUnreadChapters=false` but don't necessarily want all your manga to get downloaded when new chapters come out. 

For example; if you've started a manga, but are taking a break, or if you finished a manga that has some "extra" chapters that come out later, or something. For a setup like this, I have a "currently reading" category, and a "backlog" and "completed" category". I'd set "currently reading" to be included, and the others to be excluded, ensuring only manga in my "currently reading" get auto downloaded as new chapters are released.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
@@ -37,6 +37,7 @@ import suwayomi.tachidesk.manga.impl.download.DownloadManager.EnqueueInput
 import suwayomi.tachidesk.manga.impl.track.Track
 import suwayomi.tachidesk.manga.impl.util.source.GetCatalogueSource.getCatalogueSourceOrStub
 import suwayomi.tachidesk.manga.model.dataclass.ChapterDataClass
+import suwayomi.tachidesk.manga.model.dataclass.IncludeOrExclude
 import suwayomi.tachidesk.manga.model.dataclass.MangaChapterDataClass
 import suwayomi.tachidesk.manga.model.dataclass.PaginatedList
 import suwayomi.tachidesk.manga.model.dataclass.paginatedFrom
@@ -49,6 +50,7 @@ import suwayomi.tachidesk.server.serverConfig
 import java.time.Instant
 import java.util.TreeSet
 import java.util.concurrent.TimeUnit
+import kotlin.collections.listOf
 import kotlin.math.max
 
 object Chapter {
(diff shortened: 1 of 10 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -309,17 +310,52 @@ object Chapter {
...
+        if (!serverConfig.autoDownloadNewChapters.value) {
+            log.debug { "automatic download is not configured" }
+            return
+        }
+
+        // Verify the manga is configured to be downloaded based on it's categories.
+        var mangaCategories = CategoryManga.getMangaCategories(mangaId)
+        // if the manga has no categories, skip checking the categories configuration for automatic
+        // download
+        if (mangaCategories.isNotEmpty()) {
```
schroda, line 348: if a manga has no category it is considered to be in the default category, so you have to check its `includeInDownload` flag

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -309,17 +310,52 @@ object Chapter {
...
-        // make sure to ignore initial fetch
-        val isDownloadPossible =
-            serverConfig.autoDownloadNewChapters.value && areNewChaptersAvailable && !wasInitialFetch
-        if (!isDownloadPossible) {
-            log.debug { "download is not allowed/possible" }
+        if (!areNewChaptersAvailable) {
+            log.debug { "no new chapters available" }
+            return
+        }
+
```
schroda, line 330: I'd move these guards up to prevent unnecessary logic to get executed

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
chancez: Interesting. What's the ideal way to approach that? Should `CategoryManga.getMangaCategories` be updated to return the default category in the list, or should I query for it using `Category.getCategoryById(Category.DEFAULT_CATEGORY_ID)`?

Syer10: You should get the default category manually, CategoryManga.getMangaCategories is used in other places and we don't want to break behaviour

Syer10 (approved):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -309,20 +311,65 @@ object Chapter {
...
+            // use the unset categories as the included categories if the included categories is
+            // empty
+            val includedCategories = downloadCategoriesMap[IncludeOrExclude.INCLUDE].orEmpty().ifEmpty { unsetCategories }
+            val excludedCategories = downloadCategoriesMap[IncludeOrExclude.EXCLUDE].orEmpty()
+            // This set contains all of the categories that should be downloaded.
+            val categoriesToDownload = includedCategories.subtract(excludedCategories)
+            // Now look to see if the manga has any categories in the categoriesToDownload, if so,
+            // then this manga should have it's chapters downloaded
+            val mangaDownloadCategories = mangaCategories.toSet().intersect(categoriesToDownload)
+            if (mangaDownloadCategories.isNotEmpty()) {
```
schroda, line 363: we should only download a manga if it is not in an excluded category, even if it is in an included one

chancez: That's how it's implemented, line 359 I subtract the excluded categories from the included categories.

Syer10: Thats not how it works though, categories wont be both included and excluded, so they can't be subtracted

chancez: Ahh right.

chancez: Alright, updated.

Syer10: I think its still wrong, if the manga has *any* excluded category, it should not be automatically downloaded

chancez: Oh, right, so I probably need to subtract first, before intersecting

Syer10: Just do a `if (excludedCategories.isNotEmpty()) return`

chancez: `excludedCategories` is not the _manga_ excluded categories here, it's the full set of excluded categories, so to do that I have to refactor things a bit. Let me test it.

Syer10: So it would be something like `if (excludedCategories.any { it.id in mangaCategories }) return` wouldn't it?

chancez: Yeah that looks right, I was going to use `intersect` here too but your approach is a bit simpler

chancez: Ok, I added some logic to check if the manga's categories is in the list of excluded categories, and it logs them if there were any before returning. I tested with a manga in two categories, one include, one exclude and got the expected behavior.

schroda (approved):

PR #853, opened 2024-01-25
Use a session with flaresolverr

chancez:
I hardcoded the session name, and TTL but we could make them configurable in the future, but this alone makes a huge difference in reliability, and performance.


Syer10 (approved):
Great!

PR #856, opened 2024-01-27
Check download categories in downloadAheadChapters

chancez:
Obey download categories configuration in download ahead feature. I missed handling this in the PR introducing download the categories setting: https://github.com/Suwayomi/Suwayomi-Server/pull/832.

I opted to not do this in the download manager because users should still be able to download chapters if they decide to explicitly download them, rather than during an update/while reading.


Syer10:
@chancez @schroda I saw a conversation about this yesterday, is this still relevant? Ideally we should do the same as Mihon/Tachiyomi.

PR #882, opened 2024-02-18
Seperate out migrations to allow run-once migrations

Syer10:
Will fix log-spam that sometimes happen on linux


server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -298,3 +275,47 @@ fun migratePreferences(
...
+        // Migrate from old preferences api
+        val prefRootNode = "suwayomi/tachidesk"
+        val isMigrationRequired = Preferences.userRoot().nodeExists(prefRootNode)
+        if (isMigrationRequired) {
+            val preferences = Preferences.userRoot().node(prefRootNode)
+            migratePreferences(null, preferences)
+            preferences.removeNode()
+        }
+    }
+    migrationPreferences.edit().putInt("version", MIGRATION_VERSION).apply()
```
AriaMoradi: should probably move the code to a separate file

PR #898, opened 2024-03-05
Always fetch thumbnail of manga from local source

schroda:
The local manga thumbnail got "downloaded" to thumbnail download folder of in library manga. Since the "thumbnail url" of a local source manga never changes, the "downloaded" manga thumbnail never got updated

Regression introduced with f2dd67d87f38c30c8df6f3718ce392197afbff9a

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
@@ -315,7 +315,7 @@ object Manga {
     suspend fun getMangaThumbnail(mangaId: Int): Pair<InputStream, String> {
         val mangaEntry = transaction { MangaTable.select { MangaTable.id eq mangaId }.first() }
 
-        if (mangaEntry[MangaTable.inLibrary]) {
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Manga.kt
```diff
@@ -315,7 +315,7 @@ object Manga {
     suspend fun getMangaThumbnail(mangaId: Int): Pair<InputStream, String> {
         val mangaEntry = transaction { MangaTable.select { MangaTable.id eq mangaId }.first() }
 
-        if (mangaEntry[MangaTable.inLibrary]) {
+        if (mangaEntry[MangaTable.inLibrary] && getCatalogueSourceOrStub(mangaEntry[MangaTable.sourceReference]) !is LocalSource) {
```
Syer10:
```suggestion
        if (mangaEntry[MangaTable.inLibrary] && mangaEntry[MangaTable.sourceReference] != LocalSource.ID) {
```

Syer10 (changes requested):

Syer10 (approved):

PR #915, opened 2024-03-29
fix: adds condition to the tracker update mutation to update the db to match the tracker last chapter read tracker when none is passsed

taos15:
- Updates the database to match the tracker last chapter read when that field is not been change in the update mutation. [Issue: 916](https://github.com/Suwayomi/Suwayomi-Server/issues/916)

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/TrackType.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/TrackType.kt
@@ -60,7 +60,7 @@ class TrackStatusType(
     val name: String,
 )
 
-        class TrackRecordType(
+class TrackRecordType(
     val id: Int,
     val mangaId: Int,
     val trackerId: Int,
@@ -74,8 +74,8 @@ class TrackStatusType(
     val remoteUrl: String,
     val startDate: Long,
     val finishDate: Long,
-    ) : Node {
-        constructor(row: ResultRow) : this(
+) : Node {
+    constructor(row: ResultRow) : this(
         row[TrackRecordTable.id].value,
         row[TrackRecordTable.mangaId].value,
         row[TrackRecordTable.trackerId],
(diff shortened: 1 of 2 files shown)


taos15:
Since the tracker is the source of true, when we are not changing the last chapter read the database should be updated to match the tracker.

taos15 (commented):
I cleaned up the previous code. Only a call to the tracker refresh method was needed. now when a mutation that does not change the last chapter read is sent, it will check the tracker before making any change ```graphql
mutation MyMutation {
  updateTrack(input: {recordId: 1}) {
    trackRecord {
      lastChapterRead
      trackerId
      tracker {
        name
      }
    }
  }
}```

taos15 (commented):
> I cleaned up the previous code. Only a call to the tracker refresh method was needed. now when a mutation that does not change the last chapter read is sent, it will check the tracker before making any change 
```graphql
mutation MyMutation {
  updateTrack(input: {recordId: 1}) {
    trackRecord {
      lastChapterRead
      trackerId
      tracker {
        name
      }
    }
  }
}```

schroda:
imo this should be done in a separate mutation, at least to me it is not clear from looking at the available mutations that `updateTrack` will fetch the latest data from the tracker in case it's called with only the `remoteId`

taos15:
Actually it gets the latest every time. This does not address how to address the refresh of the tracker, it fox how the update itself works. Currently the update mutation is comparing to all data, this male sure it compare to the most resent data. I agree that another migration just to refresh "could" be beneficial, but that should be a separate issues. This address the current bug in the update mutation, we have the logic to check that the input.lastRead has to be greater to change the current last read chapter, but because it does not fetch the data, that logic is not doing what we expect, so it behave as there is not check at all.

schroda:
these changes do not work at all as intended, they just prevent the server from sending the actual update input at all because it will just fetch the latest data from the tracker and resend the just fetched data to the tracker again as an update

taos15:
Closing the PR since core maintaner already have [PR 917](https://github.com/Suwayomi/Suwayomi-Server/pull/917) where they are working on it.

PR #938, opened 2024-04-27
Prevent greater last page read than page count

schroda:
In case multiple chapters are getting updated, the last page read might be higher than the available pages of a chapter


Syer10:
What about chapters where we haven't fetched the pages so the pagecount is -1?

schroda:
> What about chapters where we haven't fetched the pages so the pagecount is -1?

it would have broken coerceIn...
lastPageRead will now be 0 in that case

Syer10 (approved):

PR #954, opened 2024-06-02
Support partial mutation responses

schroda:
In case e.g. a mutation was made which looked like this

myMutation {
  mutationA { ... }
  mutationB { ... }
  mutationC { ... }
}

and mutation A and B succeeded while mutation C failed, the response only included the error of C and the successful mutation data response of A and B was missing

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/AsDataFetcherResult.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/AsDataFetcherResult.kt
@@ -1,0 +1,25 @@
+package suwayomi.tachidesk.graphql
+
+import com.expediagroup.graphql.server.extensions.toGraphQLError
+import graphql.execution.DataFetcherResult
+import mu.KotlinLogging
+
+val logger = KotlinLogging.logger { }
+
+inline fun <T> asDataFetcherResult(block: () -> T): DataFetcherResult<T?> {
+    val result =
+        runCatching {
+            block()
+        }
+
+    if (result.isFailure) {
+        logger.error(result.exceptionOrNull()) { "asDataFetcherResult: failed due to" }
+        return DataFetcherResult.newResult<T?>()
+            .error(result.exceptionOrNull()?.toGraphQLError())
+            .build()
+    }
+
+    return DataFetcherResult.newResult<T?>()
+        .data(result.getOrNull())
+        .build()
+}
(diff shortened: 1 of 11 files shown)


server/src/main/kotlin/suwayomi/tachidesk/graphql/AsDataFetcherResult.kt
```diff
@@ -0,0 +1,23 @@
...
+import mu.KotlinLogging
+
+inline fun <T> asDataFetcherResult(block: () -> T): DataFetcherResult<T?> {
+    val result =
+        runCatching {
+            block()
+        }
+
+    if (result.isFailure) {
+        KotlinLogging.logger { }.error(result.exceptionOrNull()) { "asDataFetcherResult: failed due to" }
```
Syer10: Bring KotlinLogging.logger {} outside the function, we don't need to recreate the logger every time

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt
```diff
@@ -89,36 +91,40 @@ class ChapterMutation {
         }
     }
 
-    fun updateChapter(input: UpdateChapterInput): UpdateChapterPayload {
-        val (clientMutationId, id, patch) = input
+    fun updateChapter(input: UpdateChapterInput): DataFetcherResult<UpdateChapterPayload?> {
```
Syer10, line 94: Is making everything nullable changing the scheme?

server/src/main/kotlin/suwayomi/tachidesk/graphql/AsDataFetcherResult.kt
schroda: done

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt
schroda: yes, the result types will be nullable

it also changes the response in case of an error
instead of only returning the error and having an undefined data property, the `data` object will be present with the actual mutation result object being null

e.g.
response when providing an invalid manga id for the `updateManga` mutation
```
{
  "data": {
    "updateManga": null
  },
  "errors": [
    {
      "message": "Collection is empty."
    }
  ]
}
```

in case of the example in the pr description

```
{
  data: {
    mutationA: result // success
    mutationB: result // success
    mutationC: null // failure
  },
 errors: [{ message: "some error"  }]
}
```

instead of 
```
{
 errors: [{ message: "some error"  }]
}
```

Syer10 (approved):

PR #960, opened 2024-06-08
Feature/gql simpilify filtering for multiple values

schroda:
this is a breaking change to the gql schema

Makes it easier to filter for multiple values at ones without having to nest filters with multiple "and".

e.g.

```gql
query MyQuery {
 mangas(
  filter: {genre: {includesInsensitive: "action"}, and: {genre: {includesInsensitive: "adventure"}, and: { ... }}}
 ) {
  nodes {
   id
  }
 }
}
```

can be simplified to

```gql
query MyQuery {
 mangas(
  filter: {genre: {includesInsensitive: ["action", "adventure", ...]}}
 ) {
  nodes {
   id
  }
 }
}
```

Makes it easier to filter for entries that match any value without having to nest filters with multiple "or".

e.g.

```gql
query MyQuery {
 mangas(
  filter: {genre: {includesInsensitiveAny: ["action", "adventure", ...]}}
 ) {
  nodes {
   id
  }
 }
}
```

instead of

```gql
query MyQuery {
 mangas(
  filter: {genre: {includesInsensitive: "action", or: {genre: includesInsensitive: "adventure", or: {...}}}}
 ) {
  nodes {
   id
  }
 }
}
```


Syer10:
Hmm, I'm not sure about this. Its already possible to do with with a more involved manner, and if this breaks those queries I'm not sure its worth it.

schroda:
in itself the queries still work, the values would just have to be wrapped in a list (which will depend on the client, since graphql doesn't care if a single value is wrapped in a list or not)

but it's only for convenience, since I think it's kinda ugly having to create these nested filters
if it's only about the breaking change, it could also be added as a new filter like `<filter>All` (e.g. `includeAll`) leaving the current filter (`include`) as is

Syer10:
I like the idea of it being a new filter. It wouldn't be a scheme break and it would add new options for those who need it

Syer10 (commented):
A lot of the `not` versions are still changed, will those be reverted?

schroda:
yeah, i've missed them, everything should be fine now

Syer10:
Can you add a any option for the exclude filters? Would be more useful for exclude then All

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/filter/Filter.kt
```diff
@@ -451,15 +520,8 @@ fun <T : Comparable<T>> andFilterWithCompareEntity(
...
-    val opAnd = OpAnd(andFilterEntity(column, filter))
-
-    opAnd.andWhere(filter.lessThan) { column less it }
-    opAnd.andWhere(filter.lessThanOrEqualTo) { column lessEq it }
-    opAnd.andWhere(filter.greaterThan) { column greater it }
-    opAnd.andWhere(filter.greaterThanOrEqualTo) { column greaterEq it }
-
-    return opAnd.op
+    @Suppress("UNCHECKED_CAST")
+    return andFilterWithCompare(column as Column<T>, filter)
```
Syer10, line 551: Huh, this cast works? Wouldn't it fail because T is wrapped by an entity

schroda: I assume it works because it has to be a column with a comparable value, which EntityID is

Syer10: This is a simple query to test it, let me know if it works
```gql
query GetManga {
    mangas(filter: { id: { in: [1, 2] } }) {
        id
    }
}
```

schroda: it's working (I wouldn't have created the pr otherwise 😄)

from googling: it works because of generic type erasure, so at runtime it's just Column to Column

Syer10 (approved):

PR #984, opened 2024-06-28
Cleanup only created auto backup files

schroda:
The automated backup cleanup just deleted every file (recursively in subfolders as well) in the set folder in case it was older than the set backup ttl. This made it impossible to save the automated backups into a folder with different files.


Syer10:
I think would would be better to filter for .tachibk files instead of storing a list of old backups

schroda:
what if there are non automated or mihon backups in the same folder that are not supposed to be cleaned up?
wouldn't this then basically be the same issue again as the person from discord had with unintended files getting deleted?

Syer10:
In Mihon there is one main backup folder, and a subfolder for auto backups. Maybe we can do that instead
Like
```
> backups
    -> auto
```

Syer10:
We can alternatively change it so we add auto to the filename and do something like `filename.startsWith("org.suwayomi.tachidesk.auto")` filter as well, to specify only Suwayomi backups

Robonau:
so unexpectedly delete any files form a possible(pretty reasonable tbh) sub folder?

schroda:
> We can alternatively change it so we add auto to the filename and do something like filename.startsWith("org.suwayomi.tachidesk.auto") filter as well, to specify only Suwayomi backups

done

Syer10 (approved):

PR #990, opened 2024-07-15
Add setting to use the flaresolverr response

AeonLucid:
In some cases the Suwayomi server receives a cloudflare challenge but the flaresolverr does not.
When this happens and this setting is configured, it passes the response from flaresolverr back to suwayomi.

Sadly this does not work for images because chrome render a small html template which is what flaresolverr responds with, instead of the actual image data.

### DIFF
--- a/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
+++ b/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
@@ -23,6 +23,7 @@ import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
 import okhttp3.Response
+import okhttp3.ResponseBody.Companion.toResponseBody
 import suwayomi.tachidesk.server.serverConfig
 import uy.kohesive.injekt.injectLazy
 import java.io.IOException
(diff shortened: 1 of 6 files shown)


server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -56,11 +57,38 @@ class CloudflareInterceptor(
             originalResponse.close()
             // network.cookieStore.remove(originalRequest.url.toUri())
 
-            val request =
+            val flareResponseFallback = serverConfig.flareSolverrAsResponseFallback.value
+            val flareResponse =
                 runBlocking {
-                    CFClearance.resolveWithFlareSolverr(setUserAgent, originalRequest)
+                    CFClearance.resolveWithFlareSolver(originalRequest, !flareResponseFallback)
```
Syer10, line 63: Why was setUserAgent removed?

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -56,11 +57,38 @@ class CloudflareInterceptor(
...
+                            .code(flareResponse.solution.status)
+                            .body(flareResponse.solution.response.toResponseBody())
+                            .build()
+                    } else {
+                        logger.debug { "FlareSolverr response is an image html template, not falling back" }
+                    }
+                }
+            }
+
+            val request = CFClearance.requestWithFlareSolverr(flareResponse, setUserAgent, originalRequest)
```
Syer10, line 90: I don't really like duplicating requests to FlareSolverr

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
AeonLucid: It is not removed but simply moved elsewhere. In the case of `flareResponseFallback` being set it's called when the flareSolver browser received a succesful response. Otherwise it's applied in `requestWithFlareSolverr` as normal.

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
AeonLucid: It's not sending duplicate requests, the method name might be causing confusion here. This method has been changed to only craft the new request. It re-uses the flaresolver response from the `resolveWithFlareSolver` call.

Syer10: I see, looks good then

Syer10 (approved):

PR #994, opened 2024-07-25
Fix/downloaded chapters with lost page count

schroda:


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -265,7 +271,16 @@ object Chapter {
...
-                                this[ChapterTable.isDownloaded] = chapter.chapterNumber in deletedDownloadedChapterNumbers
+
+                                // only preserve download status for chapters of the same scanlator, otherwise,
+                                // the downloaded files won't be found anyway
+                                val downloadedChapterInfo = deletedDownloadedChapterNumberInfoMap[chapter.chapterNumber]
+                                val wasDownloaded = downloadedChapterInfo?.contains(chapter.scanlator) ?: false
+                                if (wasDownloaded) {
+                                    this[ChapterTable.isDownloaded] = true
+                                    // in case the chapter was downloaded it should not be possible that this is null
+                                    this[ChapterTable.pageCount] = downloadedChapterInfo!![chapter.scanlator]!!
```
Syer10: Would probably be better to just try to get the value instead of a check then get
```suggestion
                                val downloadedChapterInfo = deletedDownloadedChapterNumberInfoMap[chapter.chapterNumber]
                                val pageCount = downloadedChapterInfo?.get(chapter.scanlator)
                                if (pageCount != null) {
                                    this[ChapterTable.isDownloaded] = true
                                    this[ChapterTable.pageCount] = pageCount
```

Syer10 (approved):

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0039_FixDownloadedChaptersWithoutPageCount.kt
Syer10, line 1: Change this to M0041 and I will merge all the PRs at once 👍

schroda: done

PR #1011, opened 2024-08-21
Feature/graphql download queue subscription send only updates

schroda:
the old download changed subscription has been marked as deprecated and a new subscription has been added

the new subscription sends only changed downloads instead of the full status with every message

the type includes the following fields

- the `initial` download queue within the initial message **(n+1)**
- `updates` for
  - en- and dequeuing  **(n+1 without combination of omitting updates)**
  - repositioning downloads
  - errors
  - finished downloads
- an option `maxUpdates` to omit updates by setting a max update limit to prevent n+1 problem - includes a flag `omittedUpdates` which indicates that the `downloadStatus` queue should get re-fetched

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -142,7 +142,7 @@ object DownloadManager {
     val status = statusFlow.onStart { emit(getStatus()) }
 
     private val updatesFlow = MutableSharedFlow<DownloadUpdates>()
-    val updates = updatesFlow.onStart { emit(getDownloadUpdates()) }
+    val updates = updatesFlow.onStart { emit(getDownloadUpdates(addInitial = true)) }
 
     init {
         scope.launch {
@@ -200,7 +200,7 @@ object DownloadManager {
             downloadQueue.toList(),
         )
 
-    private fun getDownloadUpdates(): DownloadUpdates {
+    private fun getDownloadUpdates(addInitial: Boolean = false): DownloadUpdates {
         return DownloadUpdates(
             if (downloadQueue.none { it.state == Downloading }) {
                 Status.Stopped
@@ -208,6 +208,7 @@ object DownloadManager {
                 Status.Started
             },
             downloadUpdates.toList(),
+            if (addInitial) downloadQueue.toList() else null,
         )
     }
 
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -200,8 +200,8 @@ object DownloadManager {
             downloadQueue.toList(),
         )
 
-    private fun getDownloadUpdates(addInitial: Boolean = false): DownloadUpdates {
-        return DownloadUpdates(
+    private fun getDownloadUpdates(addInitial: Boolean = false): DownloadUpdates =
+        DownloadUpdates(
             if (downloadQueue.none { it.state == Downloading }) {
                 Status.Stopped
             } else {
@@ -210,7 +210,6 @@ object DownloadManager {
             downloadUpdates.toList(),
             if (addInitial) downloadQueue.toList() else null,
         )
-    }
 
     private val downloaderWatch = MutableSharedFlow<Unit>(extraBufferCapacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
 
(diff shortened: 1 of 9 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -119,8 +126,8 @@ object DownloadManager {
 
     private val notifyFlow = MutableSharedFlow<Unit>(extraBufferCapacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
 
-    private val statusFlow = MutableSharedFlow<DownloadStatus>()
-    val status = statusFlow.onStart { emit(getStatus()) }
+    private val updatesFlow = MutableSharedFlow<DownloadUpdates>()
+    val updates = updatesFlow.onStart { emit(getDownloadUpdates()) }
```
Syer10: Shouldn't this be getStatus so it sends all current downloads the first time, so the client can cache and use them?
```suggestion
    val updates = updatesFlow.onStart { emit(getStatus()) }
```

schroda: the initial subscription is one of the main reasons for this change, it causes the n+1 problem, because the subscription does not support data loader batching

to get the full status the download status query should be used and then after that, the subscription informs about changes

Robonau: relying on two separate requests to link up perfectly seems like a bad idea

schroda: I don't see how that is bad, the subscription is there to inform about changes and not to push the whole status everytime the progress of a single download has changed

you could even ignore the whole download queue and just use the updates from the subscription to show what is actively downloading

apollo is also using this as an example on when to use subscriptions https://www.apollographql.com/docs/react/data/subscriptions/#when-to-use-subscriptions -> Small, incremental changes to large objects

Robonau: if the query responds with ```
[
    {id:1234, status:"running"},
    {id:1235, status:"idle"}
]```
but by the time the server deals with the incoming web-socket if 1234 has finished i'tl forever be stuck in the queue on the client

inversely if you do the web-socket first you don't know if the data from the web-socket or the query is more up to date if there is a difference

Robonau: and if we are talking about only sending pieces when something changes then it would make sense to send the whole queue at first
since it is functionally changing from [] (an empty array) to the entire queue on the client
so the changes are the entire queue

Robonau: > not to push the whole status everytime the progress of a single download has changed

i don't see why you are saying this? nobody as said we want this, we already have this currently.

what syer suggested and what would fix the issue i mentioned is having the one web-socket not a web-socket and a query.
 have that one web-socket deal with both the initial full list and the updates after the initialisation.
 if you really want the initial list to be optional you can make it a separate field in the gql
 
 ```gql
subscription MySubscription {
  newdownloadChanged {
    initial {
      state
      tries
      progress
      manga {
        id
      }
      chapter {
        id
      }
    }
    updates {
      state
      tries
      progress
      manga {
        id
      }
      chapter {
        id
      }
    }
  }
}
```

where initial is null in every message but the first

schroda: > if the query responds with  [ {id:1234, status:"running"}, {id:1235, status:"idle"} ]
but by the time the server deals with the incoming web-socket if 1234 has finished i'tl forever be stuck in the queue on the client
inversely if you do the web-socket first you don't know if the data from the web-socket or the query is more up to date if there is a difference

which I'd say is a rare edge case with a relatively low chance of occurring

> and if we are talking about only sending pieces when something changes then it would make sense to send the whole queue at first
since it is functionally changing from [] (an empty array) to the entire queue on the client
so the changes are the entire queue

the changes are from the servers pov and not the clients?

> i don't see why you are saying this? nobody as said we want this, we already have this currently.

I don't get what you're saying here

> what syer suggested and what would fix the issue i mentioned is having the one web-socket not a web-socket and a query.

it's literally the main reason why I've made this change, the bigger the queue is the worse the responsiveness of the server gets, to literally making it unusable

> the initial subscription is one of the main reasons for this change, it causes the n+1 problem, because the subscription does not support data loader batching

I rather take some unlikely edge case for using a query + subscription compared to this

> if you really want the initial list to be optional

I don't want it at all, neither required nor optional - but I also don't care if something like this needs to be added

schroda:
the download subscription update now includes

- the `initial` download queue within the initial message **(n+1)**
- `updates` for
  - en- and dequeuing  **(n+1 without combination of omitting updates)**
  - repositioning downloads
  - errors
  - finished downloads
- an option `maxUpdates` to omit updates by setting a max update limit to prevent n+1 problem - includes a flag `omittedUpdates` which indicates that the `downloadStatus` queue should get re-fetched

Syer10:
@schroda there seems to be conflicts now

Syer10 (approved):

PR #1021, opened 2024-08-29
feat(comicinfo): add date fields to comic info

Belphemur:
This will be parsed by Komga, Kavita etc ... and any other library management to also have the date of the chapter.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/GetComicInfo.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/GetComicInfo.kt
@@ -15,46 +15,56 @@ import suwayomi.tachidesk.manga.model.table.MangaTable
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 import java.nio.file.Path
+import java.time.Instant
+import java.time.ZoneId
 import kotlin.io.path.deleteIfExists
 import kotlin.io.path.div
 import kotlin.io.path.outputStream
 
-/**
- * Creates a ComicInfo instance based on the manga and chapter metadata.
- */
 fun getComicInfo(
     manga: ResultRow,
     chapter: ResultRow,
     chapterUrl: String,
     categories: List<String>?,
-) = ComicInfo(
-    title = ComicInfo.Title(chapter[ChapterTable.name]),
-    series = ComicInfo.Series(manga[MangaTable.title]),
-    number =
-        chapter[ChapterTable.chapter_number].takeIf { it >= 0 }?.let {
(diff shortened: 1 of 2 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/GetComicInfo.kt
```diff
@@ -15,46 +15,56 @@ import suwayomi.tachidesk.manga.model.table.MangaTable
...
+            ),
+        categories = categories?.let { ComicInfo.CategoriesTachiyomi(it.joinToString()) },
+        inker = null,
+        colorist = null,
+        letterer = null,
+        coverArtist = null,
+        tags = null,
+        day = localDate?.dayOfMonth.let { it?.let { it1 -> ComicInfo.Day(it1) } },
+        month = localDate?.monthValue.let { it?.let { it1 -> ComicInfo.Month(it1) } },
+        year = localDate?.year.let { it?.let { it1 -> ComicInfo.Year(it1) } },
```
Syer10:
```suggestion
        day = localDate?.dayOfMonth?.let { ComicInfo.Day(it) },
        month = localDate?.monthValue?.let { ComicInfo.Month(it) },
        year = localDate?.year?.let { ComicInfo.Year(it) },
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/GetComicInfo.kt
```diff
@@ -15,46 +15,56 @@ import suwayomi.tachidesk.manga.model.table.MangaTable
...
-    letterer = null,
-    coverArtist = null,
-    tags = null,
-)
+): ComicInfo {
+    val dateUpload = chapter[ChapterTable.date_upload]
+    val localDate =
+        dateUpload.let {
+            Instant.ofEpochMilli(it).atZone(ZoneId.systemDefault()).toLocalDate()
+        }
```
Syer10:
```suggestion
        Instant.ofEpochMilli(dateUpload).atZone(ZoneId.systemDefault()).toLocalDate()
```
Please validate that we store them as Milliseconds as well, we store a lot of things as Seconds instead so this might need to be changed

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/GetComicInfo.kt
Belphemur: I tested the code and I get the right date when using milisec.

Syer10 (approved):

PR #1022, opened 2024-08-30
Delete existing download of new chapters

schroda:
The downloader checks if pages are already downloaded to prevent unnecessary downloads. However, in case this is e.g. a re-uploaded chapter, this will prevent the new pages from getting downloaded


Syer10:
I'm not sure about this, what happens with sources that constantly change the chapter URLs?

schroda:
we could make it a setting which is disabled by default

Syer10:
I guess that works for now

schroda:
how common is it that sources change only the url compared to reuploading a chapter?

since it does cause issues where you can end up with different error states
- mismatch of downloaded pages and page count in database
- mix of outdated pages and new pages
- outdated pages

jobobby04:
Its very common for a source to change chapter URLs regularly. This has nothing to do with re-uploading a chapter though, this is something some sources have on schedule.

There is no way to tell if it's a re-upload or just a URL change unless the source has specified it. There is only 1 source I know of that actively re-uploads chapters if they find a better version.

vodkapmp:
> There is only 1 source I know of that actively re-uploads chapters if they find a better version.

My main goto source does this, and if the replaced chapter has more pages than before these pages are appended to the cbz and it ends up making no sense. 
(Often happen with first chapters of a volume when replaced with an official release that has 5-6 extra pages at the start, which also means these extras at the start aren't even there just the last 5-6 pages repeated)

I dunno if it's feasible, but maybe the setting could be source specific?

PR #1023, opened 2024-08-31
Feature/update log file rotation

schroda:


AndroidCompat/Config/src/main/java/xyz/nulldev/ts/config/Logging.kt
```diff
@@ -66,16 +71,44 @@ private fun getLogger(name: String): ch.qos.logback.classic.Logger {
...
+    maxFileSize: String,
+    maxTotalSize: String,
+) {
+    val logger = getBaseLogger()
+
+    val appender = logger.getAppender(FILE_APPENDER_NAME) as RollingFileAppender<*>? ?: return
+    val rollingPolicy = appender.rollingPolicy as SizeAndTimeBasedRollingPolicy<*>
+    rollingPolicy.apply {
+        maxHistory = maxFiles
+        setMaxFileSize(FileSize.valueOf(maxFileSize))
```
Syer10, line 112: How forgiving is this? I don't want to cause a server crash if a client sends a invalid value like `10gbb`

schroda: it did not crash when sending the mutation, however, once an invalid value was in the config file the server crashed on startup

Syer10: Maybe make a extension function that wraps FileSize.valueOf in a try/catch with a default if its invalid, I want to protect against manual file editing errors as well

schroda: because it's a java class a companion extension function can't be added

Syer10 (approved):

Syer10:
@schroda Needs to be formatted

PR #1070, opened 2024-09-09
Fix cookie expiry of FlareSolverr

Belphemur:
Just realized the cookie returned by flaresolverr is actually NOT used.

This actually render the usage of flaresolverr useless.


Robonau:
correct me if i'm wrong but i believe [this](https://github.com/Suwayomi/Suwayomi-Server/pull/1070/files#diff-bf72f548640863be3dd4f534c28e1e220932fc9cfddec9ac86fb459c4941e6b2R243-R250) section adds each cookie responded to that already

Belphemur:
You're right, the issue is somwhere else.
I'll change this PR to fix the cookie being "expired" when it's not ...

Belphemur:
Been running this for close to a week and it works perfectly

Syer10 (approved):

PR #1087, opened 2024-09-17
New Windows Script

BrutuZ:
Windows script to silently launch the server directly (skipping launcher), with optional auto-updater

### DIFF
--- a/scripts/resources/Suwayomi Server.bat
+++ b/scripts/resources/Suwayomi Server.bat
@@ -18,5 +18,5 @@ for /f "tokens=2" %%i in ('curl -s https://api.github.com/repos/Suwayomi/Suwayom
 )
 
 :launch
-start "" jre\bin\javaw %* -jar bin\Suwayomi-Server.jar
+start "" jre\bin\javaw %* -jar Suwayomi-Launcher.jar --launch
 popd
--- a/scripts/resources/Suwayomi Server.bat
+++ b/scripts/resources/Suwayomi Server.bat
@@ -1,0 +1,22 @@
+@echo off
+set auto_update=1
+
+:: Move CWD to script folder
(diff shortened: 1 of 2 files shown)


scripts/resources/Suwayomi Server.bat
```diff
@@ -0,0 +1,22 @@
...
+    echo Updating to %%~ni
+		del /q "*.jar"
+		curl -sL %%i -o "%%~nxi"
+		mklink /h "Suwayomi-Server.jar" "%%~nxi"
+	)
+	popd
+)
+
+:launch
+start "" jre\bin\javaw %* -jar bin\Suwayomi-Server.jar
```
Robonau: you can (in preview for now) launch the server using the launcher and the `--launch` argument
so you don't need to run it directly like this anymore

the advantage is that the launcher deals with root directories other than the default

BrutuZ: It is redirecting additional parameters with `%*` so it'll support that being configured with `-Dsuwayomi.tachidesk.config.server.rootDir=...` as [the wiki suggests](https://github.com/Suwayomi/Suwayomi-Server/wiki/The-Data-Directory#custom)

Syer10: What Robo means is that the launcher has its own config saved thats needed for people to always get the right settings, and the new --launch parameter will directly start the server without the launcher opening
```suggestion
start "" jre\bin\javaw %* -jar Suwayomi-Launcher.jar --launch
```

BrutuZ: Fair enough. Ideally the auto-updater would be built into the launcher, but I lack the Java-fu for that so batch it is 😅

PR #1109, opened 2024-11-08
Use uy.kohesive.injekt instead of com.github.inorichi.injekt

schroda:


Syer10:
@schroda I was thinking of using https://github.com/null2264/injekt-koin instead. This would allow us to replace Injekt with a more modern solution

schroda:
> @schroda I was thinking of using https://github.com/null2264/injekt-koin instead. This would allow us to replace Injekt with a more modern solution

done

server/src/main/kotlin/lib/koin/KoinInstance.kt
```diff
@@ -0,0 +1,9 @@
+package lib.koin
+
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+inline fun <reified T> getKoinInstance(): T =
+    object : KoinComponent {
+        val value: T by inject()
+    }.value
```
Syer10: Something like this should work.
```suggestion
inline fun <reified T> getKoinInstance(): T =
    KoinPlatformTools.defaultContext().get().get<T>()
```

server/src/main/kotlin/suwayomi/tachidesk/global/impl/AppUpdate.kt
```diff
@@ -22,12 +23,12 @@ data class UpdateDataClass(
...
 
-object AppUpdate {
+object AppUpdate : KoinComponent {
     private const val LATEST_STABLE_CHANNEL_URL = "https://api.github.com/repos/Suwayomi/Suwayomi-Server/releases/latest"
     private const val LATEST_PREVIEW_CHANNEL_URL = "https://api.github.com/repos/Suwayomi/Suwayomi-Server-preview/releases/latest"
 
-    private val json: Json by injectLazy()
-    private val network: NetworkHelper by injectLazy()
+    private val json: Json by inject()
+    private val network: NetworkHelper by inject()
```
Syer10: I think its fine to keep using injectLazy and Injekt.get, since it allows us to go without KoinComponent. KoinComponent is more for scoping to a context, which we don't need for out application.

Syer10 (approved):

PR #1112, opened 2024-11-10
Switch from Kodein to Koin

Syer10:


inverce:
Why do you prefer using 

`KoinPlatformTools.defaultContext().get().inject()`
instead of simply `by inject()`
as is prefered in official documentation

Instead of  https://insert-koin.io/docs/quickstart/kotlin
For example in AndroidCompat 
The default : KoinComponent effectively translates to KoinPlatformTools.defaultCotext(), but looks much clearer 

Also since i'm already asking, what is the reason of extra injectLazy instead of simply using default inject extension from insert-koin ?

Syer10:
The default `by inject()` requires implementing `KoinComponent`. This is a unnecessary step since the server will only have 1 global context. I see no use to implementing `KoinComponent` for everything that needs Koin. This is why I prefer `injektLazy()`, since its basically a utility function that does the same as `KoinPlatformTools.defaultContext().get().inject()`, but cleaner. I will implement my own utility function that does the same as `KoinPlatformTools.defaultContext().get().inject()` if/when we remove Injekt down the road.

PR #1168, opened 2024-11-26
Feature/update dependency exposed migrations to v3.3.0

schroda:
exposed-migrations requires a newer version of kotlin logging


Syer10:
I think this should just be included in #1150.

PR #1257, opened 2025-02-02
Add support for OPDS v1.2 to browse stored CBZ files

zeedif:
This PR adds support for OPDS v1.2, allowing users to browse and access CBZ files stored on the server through OPDS-compatible applications.

### Changes:
- Introduced `OpdsController` to handle OPDS feed requests.
- Added OPDS endpoints under `/api/v1/opds/v1.2`:
  - `GET /opds/v1.2` - Root feed listing available sources
  - `GET /opds/v1.2/source/{id}` - Feed for manga in a specific source
  - `GET /opds/v1.2/manga/{id}` - Feed of chapters for a specific manga
  - `GET /manga/{id}/chapter/{index}/download` - Fetch chapters as CBZ files
- Implemented `Opds` helper class to generate XML responses following OPDS v1.2 specifications.
- Extended `ChapterDownloadHelper` to provide CBZ file streams.

### Notes:
- OPDS feeds are generated dynamically and only include manga with downloaded chapters.
- The feature is designed for integration with OPDS-compatible readers.

This implementation was done with limited knowledge of Kotlin. I relied on AI assistance but tried to make use of existing libraries in the project. However, someone with more experience might find ways to improve this initial version.
It would be good to validate the existence of CBZ files to filter the list of sources and manga that actually contain them.
Maybe there's a way to allow downloading non-CBZ chapters by compressing their folders on the fly. Additionally, supporting OPDS 2.0 could be an interesting enhancement. The documentation for both versions is available at https://opds.io/.
Functions can be adjusted as needed—I placed the logic where I thought it fit best, but feel free to move things around.
I tested this implementation with Moon Reader and KoReader.
English is not my first language, so I may have left some text in Spanish. I didn't explore how internationalization is handled in the project, but it would be great if messages were adapted based on the user's language settings.


Robonau:
im not a server dev so i cant review code but i have some comments

you probably shouldn't have used `/api/v1` as the base, v1 is the deprecated REST api
(if we follow what was done with graphql it would be at `/api/opds`)

could you not use [OPDS-PSE](https://github.com/anansi-project/opds-pse) to allow for non cbz files or even non downloaded chapters?

zeedif:
I didn’t know that the v1 API was deprecated—I downloaded the repository just two days ago, so I worked on this pretty quickly. I’ll definitely check today if I can move it to the new API, though I don’t have any experience with GraphQL.

I also wasn’t aware of OPDS-PSE, but if that solves the issue of only supporting CBZ files, I’ll look into implementing it to make non-CBZ files accessible as well. Thanks for the feedback!

Robonau:
i read through some code.
is there anything in place for authentication?
suwayomi doesn't have a proper auth/account system but it does allow the user to set a basic auth
if i remember correctly OPDS has its own auth system and basic auth wont work

zeedif:
This is an initial implementation, and there are still things about OPDS that I’m not familiar with, like the authentication method you mentioned. I assumed it would work fine with basic authentication (though I haven’t tested it yet).

My priority list in order is: first, move the endpoints I added to the GraphQL route (and see if OPDS can be adjusted to work with GraphQL), then implement OPDS-PSE, and finally, work on authentication.

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/MangaController.kt
```diff
@@ -424,4 +425,30 @@ object MangaController {
                 httpCode(HttpStatus.NOT_FOUND)
             },
         )
+
+    val downloadChapter =
+        handler(
+            pathParam<Int>("mangaId"),
+            pathParam<Int>("chapterIndex"),
```
Syer10: We have dropped mangaId/chapterIndex in the graphql api, could this be swapped to chapter id? Maybe also add an alternative for the page endpoint as well

server/src/main/kotlin/suwayomi/tachidesk/manga/controller/MangaController.kt
```diff
@@ -424,4 +425,30 @@ object MangaController {
...
+            documentWith = {
+                withOperation {
+                    summary("Descargar capítulo como CBZ")
+                    description("Obtiene el archivo CBZ del capítulo especificado")
```
Syer10: Switch this to english

server/src/main/kotlin/suwayomi/tachidesk/opds/controller/OpdsController.kt
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+    val sourceFeed =
+        handler(
+            pathParam<Long>("sourceId"),
+            queryParam<Int?>("pageNumber"),
+            documentWith = {
+                withOperation {
+                    summary("OPDS Source Feed")
+                    description("Feed OPDS de una fuente específica de mangas")
```
Syer10: English

server/src/main/kotlin/suwayomi/tachidesk/opds/controller/OpdsController.kt
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+    val mangaFeed =
+        handler(
+            pathParam<Int>("mangaId"),
+            queryParam<Int?>("pageNumber"),
+            documentWith = {
+                withOperation {
+                    summary("OPDS Manga Feed")
+                    description("Feed OPDS de capítulos de un manga específico")
```
Syer10: English

Syer10 (changes requested):

zeedif:
I have made the requested changes: I updated the endpoint descriptions to English, replaced the use of chapterIndex for downloading CBZ files with chapterId, and adjusted OPDS download links accordingly. They now use chapter.id instead of chapter.sourceOrder.

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
```diff
@@ -0,0 +1,343 @@
...
+                                },
+                            summary = manga.description?.let { OpdsDataClass.Summary(value = it) },
+                            link =
+                                listOfNotNull(
+                                    OpdsDataClass.Link(
+                                        rel = "subsection",
+                                        href = "$baseUrl/manga/${manga.id}",
+                                        type = "application/atom+xml;profile=opds-catalog;kind=acquisition",
+                                    ),
+                                    manga.thumbnailUrl?.let {
```
Syer10: Is this the proxied thumbnail url?

Robonau: does the db keep the un-proxied thumbnail url?

Syer10: It doesn't, it needs to be run through proxyThumbnailUrl function

zeedif: I just made the change to use the proxyThumbnail function, though I can say that the obtained URL is practically the same. I also took the opportunity to simplify the retrieval of now and the assignment of thumbnail in Opds.kt.

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -78,6 +78,7 @@ object MangaAPI {
             patch("{mangaId}/chapter/{chapterIndex}/meta", MangaController.chapterMeta)
 
             get("{mangaId}/chapter/{chapterIndex}/page/{index}", MangaController.pageRetrieve)
+            get("{mangaId}/chapter/{chapterId}/download", MangaController.downloadChapter)
```
Syer10: Just for clarification, why is mangaId still included?

Robonau: seems like its just a hold over, it being based on the v1 using mangaId and chapterIndex

Syer10: Then it can be removed, since this is now based on chapterId

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
```diff
@@ -0,0 +1,342 @@
...
+import java.time.Instant
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+
+object Opds {
+    private const val ITEMS_PER_PAGE = 20
+
+    fun getRootFeed(baseUrl: String): String {
+        val formattedNow = opdsDateFormatter.format(Instant.now())
+        val sources =
```
schroda, line 33: this results, if I'm not mistaken, in 3n queries

1. read all sources (one query)
2. read each source again (n source queries + n extension queries)
3. check that each source has manga with chapters (n queries)

I don't use many extensions and the source list is already at 181, thus, each request would result in 544 queries

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
zeedif: I used mangaId as a reference based on how the API was structured in v1 to retrieve individual pages. However, since we are now using chapterId as the primary identifier, I understand that mangaId is no longer necessary in this context. I will make the change to remove mangaId shortly.

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
```diff
@@ -0,0 +1,340 @@
...
+            transaction {
+                SourceTable
+                    .join(MangaTable, JoinType.INNER) { MangaTable.sourceReference eq SourceTable.id }
+                    .join(ChapterTable, JoinType.INNER) { ChapterTable.manga eq MangaTable.id }
+                    .select(SourceTable.id, SourceTable.name)
+                    .where { ChapterTable.isDownloaded eq true }
+                    .orderBy(SourceTable.name to SortOrder.ASC)
+                    .distinct()
+                    .map {
+                        object {
```
Syer10: I would prefer to avoid anonymous objects, can you move this to a data class?

Syer10 (changes requested):
Looks good to me, only this one nitpick that I see.

zeedif:
> Looks good to me, only this one nitpick that I see.

Upon closer inspection, I realized that when mapping with val mangaData = MangaTable.toDataClass(mangaEntry), it's no longer necessary to use the proxyThumbnailUrl function, as it's already handled by the first function. Therefore, I'm going back to retrieving the thumbnailUrl from the manga's property.

Syer10 (approved):
Got it, that works for me

schroda (approved):

Syer10:
@zeedif lint failed, make sure to run the formatter(`./gradlew ktlintFormat`) and fix any errors that the formatter can't automatically fix.

zeedif:
@Syer10  I ran the command you asked me to, and I received this response: "The CopyProcessingSpec.setFileMode(Integer) method has been deprecated."  I searched for the method in the code, but I can't find any matches.

```bash
PS D:\Workspace\Kotlin\Suwayomi-Server> ./gradlew ktlintFormat
Type-safe project accessors is an incubating feature.

[Incubating] Problems report is available at: file:///D:/Workspace/Kotlin/Suwayomi-Server/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/8.12/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 2s
24 actionable tasks: 1 executed, 23 up-to-date
PS D:\Workspace\Kotlin\Suwayomi-Server>                                                                                                                                                                                                                                                          
                                                             
```

zeedif:
From what I'm seeing, the problem seems to be related to a task that Gradle executes. However, the only thing I've found that references fileMode is this segment in the getAndroid.ps1 file:
```ps1
function Remove-Files-Zip($zipfile, $paths)
{
    [Reflection.Assembly]::LoadWithPartialName('System.IO.Compression') | Out-Null

    $stream = New-Object IO.FileStream($zipfile, [IO.FileMode]::Open)
    $mode   = [IO.Compression.ZipArchiveMode]::Update
    $zip    = New-Object IO.Compression.ZipArchive($stream, $mode)

    if ($paths.getType().Name -eq "Object[]")
    {
        $paths | ForEach-Object {
            $path = $_
            ($zip.Entries | Where-Object { $_.FullName -like $path }) | ForEach-Object { Write-Output "Deleting: $($_.FullName)"; $_.Delete() }
        }
    }
    else
    {
        ($zip.Entries | Where-Object { $_.FullName -like $paths }) | ForEach-Object { Write-Output "Deleting: $($_.FullName)"; $_.Delete() }
    }


    $zip.Dispose()
    $stream.Close()
    $stream.Dispose()
}
```

Syer10:
After running the command some of the project files should have been automatically formatted, make sure to commit those changes.
The deprecated warning is fine for now, you can ignore it.

zeedif:
Thanks for clarifying that for me. I hadn't noticed that it was changing my files. I've just made the commit with the changes it made.

PR #1263, opened 2025-02-10
Remove redundant code and add next/prev links

zeedif:
This pull request fixes the pagination functionality of the OPDS feeds by ensuring that the "next" and "prev" navigation links are correctly included. Previously, although the feeds were paginated, the absence of these links prevented dynamic loading of additional entries.

In addition, the code has been refactored to eliminate redundancy by centralizing the feed construction logic within the FeedBuilder. This change not only makes the implementation cleaner and more maintainable but also ensures consistency across various OPDS endpoints.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
@@ -160,29 +160,10 @@ object Opds {
                 Pair(sources, totalCount)
             }
 
-        return serialize(
-            OpdsXmlModels(
-                id = "sources",
-                title = "Sources",
-                updated = formattedNow,
-                totalResults = totalCount,
-                itemsPerPage = ITEMS_PER_PAGE,
-                startIndex = (pageNum - 1) * ITEMS_PER_PAGE + 1,
-                author = OpdsXmlModels.Author("Suwayomi", "https://suwayomi.org/"),
-                links =
-                    listOfNotNull(
-                        OpdsXmlModels.Link(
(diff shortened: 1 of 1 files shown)


Syer10 (approved):

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
```diff
@@ -780,16 +677,14 @@ object Opds {
...
-                                    if (id == "opds") {
-                                        baseUrl
-                                    } else if (searchQuery != null) {
-                                        "$baseUrl/$id?query=$searchQuery"
-                                    } else {
-                                        "$baseUrl/$id?pageNumber=$pageNum"
+                                    when {
+                                        id == "opds" -> baseUrl
+                                        searchQuery != null -> "$baseUrl/$id?query=$searchQuery"
+                                        else -> "$baseUrl/$id?pageNumber=$pageNum"
```
Syer10, line 687: What happens when both a searchQuery and a page are needed?

zeedif: In mangasFeed endpoint in the OpdsV1Controller, I intentionally force the page number to always be 1 when a search query is present. This decision was inspired by the OPDS search implementation in Kavita, where only the first 20 results are loaded without pagination. The rationale is that when performing a search, the goal is to narrow down the possibilities quickly. If the desired result isn’t among the first results, it usually indicates that the search terms might need to be refined rather than relying on pagination.

PR #1278, opened 2025-02-22
Add support for opds-pse for undownloaded chapters

shirishsaxena:
### Changes
- Added OPDS-PSE support for undownloaded chapters.
- Modified GET chapter/{{chapterId}/download to create a CBZ file for folder downloads.
- Removed manga descriptions from chapter entries to reduce XML size for manga with very long descriptions. Each chapter entry now returns the manga title, chapter name, and the scanlator.
- Adds unicodes as a prefix to chapter titles in the OPDS feed when the chapter is has been read, in progress, or not started.
- Page streaming now updates the lastPageRead for chapter

### Notes
- OPDS-PSE doesn't work without pageCount, and the redirected endpoint also fails even if we return a hardcoded count in the chapter entry. So, added a check in [ChapterForDownload](https://github.com/Suwayomi/Suwayomi-Server/blob/4f7962c98bb7fd0c04771e894a2aa72a2c155711/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/chapter/ChapterForDownload.kt#L41-L40) to update the PageTableList & PageCount only when there is a mismatch.
-I tried making this call for all chapters in the paginated manga feed, but even with just 20 entries per page, it felt pretty slow. Plus, OPDS clients don’t really seem to support seamless chapter reading (KOReader definitely doesn’t). So instead, I added a separate “detail” page for each chapter when the page count is missing. This way, we can read undownloaded chapters, and if we revisit the chapter later, it won’t redirect us to the detail page again.

**_FYI: I use KOReader to read manga, so this entire modification is based on that._**

<details> 
<summary>KOReader screen recording for this change</summary>

https://github.com/user-attachments/assets/458fb38f-510a-4d26-990c-9d0672eb6f9e

</details> 
<details> 
<summary>KoReader with "D" as prefix for downloaded chapter</summary>
Chapter 126 for Slime's manga was downloaded in a folder. 

image

https://github.com/user-attachments/assets/01dacacc-4abb-4f94-beda-b22dc91b2da2

</details>

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
@@ -29,6 +29,7 @@ import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.statements.BatchUpdateStatement
 import org.jetbrains.exposed.sql.transactions.transaction
 import org.jetbrains.exposed.sql.update
+import suwayomi.tachidesk.graphql.mutations.ChapterMutation
 import suwayomi.tachidesk.manga.impl.Manga.getManga
 import suwayomi.tachidesk.manga.impl.download.DownloadManager
 import suwayomi.tachidesk.manga.impl.download.DownloadManager.EnqueueInput
@@ -693,4 +694,31 @@ object Chapter {
                     }
             }
         }
+
+    fun updateChapterProgress(
+        mangaId: Int,
+        chapterIndex: Int,
+        pageNo: Int,
+    ) {
+        val chapterData =
+            transaction {
+                ChapterTable
+                    .selectAll()
+                    .where {
+                        (ChapterTable.sourceOrder eq chapterIndex) and
+                            (ChapterTable.manga eq mangaId)
+                    }.first()
+                    .let { ChapterTable.toDataClass(it) }
+            }
+
+        val isRead = chapterData.pageCount == pageNo + 1
+        val chapterMutation = ChapterMutation()
+        val input =
+            ChapterMutation.UpdateChapterInput(
+                id = chapterData.id,
+                patch = ChapterMutation.UpdateChapterPatch(lastPageRead = pageNo + 1, isRead = isRead),
+            )
+
+        chapterMutation.updateChapter(input)
+    }
 }
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
@@ -717,7 +717,7 @@ object Chapter {
             mangaId,
             chapterIndex,
             isRead = isRead,
-            lastPageRead = oneIndexedPageNo,
+            lastPageRead = pageNo,
             isBookmarked = null,
             markPrevRead = null,
         )
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
@@ -29,7 +29,6 @@ import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.statements.BatchUpdateStatement
 import org.jetbrains.exposed.sql.transactions.transaction
 import org.jetbrains.exposed.sql.update
-import suwayomi.tachidesk.graphql.mutations.ChapterMutation
 import suwayomi.tachidesk.manga.impl.Manga.getManga
 import suwayomi.tachidesk.manga.impl.download.DownloadManager
 import suwayomi.tachidesk.manga.impl.download.DownloadManager.EnqueueInput
@@ -263,8 +262,8 @@ object Chapter {
                 // we got some clean up due
                 if (chaptersIdsToDelete.isNotEmpty()) {
                     transaction {
-                        PageTable.deleteWhere { PageTable.chapter inList chaptersIdsToDelete }
-                        ChapterTable.deleteWhere { ChapterTable.id inList chaptersIdsToDelete }
+                        PageTable.deleteWhere { chapter inList chaptersIdsToDelete }
+                        ChapterTable.deleteWhere { id inList chaptersIdsToDelete }
                     }
                 }
 
@@ -322,7 +321,7 @@ object Chapter {
                     }
 
                     MangaTable.update({ MangaTable.id eq mangaId }) {
-                        it[MangaTable.chaptersLastFetchedAt] = Instant.now().epochSecond
+                        it[chaptersLastFetchedAt] = Instant.now().epochSecond
                     }
                 }
 
@@ -444,7 +443,7 @@ object Chapter {
                     }
                     lastPageRead?.also {
                         update[ChapterTable.lastPageRead] = it
-                        update[ChapterTable.lastReadAt] = Instant.now().epochSecond
+                        update[lastReadAt] = Instant.now().epochSecond
                     }
                 }
             }
@@ -535,7 +534,7 @@ object Chapter {
                 }
                 lastPageRead?.also {
                     update[ChapterTable.lastPageRead] = it
-                    update[ChapterTable.lastReadAt] = now
+                    update[lastReadAt] = now
                 }
             }
         }
@@ -604,7 +603,7 @@ object Chapter {
                 ChapterMetaTable.insert {
                     it[ChapterMetaTable.key] = key
                     it[ChapterMetaTable.value] = value
-                    it[ChapterMetaTable.ref] = chapterId
+                    it[ref] = chapterId
                 }
             } else {
                 ChapterMetaTable.update({ (ChapterMetaTable.ref eq chapterId) and (ChapterMetaTable.key eq key) }) {
@@ -711,14 +710,16 @@ object Chapter {
                     .let { ChapterTable.toDataClass(it) }
             }
 
-        val isRead = chapterData.pageCount == pageNo + 1
-        val chapterMutation = ChapterMutation()
-        val input =
-            ChapterMutation.UpdateChapterInput(
-                id = chapterData.id,
-                patch = ChapterMutation.UpdateChapterPatch(lastPageRead = pageNo + 1, isRead = isRead),
-            )
+        val oneIndexedPageNo = pageNo.inc()
+        val isRead = chapterData.pageCount.takeIf { it == oneIndexedPageNo }?.let { true }
 
-        chapterMutation.updateChapter(input)
+        modifyChapter(
+            mangaId,
+            chapterIndex,
+            isRead = isRead,
+            lastPageRead = oneIndexedPageNo,
+            isBookmarked = null,
+            markPrevRead = null,
+        )
     }
 }
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
@@ -52,7 +52,10 @@ object ChapterDownloadHelper {
         chapterId: Int,
     ): Pair<InputStream, Long> = provider(mangaId, chapterId).getAsArchiveStream()
 
-    fun getCbzForDownload(chapterId: Int): Triple<InputStream, String, Long> {
+    fun getCbzForDownload(
+        chapterId: Int,
+        markAsRead: Boolean?,
+    ): Triple<InputStream, String, Long> {
         val (chapterData, mangaTitle) =
             transaction {
                 val row =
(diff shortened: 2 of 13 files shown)


Robonau:
@zeedif your input is probably welcomed here

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
```diff
@@ -48,7 +52,7 @@ object ChapterDownloadHelper {
         return FolderProvider(mangaId, chapterId)
     }
 
-    suspend fun getCbzDownload(chapterId: Int): Triple<InputStream, String, String> {
+    fun getCbzForDownload(chapterId: Int): Triple<InputStream, String, String> {
```
schroda: how about moving all this logic to the providers by adding a `getAsArchiveStream` function

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/chapter/ChapterForDownload.kt
```diff
@@ -27,13 +28,22 @@ import suwayomi.tachidesk.manga.model.table.MangaTable
 import suwayomi.tachidesk.manga.model.table.PageTable
 import suwayomi.tachidesk.manga.model.table.toDataClass
 
-suspend fun getChapterDownloadReady(
+suspend fun getPageCountForChapter(
```
schroda: instead of this and the new `refreshPageListForChapter` function, `getChapterDownloadReady` should be used to ensure that the chapter is in a "proper" state to be read (see added comment for `chapterMetadataFeed`)

server/src/main/kotlin/suwayomi/tachidesk/opds/controller/OpdsV1Controller.kt
```diff
@@ -278,6 +278,31 @@ object OpdsV1Controller {
             },
         )
 
+    var chapterMetadataFeed =
```
schroda, line 281: what speaks against just returning this feed for every chapter of a manga by default?

currently the opds implementation does not follow to intended flow for displaying a chapter (checking that the chapter is ready before trying to display any pages) which potentially leads to issues, e.g. in case a chapter is incorrectly flagged as downloaded

zeedif:
@ShirishSaxena Thank you for the contribution. While the `[D]` prefix for downloaded chapters is functional, I find it creates visual inconsistency between prefixed and non-prefixed entries. We could use **emoji pairs** to indicate downloaded/undownloaded states while maintaining alignment:

**Downloaded Chapter Emojis**  
- ✅ **Checkmark** - Universal "completed" symbol  
- 📁 **File Folder** - Direct file reference  
- 💾 **Floppy Disk** - Classic "saved" metaphor  
- 📥 **Inbox Tray** - Content "received"  
- 🗃️ **Card File Box** - Organized storage  
- 🟢 **Green Circle** - Traffic light "go" system  
- ✔️ **Heavy Checkmark** - Definitive confirmation  
- 📦 **Package** - Contained/archived content  
- ⬇️ **Down Arrow** - Downloaded state  
- 🔒 **Locked** - Local secured access  

**Undownloaded Chapter Emojis**  
- 🌐 **Globe** - Streaming from web  
- ⏳ **Hourglass** - Pending action  
- ↗️ **NE Arrow** - External resource  
- ☁️ **Cloud** - Remote content  
- ⚪ **White Circle** - Empty/unfilled state  
- 🔄 **Arrows CW** - Refresh/stream again  
- ➖ **Dash** - Minimal "not present"  
- ⏯️ **Play/Pause** - On-demand streaming  
- 🚀 **Rocket** - Instant access metaphor  
- 🔗 **Link** - Network dependency

zeedif:
I find the support for reading manga without downloading useful, as it saves the hassle of going to the web UI and downloading them manually. However, I wonder if it’s possible to mark a chapter as 'read' when accessed via OPDS. For example, triggering the 'read' status either after downloading its CBZ through OPDS or when reaching the last image via OPDS-PSE streaming.

server/src/main/kotlin/suwayomi/tachidesk/opds/controller/OpdsV1Controller.kt
shirishsaxena: It was me just being kinda lazy and only focusing on adding a subsection to fetch, and I figured these sorts of changes could come in later stages... But yeah, this kinda makes sense, especially when we include page tracking, outdated page URLs, and such.

shirishsaxena:
@zeedif , I didn’t give much thought to the "[D]" identifier, and yeah, you’re right. 
I added the Unicode characters that I felt were appropriate (_Most I searched were sadly not supported by KOReader and this is my main goto..._) for chapter progression in the Chapters feed. Now, each chapter feed goes to the chapterMetadata feed regardless of whether it’s downloaded or not, so adding a downloaded identifier didn’t make much sense anymore. We could possibly include it in the metadata page/feed...
Also, added page progression updates for PSE streaming. While it's possible to mark chapters as read whenever a download occurs through OPDS, I feel it doesn’t make sense. It makes more sense to have it in page streaming since we're actively reading there.

https://github.com/user-attachments/assets/14b890da-6638-464d-82d8-0dc43f0168fd

zeedif:
The implementation seems fine to me, though requiring a chapter "detail feed" adds unnecessary friction to the OPDS navigation flow. While I understand the current limitation of slow pageCount checks on paginated feeds, a longer-term solution would be to pre-fetch pageCount once during library addition or chapter discovery (e.g., background tasks when manga is added/updated). This would eliminate the need for the detail feed entirely.

Syer10:
> The implementation seems fine to me, though requiring a chapter "detail feed" adds unnecessary friction to the OPDS navigation flow. While I understand the current limitation of slow pageCount checks on paginated feeds, a longer-term solution would be to pre-fetch pageCount once during library addition or chapter discovery (e.g., background tasks when manga is added/updated). This would eliminate the need for the detail feed entirely.

Pre-fetching page count is not possible, since that requires a request to the source and there are manga that have hundreds of chapters. Its not viable.

shirishsaxena:
I think I have all the changes from my end, but let me know if I need to make any other changes... Final change also includes the library updates feed, similar to 'Updates' we have in the web UI & Android app.

<details>
<summary> LIbrary Feed SS </summary>

Screenshot_20250224_003448

</details>

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -693,4 +694,31 @@ object Chapter {
...
+                ChapterTable
+                    .selectAll()
+                    .where {
+                        (ChapterTable.sourceOrder eq chapterIndex) and
+                            (ChapterTable.manga eq mangaId)
+                    }.first()
+                    .let { ChapterTable.toDataClass(it) }
+            }
+
+        val isRead = chapterData.pageCount == pageNo + 1
```
schroda: this currently will mark an already read chapter as unread, to prevent this it would need to be `null` instead of `false`

server/src/main/kotlin/suwayomi/tachidesk/manga/model/table/ChapterTable.kt
```diff
@@ -41,31 +41,43 @@ object ChapterTable : IntIdTable() {
...
-        lastReadAt = chapterEntry[lastReadAt],
-        index = chapterEntry[sourceOrder],
-        fetchedAt = chapterEntry[fetchedAt],
-        realUrl = chapterEntry[realUrl],
-        downloaded = chapterEntry[isDownloaded],
-        pageCount = chapterEntry[pageCount],
-        chapterCount =
+fun ChapterTable.toDataClass(
+    chapterEntry: ResultRow,
+    fetchChapterCount: Boolean = true,
```
schroda: since nothing gets fetched here something like `include/add` would fit better

server/src/main/kotlin/suwayomi/tachidesk/manga/model/table/MangaTable.kt
```diff
@@ -46,28 +46,35 @@ object MangaTable : IntIdTable() {
...
-        inLibraryAt = mangaEntry[inLibraryAt],
-        meta = getMangaMetaMap(mangaEntry[id].value),
-        realUrl = mangaEntry[realUrl],
-        lastFetchedAt = mangaEntry[lastFetchedAt],
-        chaptersLastFetchedAt = mangaEntry[chaptersLastFetchedAt],
-        updateStrategy = UpdateStrategy.valueOf(mangaEntry[updateStrategy]),
-    )
+fun MangaTable.toDataClass(
+    mangaEntry: ResultRow,
+    fetchMangaMeta: Boolean = true,
```
schroda: same regarding `fetch`

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
```diff
@@ -422,56 +420,140 @@ object Opds {
...
-                    } else {
+                        ),
+                    )
+                }
+                if (isMetaDataEntry) {
+                    add(
                         OpdsXmlModels.Link(
                             rel = "http://vaemendis.net/opds-pse/stream",
-                            href = "/api/v1/manga/${manga.id}/chapter/${chapter.index}/page/{pageNumber}",
+                            href = "/api/v1/manga/${manga.id}/chapter/${chapter.index}/page/{pageNumber}?updateProgress=true",
```
schroda, line 518: what about `/api/v1/chapter/${chapter.id}/download` regarding marking a chapter as read?
only adding it here would mean both links have different behaviour which no user will know about

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -693,4 +694,31 @@ object Chapter {
...
+        val chapterMutation = ChapterMutation()
+        val input =
+            ChapterMutation.UpdateChapterInput(
+                id = chapterData.id,
+                patch = ChapterMutation.UpdateChapterPatch(lastPageRead = pageNo + 1, isRead = isRead),
+            )
+
+        chapterMutation.updateChapter(input)
```
schroda:
```suggestion
        Chapter.modifyChapter(mangaId, chapterIndex, isRead, lastPageRead = pageNo)
```

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
shirishsaxena: Added for download as well.

schroda: I'm not sure why, but I thought both links were different ways to stream the chapter, so I'm not sure if you want to keep marking it as read if you download it, since you hadn't added it here anyway

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Chapter.kt
```diff
@@ -693,4 +693,33 @@ object Chapter {
...
+            }
+
+        val oneIndexedPageNo = pageNo.inc()
+        val isRead = chapterData.pageCount.takeIf { it == oneIndexedPageNo }?.let { true }
+
+        modifyChapter(
+            mangaId,
+            chapterIndex,
+            isRead = isRead,
+            lastPageRead = oneIndexedPageNo,
```
schroda:
```suggestion
            lastPageRead = pageNo,
```
`lastPageRead` is index based

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
shirishsaxena: Oh... yeah, I missed the flag in the OPDS feed. I think I'll keep it as it is? It doesn't really make sense to mark a chapter as read when you're just downloading it and not actively reading/streaming...

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
```diff
@@ -48,7 +47,15 @@ object ChapterDownloadHelper {
...
 
-    suspend fun getCbzDownload(chapterId: Int): Triple<InputStream, String, String> {
+    fun getArchiveStreamWithSize(
+        mangaId: Int,
+        chapterId: Int,
+    ): Pair<InputStream, Long> = provider(mangaId, chapterId).getAsArchiveStream()
+
+    fun getCbzForDownload(
+        chapterId: Int,
+        markAsRead: Boolean?,
```
schroda: can also be removed (+ https://github.com/Suwayomi/Suwayomi-Server/pull/1278/files#diff-233bf65f8dcc857862c8c602936b370223867d3f07d30d9eec569fca6becaf1aR437)

schroda (approved):

shirishsaxena:
All comments should be addressed. Anything else needed from my side for merging?

Syer10 (approved):

Issue #1317, opened 2025-03-20
[Bug] Cannot open ComicK extension settings


zerebos:
## Device information
- Suwayomi-Server version: v1.1.1-r1701
- Server Operating System: Ubuntu 23.04
- Server Desktop Environment: N/A
- Server JVM version: eclipse-temurin:21.0.6_7-jre-noble
- Client Operating System: windows, mac, android
- Client Web Browser: Firefox 136

## Steps to reproduce
1. Add extension repository from Keiyoushi or Komikku (I tried both)
2. Install ComicK extension
3. Browse ComicK
4. Press the cog at the top right

### Expected behavior
The settings page should be displayed and be able to change settings.

### Actual behavior

<img width="668" alt="Image" src="https://github.com/user-attachments/assets/522c545e-ecc5-4f45-96c0-e1b922691ff2" />

mtarabba:
![Image](https://github.com/user-attachments/assets/7d18c25c-bc04-42f0-8d24-c632acf5bfef)

maybe related to this? it gave me this error when im trying to update manually

schroda:
<details>
<summary>stack trace of the error</summary>

```
java.lang.RuntimeException: Stub!
at androidx.preference.TwoStatePreference.setSummaryOn(TwoStatePreference.java:31)
at eu.kanade.tachiyomi.extension.all.comickfun.Comick.setupPreferenceScreen(Unknown Source)
at suwayomi.tachidesk.manga.impl.Source.getSourcePreferencesRaw(Source.kt:112)
at suwayomi.tachidesk.graphql.types.SourceType.preferences(SourceType.kt:76)
at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at kotlin.reflect.jvm.internal.calls.CallerImpl$Method.callMethod(CallerImpl.kt:97)
at kotlin.reflect.jvm.internal.calls.CallerImpl$Method$Instance.call(CallerImpl.kt:113)
at kotlin.reflect.jvm.internal.KCallableImpl.callDefaultMethod$kotlin_reflection(KCallableImpl.kt:207)
at kotlin.reflect.jvm.internal.KCallableImpl.callBy(KCallableImpl.kt:112)
[...]

PR #1319, opened 2025-03-22
Feature/graphql log execution exceptions

schroda:


server/src/main/kotlin/suwayomi/tachidesk/graphql/server/TachideskGraphQLServer.kt
```diff
@@ -44,11 +49,34 @@ class TachideskGraphQLServer(
...
+                    val exception = handlerParameters.exception
+                    val sourceLocation = handlerParameters.sourceLocation
+                    val path = handlerParameters.path
+
+                    logger.error(exception) { "GraphQL execution failed due to" }
+
+                    val error =
+                        ExceptionWhileDataFetching(
+                            path,
+                            Throwable(exception.message + "\r\n\r\n" + exception.stackTraceToString()),
```
schroda: not sure if we want to include the stack trace, but it would probably improve bug reports since most users don't have a clue about server logs anyway

Syer10: Why is the stacktrace being added as a message instead of passing the exception to the second parameter of throwable?

schroda: `ExceptionWhileDataFetching` only uses the message of the exception
e.g. in case of #1317 the stack trace is missing and the error is just `Stub!`

Syer10: I see, I think it would be fine to include it in the message. I would prefer that clients have it under an expand for more info section though since it would confuse the user to see a whole block of text. Would WebUI be able to regex/split by `\r\n\r\n` to capture and only show the basic message until they expand the message?

schroda: sure, I can do that in the webUI

Syer10 (approved):

PR #1387, opened 2025-05-16
Improve Downloads Handling

Syer10:
- If a download already exists and it has a page count over the known page count, it is set as complete
- If a download as 1000+ pages, it uses a `%04d` string format, 10000+ pages, it uses a `%05d`, etc
- Page count after a download is updated in the database based on how many image files there are
- Updates page count for downloaded chapters when getting ready to read
- Assume downloaded if first page is found, this means that chapters not marked as read will be assumed downloaded when getting page count. This is because we now store non-downloaded chapters in the system temp folders

Will fix issues with Mihon long page split chapters and 1000+ page chapters

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
@@ -104,7 +104,13 @@ abstract class ChaptersFilesProvider<Type : FileType>(
         downloadCacheFolder.mkdirs()
 
         val pageCount = download.chapter.pageCount
-        if (downloadCacheFolder.listFiles().orEmpty().size >= pageCount) {
+        if (
+            downloadCacheFolder
+                .listFiles()
+                .orEmpty()
+                .filter { it.name != COMIC_INFO_FILE }
+                .size >= pageCount
+        ) {
             download.progress = 1f
             step(download, false)
         } else {
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
@@ -9,8 +9,10 @@ import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.sample
 import org.apache.commons.compress.archivers.zip.ZipArchiveEntry
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.transactions.transaction
+import org.jetbrains.exposed.sql.update
 import suwayomi.tachidesk.manga.impl.Page
 import suwayomi.tachidesk.manga.impl.download.model.DownloadChapter
 import suwayomi.tachidesk.manga.impl.util.createComicInfoFile
@@ -76,6 +78,8 @@ abstract class ChaptersFilesProvider<Type : FileType>(
         return Pair(getImageInputStream(image).buffered(), "image/$imageFileType")
     }
 
+    fun getImageCount(): Int = getImageFiles().filter { it.getName() != COMIC_INFO_FILE }.size
+
     override fun getImage(): RetrieveFile1Args<Int> = RetrieveFile1Args(::getImageImpl)
 
     /**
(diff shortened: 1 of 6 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
```diff
@@ -153,6 +162,12 @@ abstract class ChaptersFilesProvider<Type : FileType>(
 
         handleSuccessfulDownload()
 
+        transaction {
```
schroda, line 171: is it possibel to get here without having an up-to-date page count?

schroda:
> If a download as 1000+ pages, it uses a %04d string format, 10000+ pages, it uses a %05d, etc

does this work with existing downloads that have more than 999 pages?

> Will fix issues with Mihon long page split chapters and 1000+ page chapters

how does it fix the long page split pages?

Syer10:
> does this work with existing downloads that have more than 999 pages?

No, currently chapters with 1000+ pages being downloaded end up being named things like `005.jpg` `999.jpg` and `1500.jpg`. This breaks our index sorting of pages that requires the prefixed 0. This is why the fix is needed

> how does it fix the long page split pages?

Long split pages from Mihon take the prefix name and then add a split count after it, like `001_001.jpg` and `001_002.jpg`.

Since we only rely on the index + page count, we can easily map to the correct page as long as we have the correct pagecount. This just means we have to assume the chapter is correct when checking the download and it has a pagecount(which includes the split pages as a seperate page) of equal or higher then we expect

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
Syer10: This is mainly for chapters like we get from Mihon that have more pages then expected, so we update our known page-count

Syer10: I've also got it so that if the chapter has not been fetched, it will accept a download with any page count. This allows for offline migration of chapter downloads

schroda:
> No, currently chapters with 100+ pages being downloaded end up being named things like 005.jpg 999.jpg and 1500.jpg. This breaks our index sorting of pages that requires the prefixed 0. This is why the fix is needed

What I meant with this was if the changes can handle reading existing downloads with 1000+ pages, or will it break them, due to the additional leading 0s which the existing page file names won't have?

> Long split pages from Mihon take the prefix name and then add a split count after it, like 001_001.jpg and 001_002.jpg.
>
> Since we only rely on the index + page count, we can easily map to the correct page as long as we have the correct pagecount. > This just means we have to assume the chapter is correct when checking the download and it has a pagecount of equal or > higher then we expect

So the original page still exists with the expected name and it will get selected when the page is requested?

Won't we have to filter out the split pages then to get the correct page count, since we don't actually handle them and instead just ignore them?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
schroda: isn't this already handled in `ChapterForDownoad#asDownloadReady` with the added `updatePageCount(ChapterDownloadHelper.getImageCount(mangaId, chapterId))` in case the chapter is considered to be already downloaded?

Syer10:
> What I meant with this was if the changes can handle reading existing downloads with 1000+ pages, or will it break them, due to the additional leading 0s which the existing page file names won't have?

It doesn't change how existing pages are handled, its just a fix for how downloaded pages end up being named when downloaded, based on the total page count. The 1000+ page downloads are broken on thier own currently.



> So the original page still exists with the expected name and it will get selected when the page is requested?
> 
> Won't we have to filter out the split pages then to get the correct page count, since we don't actually handle them and instead just ignore them?

Nope, the original page doesn't exist, just the split pages.

We don't actually use the page name after we create it, the name is just for index sorting. We rely on the page index in our sorted list to decide what page to provide, which is why changing the page count will help fix it

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
Syer10: Technically yes, but I think its fine to have it in both places? This allows for more updated data since the user might not read the chapter for a long time

schroda:
oh, nevermind, I got confused with the created file name in the retrieve function, but that is only used for the non downloaded case

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
schroda: atm at least `getChapterDownloadReadyById` is required to be called before a chapter download is possible

schroda (approved):

PR #1392, opened 2025-05-18
Feat: Adds OPDS Chapter Filtering/Ordering

### LINKED ISSUE #1390: Feature Request: Advanced OPDS Sorting and Filtering (e.g. show unread only, sort by oldest/newest)

Hello and thank you for your work on Suwayomi!

I am using the OPDS catalog feature with KOReader and I noticed that the OPDS feed always returns chapters in the same order (usually newest first) and paginates results (e.g. 20 chapters per page).
For users with a large library, it would be very helpful to have more control over how chapters are displayed in the OPDS feed.

Feature suggestions:
Add an option to sort chapters in ascending or descending order (e.g. oldest first or newest first).

Add an option to filter out read chapters, so only unread chapters appear in the OPDS feed.

Optionally, allow users to increase the number of chapters per page or disable pagination for OPDS (for clients that support it).

This would make it much easier to use Suwayomi with e-readers and OPDS clients like KOReader, especially for people who want to read series in chronological order or only see unread chapters.

Thank you for considering this feature request!

v2.0.1727

### Other details

_No response_

### Acknowledgements

shirishsaxena:
Closes #1390
### OPDS
- Moved existing hardcoded OPDS settings to `server.conf`.
- Added `opdsMarkAsReadOnDownload`. Previously, I didn’t think it made much sense since we were hardcoding it, but now that we're using server.conf, I think it’s best to leave it up to the user
- Added a HEAD request for CBZ downloads. Some OPDS clients (mainly KOReader :3) use HEAD requests to pre-fetch the filename before downloading. _Clients usually fall back to a default `<Title> + <Author>.cbz` format if the HEAD request fails..._
- Added "⬇" in the ChapterEntry title for ChapterMetadataFeed to better differentiate it from the regular chapter feed.
Previously, both had the same title, which was confusing—especially for manga with only one chapter—making it feel like navigating through two identical feeds to reach the actual chapter. ~ Ref to end

```conf
# OPDS
server.opdsItemsPerPage = 50 # Range (1 - 5000)
server.opdsEnablePageReadProgress = true
server.opdsMarkAsReadOnDownload = false
server.opdsShowOnlyUnreadChapters = false
server.opdsChapterSortOrder = "DESC" # "ASC", "DESC"
```

If all goes well, hoping to add those in default web-ui as well...
<details>
<summary>Default Web-UI Setting tab with these setttings</summary>

image

</details>

<details>
<summary>"⬇" in the ChapterEntry title for ChapterMetadataFeed</summary>

https://github.com/user-attachments/assets/884064e0-e3e8-420f-92dc-a50dade2db2e
</details>

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
@@ -186,6 +186,7 @@ class SettingsMutation {
         updateSetting(settings.opdsEnablePageReadProgress, serverConfig.opdsEnablePageReadProgress)
         updateSetting(settings.opdsMarkAsReadOnDownload, serverConfig.opdsMarkAsReadOnDownload)
         updateSetting(settings.opdsShowOnlyUnreadChapters, serverConfig.opdsShowOnlyUnreadChapters)
+        updateSetting(settings.opdsShowOnlyDownloadedChapters, serverConfig.opdsShowOnlyDownloadedChapters)
         updateSetting(settings.opdsChapterSortOrder, serverConfig.opdsChapterSortOrder)
     }
 
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
@@ -94,6 +94,10 @@ class SettingsMutation {
 
         // local source
         validateFilePath(settings.localSourcePath, "localSourcePath")
+
+        // opds
+        validateValue(settings.opdsItemsPerPage, "opdsItemsPerPage") { it in 10..5000 }
+        validateValue(settings.opdsChapterSortOrder, "opdsChapterSortOrder") { it in listOf("ASC", "DESC") }
     }
 
     private fun <SettingType : Any> updateSetting(
@@ -177,6 +181,13 @@ class SettingsMutation {
         updateSetting(settings.flareSolverrSessionName, serverConfig.flareSolverrSessionName)
         updateSetting(settings.flareSolverrSessionTtl, serverConfig.flareSolverrSessionTtl)
         updateSetting(settings.flareSolverrAsResponseFallback, serverConfig.flareSolverrAsResponseFallback)
+
+        // opds
+        updateSetting(settings.opdsItemsPerPage, serverConfig.opdsItemsPerPage)
+        updateSetting(settings.opdsEnablePageReadProgress, serverConfig.opdsEnablePageReadProgress)
+        updateSetting(settings.opdsMarkAsReadOnDownload, serverConfig.opdsMarkAsReadOnDownload)
+        updateSetting(settings.opdsShowOnlyUnreadChapters, serverConfig.opdsShowOnlyUnreadChapters)
+        updateSetting(settings.opdsChapterSortOrder, serverConfig.opdsChapterSortOrder)
     }
 
     fun setSettings(input: SetSettingsInput): SetSettingsPayload {
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
@@ -97,7 +97,6 @@ class SettingsMutation {
 
         // opds
         validateValue(settings.opdsItemsPerPage, "opdsItemsPerPage") { it in 10..5000 }
-        validateValue(settings.opdsChapterSortOrder, "opdsChapterSortOrder") { it in listOf("ASC", "DESC") }
     }
 
     private fun <SettingType : Any> updateSetting(
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SettingsType.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SettingsType.kt
@@ -8,6 +8,7 @@
 package suwayomi.tachidesk.graphql.types
 
 import com.expediagroup.graphql.generator.annotations.GraphQLDeprecated
+import org.jetbrains.exposed.sql.SortOrder
 import suwayomi.tachidesk.graphql.server.primitives.Node
 import suwayomi.tachidesk.server.ServerConfig
 import suwayomi.tachidesk.server.serverConfig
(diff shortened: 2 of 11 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/controller/MangaController.kt
```diff
@@ -434,20 +435,26 @@ object MangaController {
...
                 ctx.future {
-                    future { ChapterDownloadHelper.getCbzForDownload(chapterId) }
+                    future { ChapterDownloadHelper.getCbzForDownload(chapterId, shouldMarkAsRead) }
                         .thenApply { (inputStream, fileName, fileSize) ->
                             ctx.header("Content-Type", "application/vnd.comicbook+zip")
                             ctx.header("Content-Disposition", "attachment; filename=\"$fileName\"")
                             ctx.header("Content-Length", fileSize.toString())
-                            ctx.result(inputStream)
+                            if (ctx.method() == HandlerType.HEAD) {
+                                ctx.status(200)
```
Syer10: The InputStream needs to be closed if we are not using it here

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SettingsType.kt
```diff
@@ -226,6 +239,12 @@ class SettingsType(
     override val flareSolverrSessionName: String,
     override val flareSolverrSessionTtl: Int,
     override val flareSolverrAsResponseFallback: Boolean,
+    // opds
+    override val opdsItemsPerPage: Int?,
+    override val opdsEnablePageReadProgress: Boolean?,
+    override val opdsMarkAsReadOnDownload: Boolean?,
+    override val opdsShowOnlyUnreadChapters: Boolean?,
+    override val opdsChapterSortOrder: String?,
```
schroda:
```suggestion
    override val opdsItemsPerPage: Int,
    override val opdsEnablePageReadProgress: Boolean,
    override val opdsMarkAsReadOnDownload: Boolean,
    override val opdsShowOnlyUnreadChapters: Boolean,
    override val opdsChapterSortOrder: String,
```

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
```diff
@@ -102,14 +105,14 @@ object Opds {
...
-                            if (conditions.isEmpty()) (MangaTable.inLibrary eq true) else conditions.reduce { acc, op -> acc and op }
+                            conditions.reduce { acc, op -> acc and op }
                         }.groupBy(MangaTable.id)
                         .orderBy(MangaTable.title to SortOrder.ASC)
                 val totalCount = query.count()
                 val mangas =
                     query
-                        .limit(ITEMS_PER_PAGE)
-                        .offset(((pageNum - 1) * ITEMS_PER_PAGE).toLong())
+                        .limit(serverConfig.opdsItemsPerPage.value)
```
Syer10: Add this to every use of opdsItemsPerPage, or something similar like a function so that we properly get inside the bounds when using it
```suggestion
                        .limit(serverConfig.opdsItemsPerPage.value.coerceIn(10, 5000))
                        .offset(((pageNum - 1) * ITEMS_PER_PAGE).toLong())
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
shirishsaxena: Added. Let me know if it’s not the right way, still getting used to Kotlin and all...

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
```diff
@@ -94,6 +94,10 @@ class SettingsMutation {
 
         // local source
         validateFilePath(settings.localSourcePath, "localSourcePath")
+
+        // opds
+        validateValue(settings.opdsItemsPerPage, "opdsItemsPerPage") { it in 10..5000 }
+        validateValue(settings.opdsChapterSortOrder, "opdsChapterSortOrder") { it in listOf("ASC", "DESC") }
```
schroda: if you use an enum as the type instead of string then graphql would type it accordingly and throw an error if a wrong value is passed in the request

schroda (approved):

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/Opds.kt
Syer10: Looks good

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
shirishsaxena: Would something like this work? https://github.com/Suwayomi/Suwayomi-Server/pull/1392/commits/15f7098290575d888766bd472cfd3cd26d81749b
I'll revert it if that is not the case...

shirishsaxena: I tested the above change a bit using GraphQL and got this error on this line (passing value `ASC` in GraphQL:
```com.typesafe.config.ConfigException$BugOrBroken: bug in method caller: not valid to create ConfigValue from: ASC```

From debugging, I see that the value is correctly deserialized to SortOrder.ASC, but passing the enum instance to ConfigValueFactory.fromAnyRef is causing it to throw an error... it seems fromAnyRef does not support enum instances directly?

https://github.com/Suwayomi/Suwayomi-Server/blob/1d541a30ae03fd69ba049d0611eab41c92785146/AndroidCompat/Config/src/main/java/xyz/nulldev/ts/config/ConfigManager.kt#L115

Adding this check in that method appears to fix the issue, but I’m not sure if there’s a better workaround (if any):

```kotlin
val actualValue = if (value is Enum<*>) value.name else value
val configValue = ConfigValueFactory.fromAnyRef(actualValue)
```

Let me know if I should revert this change or add the above check...

schroda: looks good, your approach is definitely better then what I did, which is handling enums as string values e.g., `webUIChannel`
```
updateSetting(settings.webUIChannel?.name?.lowercase(), serverConfig.webUIChannel)

WebUIChannel.from(serverConfig.webUIChannel.value)
```

shirishsaxena:
I've updated the opdsChapterSortOrder field type to SortOrder enum and confirmed it's working correctly with GraphQL after the small fix mentioned above...
Also added a new config option: showOnlyDownloadedChapters. It wasn't requested, but I figured it'd be a nice to have.

Let me know if I missed anything again :)

<details>
<summary>OPDS Config - Sample</summary>

https://github.com/user-attachments/assets/8097b401-e199-46a5-9f57-797d409dd702

</details>

Syer10 (approved):
I have an idea on how to improve the enum config adapter, but I'll explore that on my own time.

I will merge this once I get the launcher ready

PR #1407, opened 2025-05-25
Fix: Validate zipEntry directories during extension asset decompression

EdgeAtZero:
IpInputStream.nextEntry may be a folder. If it is not checked, the folder will be output as a file by mistake, which will cause the contents in the subsequent folders to be unable to be copied, resulting in the failure of the extension installation.
Although most extension apk files will not be like this, I think it is still necessary to modify the code here to add compatibility
屏幕截图 2025-05-26 073743

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
@@ -240,7 +240,7 @@ object Extension {
         ZipInputStream(apkFile.inputStream()).use { zipInputStream ->
             var zipEntry = zipInputStream.nextEntry
             while (zipEntry != null) {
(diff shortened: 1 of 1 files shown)


EdgeAtZero:
This is my first time submitting a PR. Due to my editing error, there are character errors in the file I just submitted, so I submitted it again. Sorry for bothering you. Thank you.

Robonau:
you could have just committed the fixes to the branch, it would have updated the pr

EdgeAtZero:
> you could have just committed the fixes to the branch, it would have updated the pr

Sorry, I don't quite understand what you mean by committed, My English is not very good.

Robonau:
it doesn't matter

EdgeAtZero:
> it doesn't matter

Got it, thanks for your help!

Syer10 (approved):

PR #1430, opened 2025-06-01
Feature/backup suwayomi data

schroda:


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/backup/proto/ProtoBackupImport.kt
```diff
@@ -176,12 +188,20 @@ object ProtoBackupImport : ProtoBackupBase() {
 
         val validationResult = validate(backup)
 
-        restoreAmount = backup.backupManga.size + 1 // +1 for categories
+        restoreAmount = backup.backupManga.size + 3 // +1 for categories, +1 for meta
```
Syer10: Why was this bumped by 2 instead of 1?

schroda: Forgot to decrease after removing something.
I kept it and added `settings` as an extra step. `restoreAmount` was also never used anywhere, updated that as well

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/backup/proto/ProtoBackupExport.kt
```diff
@@ -287,31 +305,135 @@ object ProtoBackupExport : ProtoBackupBase() {
...
+    private fun backupGlobalMeta(flags: BackupFlags): Map<String, String> {
+        if (!flags.includeClientData) {
+            return emptyMap()
+        }
+
+        return GlobalMeta.getMetaMap()
+    }
+
+    private fun backupServerSettings(flags: BackupFlags): BackupServerSettings? {
+        if (!flags.includeClientData) {
```
Syer10:
```suggestion
        if (!flags.includeServerSettings) {
```

Syer10 (changes requested):

Syer10 (approved):

Issue #161, opened 2021-07-29
[Meta] Extensions known that don't work


AriaMoradi:
- MyReadingManga (EN) -> `Stub!` Exception on  `android.webkit.WebSettings.getDefaultUserAgent` (might be fixed?)
- Manhuaren/漫画人 (ZH) ->`Stub!` Exception on `android.text.format.DateFormat.format`
- Imperfect Comics -> needs investigation

Issue #1101, opened 2024-10-15
[Bug] COLAMANGA cant read manga


bk1412:
## Device information
- Suwayomi-Server version: v1.1.1-r1535
- Server Operating System: Ubuntu 23.04
- Server Desktop Environment: N/A
- Server JVM version: Java 8
- Client Operating System: Win10
- Client Web Browser: Google Chrome 129.0.6668.100

## Steps to reproduce
1. open COLAMANGA
2. choose any manga
3. choose any chapter

### Expected behavior
It should be open a comic picture after.

### Actual behavior
cant read,show " Stub!"

## Other details

Same extension test on mihon,everthing is working well

logs:

10月 15 17:10:15 tokyo suwayomi-server[1979945]: 17:10:15.034 [DefaultDispatcher-worker-7] DEBUG notprivacysafe.graphql.execution.Execution -- Executing '9bfbaec9-f257-49b7-80bb-8bf1403c3486' query operation: 'MUTATION' using 'graphql.e>
10月 15 17:10:15 tokyo suwayomi-server[1979945]: 17:10:15.035 [DefaultDispatcher-worker-7] DEBUG graphql.execution.instrumentation.dataloader.DataLoaderDispatcherInstrumentation -- Dispatching data loaders ([CategoryMetaDataLoader, Last>
10月 15 17:10:15 tokyo suwayomi-server[1979945]: 17:10:15.160 [DefaultDispatcher-worker-16] DEBUG eu.kanade.tachiyomi.network.NetworkHelper -- --> GET https://www.colamanga.com/manga-od825111/1/381.html h2
10月 15 17:10:15 tokyo suwayomi-server[1979945]: 17:10:15.467 [DefaultDispatcher-worker-16] DEBUG eu.kanade.tachiyomi.network.NetworkHelper -- <-- 200 https://www.colamanga.com/manga-od825111/1/381.html (306ms, unknown-length body)
10月 15 17:10:15 tokyo suwayomi-server[1979945]: 17:10:15.472 [DefaultDispatcher-worker-16] ERROR suwayomi.tachidesk.graphql.AsDataFetcherResult -- asDataFetcherResult: failed due to
10月 15 17:10:15 tokyo suwayomi-server[1979945]: java.lang.RuntimeException: Stub!
10月 15 17:10:15 tokyo suwayomi-server[1979945]:         at android.os.Looper.getMainLooper(Looper.java:30)
[...]

PR #1434, opened 2025-06-07
Implement WebView via Playwright

cpiber:
Partially addresses #161

This also contains an implementation of Looper and MessageQueue, since extensions typically put WebView handling on the loop.

All stubs and interfaces are based on recent upstream files as seen on cs.android.com. Thus they do not 100% match the version in the stubs, but since they're backward compatible that should be fine.

This enables the Cubari extension:
20250607_13h06m05s_grim

There seems to be a discrepancy somewhere though; an invalid URL should show the message "Unable to parse. Is your query in the format of cubari:<source>/<slug>?", instead we get the following backtrace:

```
12:47:54.572 [DefaultDispatcher-worker-10] ERROR suwayomi.tachidesk.graphql.AsDataFetcherResult -- asDataFetcherResult: failed due to
kotlinx.serialization.json.internal.JsonDecodingException: Cannot read Json element because of unexpected end of the input at path: $
JSON input: 
        at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
        at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:32)
        at kotlinx.serialization.json.internal.AbstractJsonLexer.fail(AbstractJsonLexer.kt:587)
        at kotlinx.serialization.json.internal.AbstractJsonLexer.fail$default(AbstractJsonLexer.kt:585)
        at kotlinx.serialization.json.internal.JsonTreeReader.read(JsonTreeReader.kt:107)
        at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeJsonElement(StreamingJsonDecoder.kt:50)
        at kotlinx.serialization.json.JsonElementSerializer.deserialize(JsonElementSerializers.kt:50)
        at kotlinx.serialization.json.JsonElementSerializer.deserialize(JsonElementSerializers.kt:27)
        at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:69)
        at kotlinx.serialization.json.Json.decodeFromString(Json.kt:149)
        at kotlinx.serialization.json.Json.parseToJsonElement(Json.kt:179)
        at eu.kanade.tachiyomi.extension.all.cubari.Cubari.searchMangaParse(Unknown Source)
        at eu.kanade.tachiyomi.extension.all.cubari.Cubari.access$searchMangaParse(Unknown Source)
        at eu.kanade.tachiyomi.extension.all.cubari.Cubari$fetchSearchManga$2.invoke(Unknown Source)
        at eu.kanade.tachiyomi.extension.all.cubari.Cubari$fetchSearchManga$2.invoke(Unknown Source)
        at eu.kanade.tachiyomi.extension.all.cubari.Cubari.fetchSearchManga$lambda$15(Unknown Source)
        at eu.kanade.tachiyomi.extension.all.cubari.Cubari.$r8$lambda$PnjAk5VfRyZhxUCoiHrOSDkWIgE(Unknown Source)
        at eu.kanade.tachiyomi.extension.all.cubari.Cubari$$ExternalSyntheticLambda7.call(Unknown Source)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:69)
        at rx.internal.operators.OnSubscribeDoOnEach$DoOnEachSubscriber.onNext(OnSubscribeDoOnEach.java:101)
        at eu.kanade.tachiyomi.network.OkHttpExtensionsKt$asObservable$1$requestArbiter$1.request(OkHttpExtensions.kt:38)
        at rx.Subscriber.setProducer(Subscriber.java:211)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.setProducer(OnSubscribeMap.java:102)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.setProducer(OnSubscribeMap.java:102)
        at rx.Subscriber.setProducer(Subscriber.java:205)
        at eu.kanade.tachiyomi.network.OkHttpExtensionsKt.asObservable$lambda$0(OkHttpExtensions.kt:56)
        at rx.Observable.unsafeSubscribe(Observable.java:10327)
        at rx.internal.operators.OnSubscribeDoOnEach.call(OnSubscribeDoOnEach.java:41)
        at rx.internal.operators.OnSubscribeDoOnEach.call(OnSubscribeDoOnEach.java:30)
        at rx.Observable.unsafeSubscribe(Observable.java:10327)
        at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:48)
        at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:33)
        at rx.Observable.unsafeSubscribe(Observable.java:10327)
        at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:48)
        at rx.internal.operators.OnSubscribeMap.call(OnSubscribeMap.java:33)
        at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:48)
        at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:30)
        at rx.Observable.subscribe(Observable.java:10423)
        at rx.Observable.subscribe(Observable.java:10390)
        at suwayomi.tachidesk.manga.impl.util.lang.RxCoroutineBridgeKt.awaitOne(RxCoroutineBridge.kt:28)
        at suwayomi.tachidesk.manga.impl.util.lang.RxCoroutineBridgeKt.awaitSingle(RxCoroutineBridge.kt:22)
        at eu.kanade.tachiyomi.source.CatalogueSource$DefaultImpls.getSearchManga(CatalogueSource.kt:41)
        at eu.kanade.tachiyomi.source.online.HttpSource.getSearchManga(HttpSource.kt:28)
        at suwayomi.tachidesk.graphql.mutations.SourceMutation$fetchSourceManga$1.invokeSuspend(SourceMutation.kt:128)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:124)
        at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:820)
Caused by: rx.exceptions.OnErrorThrowable$OnNextValue: OnError while emitting onNext value: okhttp3.Response.class
        at rx.exceptions.OnErrorThrowable.addValueAsLastCause(OnErrorThrowable.java:118)
        at rx.internal.operators.OnSubscribeMap$MapSubscriber.onNext(OnSubscribeMap.java:73)
        ... 33 common frames omitted
```
The extension seems to assume errors are translated into empty lists, and this appears to work on Mihon
https://github.com/keiyoushi/extensions-source/blob/bdb776bfc9844a0954dbb48a6929f3e605217885/src/all/cubari/src/eu/kanade/tachiyomi/extension/all/cubari/Cubari.kt#L245-L256

Most other extensions that use WebView are **not yet** enabled by this. This implementation stubs most settings, some of which are used. More importantly, events like `shouldInterceptRequest` are not yet implemented. I would appreciate a review of the existing code while I see into implementing more parts.


cpiber:
#1101 ColaManga is difficult to support, since it's very weirdly implemented, which seems incompatible with Playwright (SecurityException on accessing `document.cookie`, by embedding the code inside the page instead of using ~~`evaluateJavascript` the bindings do not work~~, etc)

Schale.network/Koharu seems to require user interaction, so also not supported (opening in the browser, there seems to be a Captcha that needs solving, but not Cloudflare, so flaresolverr fails).

InfinityScans could not test since Cloudflare bypass does not like it (error 500), but should work

LibGroup (HentaiLib, MangaLib, YaoiLib) works well enough to load manga list, but some require a login to not get a 404, which requires user interaction and is thus not supported.

Bokugen: Supported

JapScans not tested due to error 500, but likely same as #1101

Comikey is broken. ~~Not sure why, on error it says it cannot call the exposed function (works in e.g. Cubari), otherwise~~ never tries anything, possibly due to broken source. Requires user interaction to unlock chapters anyways, so not useful

That should be all sources that use WebView

cpiber:
I was wrong about the method exposing, it never worked, Cubari just didn't actually need it apparently.
I've also now fixed a bug where the MessageQueue went to sleep and never woke up, so now the error message in Cubari also works properly.

No change in source support from that though sadly.

bk1412:
> #1101 ColaManga is difficult to support, since it's very weirdly implemented, which seems incompatible with Playwright (SecurityException on accessing `document.cookie`, by embedding the code inside the page instead of using ~`evaluateJavascript` the bindings do not work~, etc)
> 
> Schale.network/Koharu seems to require user interaction, so also not supported (opening in the browser, there seems to be a Captcha that needs solving, but not Cloudflare, so flaresolverr fails).
> 
> InfinityScans could not test since Cloudflare bypass does not like it (error 500), but should work
> 
> LibGroup (HentaiLib, MangaLib, YaoiLib) works well enough to load manga list, but some require a login to not get a 404, which requires user interaction and is thus not supported.
> 
> Bokugen: Supported
> 
> JapScans not tested due to error 500, but likely same as #1101
> 
> Comikey is broken. ~Not sure why, on error it says it cannot call the exposed function (works in e.g. Cubari), otherwise~ never tries anything, possibly due to broken source. Requires user interaction to unlock chapters anyways, so not useful
> 
> That should be all sources that use WebView

COLAMANGA involves a lot of encryption operations, relying primarily on WebView to execute decryption JavaScript and obtain the final image address links. I once attempted to create a dedicated COLAMANGA plugin for Suwayomi, but I ultimately gave up for the following reasons:

Regarding the WebView issue, I couldn’t find a good solution. I tried exploring alternative components and investigated various approaches, such as JavaFX’s WebView for Java. My final investigation led me to JxBrowser (a paid component, though there seem to be cracked versions available), but I put this on hold and didn’t take further action.
After the above, on one occasion when I used COLAMANGA’s official website, I noticed that even under normal browsing conditions, the site would trigger anti-crawler blocking operations, resulting in a poor reading experience. The experience of downloading an entire manga using Suwayomi was even worse (unimaginably bad…).
I hope my investigation into WebView can be helpful to someone. From my research, JxBrowser seems like a potential good alternative (if the payment issue can be resolved). It appears to cover many of the underlying functions similar to WebView and might be a better option than Playwright.

cpiber:
> COLAMANGA involves a lot of encryption operations, relying primarily on WebView to execute decryption JavaScript and obtain the final image address links. I once attempted to create a dedicated COLAMANGA plugin for Suwayomi, but I ultimately gave up for the following reasons:

From what I can tell, the extension has that covered. It just does it in a weird way. If it just used `loadUrl` and `evaluateJavascript` like other extensions, it might be easier to support. That said, I haven't looked in JxBrowser. I also only recently noticed #1154, which uses JCef as a browser, which might have better support for this as well.

@syer10 @schroda Opinions? Should I rewrite with another underlying implementation?

AndroidCompat/src/main/java/android/webkit/PlaywrightWebSettings.java
```diff
@@ -0,0 +1,534 @@
...
+
+    public void setDefaultTextEncodingName(String encoding) {
+        throw new RuntimeException("Stub!");
+    }
+
+    public String getDefaultTextEncodingName() {
+        throw new RuntimeException("Stub!");
+    }
+
+    public void setUserAgentString(@Nullable String ua) {
```
Syer10: Do we call this, or does the extension? If its the extension we need to change the default to NetworkHelper.userAgent, either by moving the userAgent MutableStateFlow to AndroidCompat or by setting a system property.

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SettingsType.kt
```diff
@@ -208,6 +217,10 @@ class SettingsType(
     override val autoDownloadIgnoreReUploads: Boolean?,
     // extension
     override val extensionRepos: List<String>,
+    // playwright webview
+    override val playwrightBrowser: String?,
+    override val playwrightWsEndpoint: String?,
+    override val playwrightSandbox: Boolean?,
```
Syer10:
```suggestion
    override val playwrightBrowser: String,
    override val playwrightWsEndpoint: String,
    override val playwrightSandbox: Boolean,
```

server/src/main/resources/server-reference.conf
```diff
@@ -32,6 +32,11 @@ server.extensionRepos = [
   # an example: https://github.com/MY_ACCOUNT/MY_REPO/tree/repo
 ]
 
+# playwright webview
+server.playwrightBrowser = "chromium"
```
Syer10: Add a comment with possible browser values

AndroidCompat/build.gradle.kts
```diff
@@ -21,6 +21,8 @@ dependencies {
     implementation(libs.bundles.shared)
     testImplementation(libs.bundles.sharedTest)
 
+    implementation("com.microsoft.playwright:playwright:1.51.0")
```
Syer10: This should be moved to gradle/libs.versions.toml

Syer10 (commented):
This is looking great! While KCEF would be preferred, Playwright is a good alternative, we have used it previously as an attempt to bypass Cloudflare, you should checkout the previous PRs before it was eventually removed. https://github.com/Suwayomi/Suwayomi-Server/pulls?q=is%3Apr+playwright+is%3Aclosed

One bad thing about Playwright is that it downloads Chromium/the browser all on its own, we previously tried adding it to the download package but that just sorta worked, it wasn't worth it, we need a pre-download initializer at least so that Suwayomi doesn't just stop in the middle of a request while it downloads.
Docker also has an issue where it will download to a unmounted folder, so it redownloads the browser on every re-create of the container, changing the browser download to somewhere in the root directory would be preferred.

For cloudflare issues we can re-add the Cloudsflare bypass scripts we tested with Playwright previously to help it bypass Cloudflare(see https://github.com/Suwayomi/Suwayomi-Server/pull/456), it just wasn't the optimial choice due to the above issues so it was removed.

JxBrowser is not possible since we won't distribute pirated or paid versions of any library in the app.

In general well done! I'm looking forward to this being improved so we can merge it.

cpiber:
> While KCEF would be preferred, Playwright is a good alternative, we have used it previously as an attempt to bypass Cloudflare, you should checkout the previous PRs before it was eventually removed. [Suwayomi/Suwayomi-Server/pulls (playwright is:closed)](https://github.com/Suwayomi/Suwayomi-Server/pulls?q=sort%3Aupdated-desc+is%3Apr+playwright+is%3Aclosed)

Yup, I saw that, I also say the use of htmlunit. I'll look into KCef, Playwright does seem kinda meh. In Java, it also simply spawns a browser *and* a node server, and uses IPC to communicate with the playwright server, which then communicates with the browser.

> One bad thing about Playwright is that it downloads Chromium/the browser all on its own, we previously tried adding it to the download package but that just sorta worked, it wasn't worth it, we need a pre-download initializer at least so that Suwayomi doesn't just stop in the middle of a request while it downloads.

That's true. It also downloads all browsers, not just the selected one, and it auto-updates.

> For cloudflare issues we can re-add the Cloudsflare bypass scripts we tested with Playwright previously to help it bypass Cloudflare(see https://github.com/Suwayomi/Suwayomi-Server/pull/456), it just wasn't the optimial choice due to the above issues so it was removed.

I don't think Playwright will work better than flaresolverr. I only mentioned Cloudflare issues because I was seeing timeouts and error 500 from the Cloudflare interceptor, I haven't debugged if there actually are issues.

> In general well done! I'm looking forward to this being improved so we can merge it.

Thanks :) Funnily enough, for Cubari (my main reason for implementing), just having some code there is enough, since it uses the regular HTTP response for most stuff.

AndroidCompat/src/main/java/android/webkit/PlaywrightWebSettings.java
cpiber: The extension does.

Good to know, will change that. I think a setter is better, and it's also how I solved configuring the browser.

AndroidCompat/build.gradle.kts
cpiber: Ups, yeah, did that for testing, forgot to move.

Syer10:
> That's true. It also downloads all browsers, not just the selected one, and it auto-updates.

True, I forgot about that! We had a fix for this previously in one of those PRs I did.

> I don't think Playwright will work better than flaresolverr. I only mentioned Cloudflare issues because I was seeing timeouts and error 500 from the Cloudflare interceptor, I haven't debugged if there actually are issues.

I agree, just commening that bypassing cloudflare using Playwright was sorta possible for sources where you are running into Cloudflare issues.

> Yup, I saw that, I also say the use of htmlunit. I'll look into KCef, Playwright does seem kinda meh. In Java, it also simply spawns a browser and a node server, and uses IPC to communicate with the playwright server, which then communicates with the browser.

HtmlUnit wasn't the most updated last I checked, I think there were multiple JS errors because it didn't implement a lot of the newer JS standards. I am fine with the extra headroom Playwright uses if we need to continue using it. Otherwise feel free to explore KCEF.

> Thanks :) Funnily enough, for Cubari (my main reason for implementing), just having some code there is enough, since it uses the regular HTTP response for most stuff.

Thats pretty funny, glad you have it working at least 😄

cpiber:
I just realized I forgot to report Yidan, not 100% sure, seems to sometimes work, but it likely also needs user interaction since it wants the user id.

I have now also implemented the webview via KCEF. For now, both exist, but only KCEF is accessible. Please choose whether you want both, or just keep one.

KCEF has its upsides and downsides. The whole model is a bit simpler with KCEF (directly via JNI instead of IPC and node), and it seems more suited to WebView, but it also has it's downsides. Specifically, the user needs some JNI libraries installed. Of those, libgluegen_rt.so is not available anymore on Ubuntu (only libgluegen2_rt.so); in general using KCEF may require some user information to get it properly set up.

Also KCEF is a bitch, so many ways to shoot yourself in the foot and just not get any callbacks...

No change in extension support after rewrite, though Colamanga at least now has no more problems due to cookies (but __cr is undefined, so no decryption possible)

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -187,7 +227,10 @@ fun applicationSetup() {
     androidCompat.startApp(app)
 
     // Initialize NetworkHelper early
-    Injekt.get<NetworkHelper>()
+    Injekt.get<NetworkHelper>().userAgentFlow.onEach {
+        SystemProperties.set("http.agent", it)
+    }
```
Syer10: Need to launch the listener
```suggestion
    }.launchIn(GlobalScope)
```

AndroidCompat/src/main/java/xyz/nulldev/androidcompat/AndroidCompatInitializer.kt
```diff
@@ -17,6 +19,8 @@ class AndroidCompatInitializer {
             SystemConfigModule.register(GlobalConfigManager.config),
         )
 
+        WebView.setProviderFactory({ view: WebView -> KcefWebViewProvider(view) })
```
Syer10, line 22: I would move this to ServerSetup and make it a user configurable setting, so that a user can choose Playwright or KCEF

Syer10 (commented):
Its looking good! A few comments, I like that you added both implementations, we can now make it user configurable with KCEF as a default

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
cpiber: Thanks, I have never worked with StateFlow

AndroidCompat/src/main/java/xyz/nulldev/androidcompat/AndroidCompatInitializer.kt
cpiber: Done. I had to leave it there because when trying to use those methods in ServerSetup, I got some weird linking errors.

Do we have to update android-lib for this? It seemed like it was complaining about differences in the WebView stub and the new compat implementation

```
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:117:33 Cannot access 'Callback' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:117:33 Cannot access 'AccessibilityEventSource' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:117:33 Cannot access 'ViewParent' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:117:33 Cannot access 'ViewManager' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:117:33 Cannot access 'OnGlobalFocusChangeListener' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:118:38 Cannot access 'Callback' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:118:38 Cannot access 'AccessibilityEventSource' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:118:38 Cannot access 'ViewParent' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:118:38 Cannot access 'ViewManager' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
e: file:///.../Suwayomi/Suwayomi-Server/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt:118:38 Cannot access 'OnGlobalFocusChangeListener' which is a supertype of 'WebView'. Check your module classpath for missing or conflicting dependencies.
```

Syer10: Looks like it, I will update the android library after I merge this. This solution will work though so its fine.

Syer10 (approved):
Looks good! I will merge it after I get the docker env variables and launcher settings added. I will also verify that KCEF works fine in the docker container.

cpiber:
Thanks! Though I guess it won't easily. It's based on Ubuntu noble, which is also what I'm using. Most libs are available in `libjogl2-jni`, but `libgluegen2-jni` only has that `libgluegen2_rt.so`, not `libgluegen_rt.so`, I had to compile that myself

Syer10:
One thing I did forget about is how big Playwright makes the jar file, a 350 MB jar is huge

cpiber:
Hm, that's true, should we just remove it entirely?

Syer10:
Ya, at this point its not worth the size that it makes the Jar. KCEF should be enough for most users anyway.

In regards to getting KCEF working with Ubuntu, would we be able to add that .so file to the jar and load it with System.loadLibrary()? I don't have much expierence with native libraries on Linux. All the issues I've encountered I just needed to install a apt dependency to fix it

cpiber:
> Ya, at this point its not worth the size that it makes the Jar. KCEF should be enough for most users anyway.

Agreed. Maybe an option would be conditionally including it, but I think that will become difficult to maintain and most users won't build from source...

> In regards to getting KCEF working with Ubuntu, would we be able to add that .so file to the jar and load it with System.loadLibrary()? I don't have much expierence with native libraries on Linux. All the issues I've encountered I just needed to install a apt dependency to fix it

KCEF will do that automatically. When the browser is created, it looks for the libraries by itself in several places. I don't know if it can be included in the jar, locally I put them in `server/natives/linux-amd64`. Quick search says you'd have to include it as a resource and extract manually ([source](https://stackoverflow.com/a/19858713))
In the docker, we could just install them system-wide, and anyone looking to run the jar directly would have to deal with this themselves.

cpiber:
>  In the docker, we could just install them system-wide, and anyone looking to run the jar directly would have to deal with this themselves.

If this approach is fine, I can update the Dockerfile tomorrow

Syer10:
> > In the docker, we could just install them system-wide, and anyone looking to run the jar directly would have to deal with this themselves.
> 
> If this approach is fine, I can update the Dockerfile tomorrow

Yep, that should work fine, thanks!

cpiber:
Updating the Dockerfile took me way longer than initially expected. It seems JCEF really expects there to be a working graphical environment. I initially tried to get CEF to launch in headless mode, for which it took me quite a while to even figure out how to pass those arguments through all the layers (KCEF, then JCEF, then JCEF's native code). Turns out, even if CEF supports ozone, JCEF (at least with the version of KCEF we're using) does not; it assumes there is a X11 server around to create a temporary window. Thus, I opted for using Xvfb, a light-weight X11 implementation. It does not use much resources, but it also does not support any GPU acceleration etc, and the container prints a nice backtrace when the browser is created (but it works, so ignore that ig)

<details>

```
Caught handled GLException: EGLGLXDrawableFactory - Could not initialize shared resources for EGLGraphicsDevice[type .egl, v0.0.0, nativeDisplayID 0x0, connection :0, unitID 0, handle 0x0, owner true, ResourceToolkitLock[obj 0x4d6d5c69, isOwner true, <50ea54ab, 94b0632>[count 1, qsz 0, owner <Thread-0-SharedResourceRunner>]]] on thread Thread-0-SharedResourceRunner
    [0]: jogamp.opengl.egl.EGLDrawableFactory$SharedResourceImplementation.createSharedResource(EGLDrawableFactory.java:555)
    [1]: jogamp.opengl.SharedResourceRunner.run(SharedResourceRunner.java:353)
    [2]: java.base/java.lang.Thread.run(Unknown Source)
Caused[0] by GLException: Failed to created/initialize EGL display incl. fallback default: native 0x0, error 0x3008/0x3005 on thread Thread-0-SharedResourceRunner
    [0]: jogamp.opengl.egl.EGLDisplayUtil.eglGetDisplayAndInitialize(EGLDisplayUtil.java:381)
    [1]: jogamp.opengl.egl.EGLDisplayUtil.access$300(EGLDisplayUtil.java:61)
    [2]: jogamp.opengl.egl.EGLDisplayUtil$1.eglGetAndInitDisplay(EGLDisplayUtil.java:404)
    [3]: com.jogamp.nativewindow.egl.EGLGraphicsDevice.open(EGLGraphicsDevice.java:228)
    [4]: jogamp.opengl.egl.EGLDrawableFactory$SharedResourceImplementation.createEGLSharedResourceImpl(EGLDrawableFactory.java:569)
    [5]: jogamp.opengl.egl.EGLDrawableFactory$SharedResourceImplementation.createSharedResource(EGLDrawableFactory.java:553)
    [6]: jogamp.opengl.SharedResourceRunner.run(SharedResourceRunner.java:353)
    [7]: java.base/java.lang.Thread.run(Unknown Source)
Caught handled GLException: EGLGLXDrawableFactory - Could not initialize shared resources for X11GraphicsDevice[type .x11, connection :0, unitID 0, handle 0x0, owner false, ResourceToolkitLock[obj 0x6eefb7a2, isOwner true, <1d7f61af, 60fbc586>[count 1, qsz 0, owner <Thread-0-SharedResourceRunner>]]] on thread Thread-0-SharedResourceRunner
    [0]: jogamp.opengl.egl.EGLDrawableFactory$SharedResourceImplementation.createSharedResource(EGLDrawableFactory.java:555)
    [1]: jogamp.opengl.SharedResourceRunner.run(SharedResourceRunner.java:353)
    [2]: java.base/java.lang.Thread.run(Unknown Source)
Caused[0] by GLException: Failed to created/initialize EGL display incl. fallback default: native 0x0, error 0x3008/0x3005 on thread Thread-0-SharedResourceRunner
    [0]: jogamp.opengl.egl.EGLDisplayUtil.eglGetDisplayAndInitialize(EGLDisplayUtil.java:381)
    [1]: jogamp.opengl.egl.EGLDisplayUtil.access$300(EGLDisplayUtil.java:61)
    [2]: jogamp.opengl.egl.EGLDisplayUtil$1.eglGetAndInitDisplay(EGLDisplayUtil.java:404)
    [3]: com.jogamp.nativewindow.egl.EGLGraphicsDevice.open(EGLGraphicsDevice.java:228)
    [4]: jogamp.opengl.egl.EGLDrawableFactory$SharedResourceImplementation.createEGLSharedResourceImpl(EGLDrawableFactory.java:569)
    [5]: jogamp.opengl.egl.EGLDrawableFactory$SharedResourceImplementation.createSharedResource(EGLDrawableFactory.java:553)
    [6]: jogamp.opengl.SharedResourceRunner.run(SharedResourceRunner.java:353)
    [7]: java.base/java.lang.Thread.run(Unknown Source)
```

</details>

> Most libs are available in `libjogl2-jni`, but `libgluegen2-jni` only has that `libgluegen2_rt.so`, not `libgluegen_rt.so`, I had to compile that myself

Turns out just renaming the library is enough. Ubuntu just packages it under a different name, because gluegen is on version 2 (and has been for a long time). Thanks, I guess. Also Java does not find the libraries where they are in the Ubuntu packages, so I just copy them somewhere it can find them, that should not cost much since they're small, but that might be a point of improvement.

cpiber:
Btw do we want to update the Readme with information on required dependencies for KCEF? The logs should show everything that is relevant, but it may come as a surprise to users.
If so, I have no idea how to install those on Windows or MacOs :sweat_smile:

Syer10:
KCEF works fine on Windows with no dependencies needed, thats where my testing was done and I've expierenced no issues. I'm not sure about MacOS but I wouldn't be suprised if no modifications are needed either. I can test on my Mac whenever I have time, so just add the Linux instructions to the README if you can.

cpiber:
@Syer10 I believe Android-stub isn't updated yet, do you need anything for that from me? Or is it not necessary?

Syer10:
No you are good, I just forgot to do it yesterday. I'll do it in a few hours

Syer10:
@cpiber Android-stub is now updated 👍

PR #1455, opened 2025-06-20
Implement Bitmap.copy, text layouting

cpiber:
I initially thought `Bitmap.copy` was all that is required, turns out it needed an entire text layout engine. I kept as much of the Android code as I could, only replacing the actual Font layout/measuring with AWT. This means, the constructor of StaticLayout computes the line information via AWT, basically everything else (centering, etc) is stock Android.

Fixes #1350
Fixes #1372

Ours:
image

Mihon:
image

Font is obviously not correct, though I also don't think the extension sets that

Also adds support for extensions that allow rendering Comments like DMZJ and DM5

image

PR #1459, opened 2025-06-22
Paint: Support Typeface

cpiber:
Extensions try to use custom fonts, which we previously ignored.

Now:
image

See #1455 for a before and Mihon comparison.
Our (Java's) implementation does not support hyphenation, which seems to be why this needs an additional line.


Syer10 (approved):

Syer10:
Looks good

PR #642, opened 2023-08-06
[WIP] Use Selenium as a WebView in the browser

Syer10:
The goal is to allow logging into sources eventually

PR #1486, opened 2025-06-29
Browser Webview

### LINKED ISSUE #968: [Feature Request] Support per extension authentication (or persistent cookies)
## What feature should be added to Suwayomi?

Some sites that some extensions use to obtain data, now requires authentication to access the chapters and respective images. In that case, to make the image download work in the original Android application, you need to manually go to the WebView button at least once, load the website, log in, and start the update process, and finally, the image download will work correctly.

Does a similar functionality already exist in [Suwayomi-Server](https://github.com/Suwayomi/Suwayomi-Server)?

Making Suwayomi-Server capable of executing the extensions and the requests made by them will use cookies or another way to persist the authentication done manually in something similar to the Webview button of the original Android application (that Android application that was forcibly discontinued by the developers)

### LINKED ISSUE #1336: zerobyw extension can't login within suwayomi-server
### Steps to reproduce

1.install zerobyw extension
2.select random content
3.start reading latest chapter 
4. return error:仅限用户观看，请先登陆（registered used only ，login first）

### Expected behavior

should have some place to fulfill my login info，a webview bottom appear in mihon case

### Actual behavior

only show an error message

### Suwayomi-Server version

v1.1.1719

### Server operating system

Macos 15.3 Orbstack

### Server Desktop Environment

_No response_

### Server JVM version

-

### Used client name

-

### Client version

-

### Used web browser

-

### Client operating system

-

### Other details

_No response_

### Acknowledgements

- [x] I have tried the troubleshooting guide described in [README.md](https://github.com/Suwayomi/Suwayomi-Server?tab=readme-ov-file#troubleshooting-and-support)
- [x] I understand that **Suwayomi does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

### LINKED ISSUE #1396: WebView feature

The open site option for Suwayomi doesn't really solve some issues that can be solved with webview running within the server. Openning sources in WebView allows the user to login into said source without support from the extension and can solve Cloudflare issues. When clicking on open site on the client side using WebUI, it opens a new tab on the client site web browser; logging in through the new openned tab does not log in or cache the logged in cookie on the server side, users can't solve Cloudflare issues this way nor be able to login into sources.

Implementing a WebView on the server side running Suwayomi-Server fixes the issue stated above.

### Other details

There are working examples of Open in WebView in Mihon app.

### Acknowledgements

cpiber:
https://github.com/user-attachments/assets/7ed64da0-fe15-48d8-b7e0-34d481019d6e

Partially based on #642

Closes #1396
Closes #968
Closes #1336

Though this is far from perfect, I couldn't get TappyToon to actually open a locked chaper, due to issues in the rest of the cookie system (more below)

---

I originally tried to implement this by sending HTML and having the client do the rendering (see earlier commits). That would be the most responsive and user-friendly option. Got that working fairly well, with one major downside: Captchas. Those are implemented in iframes, and I couldn't find a good way to transfer those. So I had to switch to image-based, like @Syer10 

Rendering is fairly inefficient at the moment:
- I send the frames as often as CEF renders, which can be quite often for animations
- I ignore the dirty region (CEF tells exactly which part of the viewport needs to be refreshed) and transfer everything to the BufferedImage
- I always send the entire image, compressed

Regarding cookies:
I had multiple issues with the cookie store. Very often, the store would just kill itself (empty file) or produce an invalid file, which meant most of my cookies weren't loaded on the next start.
That said, with just this WebView, I managed to persist the TappyToon login over multiple page reloads, and one or two server restarts (CEF does not store the cookies, so the page reload would use the in-memory cookie store from NetworkHelper).
On the other hand, as soon as I tried to use the TappyToon source, the WebView showed a half-logged-in state. TappyToon stores the email for quite a bit longer than actual logins, so I had the situation where it knew who I was, but didn't like my login. Not sure if this is a problem in the source, or if the cookies aren't applied correctly in the client. TappyToon uses quite a few cookies on both tappytoon.com and www.tappytoon.com...


Syer10:
@cpiber I did some digging and found that the CefCookieManager is an abstract class, I created an implementation that would wrap our CookieStore, but I haven't found how to inject it into CEF yet

cpiber:
I don't think you can, this seems pretty hardcoded: https://github.com/JetBrains/jcef/blob/f51082eaf12d9966452bbb3b44e3e8c67ec216e4/java/org/cef/network/CefCookieManager.java#L38-L45

It's abstract because they support remote (out-of-process) rendering, at least in newer versions of JCEF. Would be neat if you could manage that though

Syer10:
Well there goes that idea. Looking closer this is probably actually for interacting directly with the cef cookie store thats connected to the browser, not used for getting stored cookies for the browser.

cpiber:
One more thing I forgot: There's a bug I couldn't manage to fix. Enter is a bit wonky, in the video it works to press Enter to login, but that's because TappyToon is using JS to handle that. For regular forms, I couldn't get it to trigger the submit... I found a few related posts, but no solution

cpiber:
Saw your commit e83296225bf497519457a86904286b006e7ddced, sorry about that, that can actually be removed, I only had that for testing because I wasn't sure if I was rendering it right in the browser, should I delete it?

Syer10:
Yep, feel free to delete it

Syer10:
@cpiber I did find a few cookie issues, one with the expiry age. Plus we were doing 1 extra cookie conversion we didn't need to be. I beleive this solution will be more stable. Let me know if you get more issues. I'm not sure if my fixes have fixed the file writing empty though.

cpiber:
Thanks for that, I really wasn't looking forward to debugging that. I'll test a few things a bit later

Syer10:
I just noticed the mouse wheel is actually inverted, is that intentional or a bug?

cpiber:
On my system it was inverted, so I negated the delta. This may have been a bit too crude... Maybe `deltaY` on `wheel` is always absolute and ignores natural scroll setting? Not sure how we'd fix that...

cpiber:
From what I can tell, the browser sends an event which has inverted delta depending on the natural scroll setting, but if we just send that delta, the WebView will invert a second time, resulting in the wrong direction. What system are you using? If MacOs, can you check this: https://stackoverflow.com/a/23668035

Syer10:
Nope, I am on plain Windows 11, using Firefox, with no custom changes on my scroll wheel.

cpiber:
Sorry, I can't reproduce that, I tried it in a Win10 VM, and it behaved as expected. Can't even find any options in Windows regarding it. Can you please check `mousewheel.default.delta_multiplier_y` in about:config? Also if you put `document.body.addEventListener('wheel', e => console.log(e, e.deltaY))` (on any page), does it output positive numbers for down and negative for up, and how does that translate to the mouse for you? When I turn the wheel away from me (towards the front), I scroll up and get negative numbers.

Syer10:
image

Here is an example of me scrolling down(towards me), then scrolling up:
```
wheel { target: div#repo-content-pjax-container.repository-content
, buttons: 0, clientX: 303, clientY: 374, layerX: 303, layerY: 374 }
 102 [debugger eval code:1:54](chrome://devtools/content/webconsole/debugger%20eval%20code)
wheel { target: div#repo-content-pjax-container.repository-content
, buttons: 0, clientX: 305, clientY: 372, layerX: 305, layerY: 474 }
 -102
```

cpiber:
This is really stupid. We seem to be doing exactly the same thing... Only thing I can think of is that the host is Linux on my system... Should we just add a toggle to invert it?

Syer10:
That is probably the simplist solution for now. I'm currently debugging the special keys right now, have gotten enter fixed, now I just need to find why backspace isn't working.

cpiber:
Seeing Enter working is truly great :) By the way I just made some changes to allow mobile use. Do you happen to have an iPhone? I read that Safari apparently does not respect the meta key I'm using. We need to make sure that the Canvas is the visible space, even when the keyboard is open, but by default the browser will report the size under the keyboard (and on Safari even under the navigation bar if it's visible), on Firefox this works, same on Chrome

Syer10:
I do not have an iphone available to test on sadly.

I did find more issues with the keyboard so I am trying to find fixes for them, will let you know when I am done.

Syer10:
I cannot figure out how to get the backspace and arrow keys working. I simplified my enter fix so nothing that can't be typed gets typed. It seems that CEF doesn't follow standard conventions with the swing KeyEvent. I was able to get events from the Suwayomi-Launcher and tried to mimic them, but nothing worked.

I'm going to stop debugging that for now

cpiber:
> I cannot figure out how to get the backspace and arrow keys working. I simplified my enter fix so nothing that can't be typed gets typed. It seems that CEF doesn't follow standard conventions with the swing KeyEvent. I was able to get events from the Suwayomi-Launcher and tried to mimic them, but nothing worked.

Wait, really? On Linux they seem to just work, same on Mobile, in my Windows VM as well...

Syer10:
Weird, ya they don't work for me at all. No clue what is happening

cpiber:
So, there's definitely something else wrong with the cookies. Probably thanks to your changes, the cookies themselves are now much more stable. I'm logged into TappyToon even after a restart etc. However, this does not seem to translate to the built-in HTTP client, as I'm still getting "User does not have the license of requested content" even though I paid for the chapter in question. This even after a restart (Mihon also requires a restart of the app). I'm not seeing any caches in the file system, so I don't think it's that.

Any ideas where to look? I put some logging into the cookie jar, and it seems to return cookies, both for `www.tappytoon.com` and `tappytoon.com`...

Also, just because I noticed, in `NetworkHelper`:
```kotlin
                        Cache(
                            directory = File.createTempFile("tachidesk_network_cache", null),
                            maxSize = 5L * 1024 * 1024, // 5 MiB
                        ),
```
That's a temp **file**, not sure what that does on Windows, but on Linux that's an actual file, which is empty, because Okhttp expects a directory, is this intentional? :thinking:

cpiber:
> I put some logging into the cookie jar, and it seems to return cookies, both for www.tappytoon.com and tappytoon.com...

Nevermind, that was wrong, did not check closely enough. It does *not* see the cookies for `tappytoon.com`, there's just a few cookies stored under `www.tappytoon.com` that have `domain=tappytoon.com` set, so it *looks* like it's correctly loading them, but it isn't.

cookies for `tappytoon.com` **should** however be visible on `www.tappytoon.com`, so we need to fix that, I just need to figure out if we need to look at the domain or not. Also I think to be correct we should be preserving the leading dot, but maybe that's fine for now (leading dot appears to be like a wildcard, so all subdomains included, from what I can tell)

Syer10:
Oh, well feel free to fix the network cache 😂 

You could look at adding another network interceptor to see what cookies are being sent
```
                    .addNetworkInterceptor {
                        it.proceed(it.request().also { println(it.header("Cookie")) })
                    }
```

Syer10:
I'll look at fixing the cookies, it doesn't sound like a difficult change

cpiber:
Was just about to do that ^^

Have a look at this: https://github.com/franmontiel/PersistentCookieJar/blob/02711dace737017e91e74bcdecade858164676a6/okhttp3-persistent-cookiejar/src/main/java/com/franmontiel/persistentcookiejar/PersistentCookieJar.java#L59-L78 Maybe we should get rid of internal domains and just store everything as a list instead

okhttp can tell us if the cookie matches the domain

cpiber:
> I'll look at fixing the cookies, it doesn't sound like a difficult change

To avoid misunderstandings: Are you working on that, or did my message stop you?

Syer10:
I'm working on it, sadly OkHttp doesn't include a matches function on the Cookie class anymore, but I'm almost done anyway

Syer10:
@cpiber This should work better with situations like that, please test it when you can.

cpiber:
Yes, this is sick. I logged in on the WebView, then managed to open a locked chapter in Suwayomi (which I bought earlier).

Tried it a few times, even clearing the cookie store, seems to be quite stable. No restart required even, last I checked Mihon had to be restarted for the cookies to apply.

cpiber:
I realized I'd forgotten the discussed mouse direction toggle, so I quickly added that.
But aside from that, the PR is finished to me. Is there anything else you'd like to see? I think rendering improvements (if even necessary) can be a separate PR/discussion

Syer10 (approved):
Nope, we are good. So glad this is finished! This is perfect for the sources that need it.

yeezylife:
where's the webview entrance? using latest v2.1.1867

Syer10:
> where's the webview entrance? using latest v2.1.1867

http://localhost:4567/api/v1/webview

yeezylife:
> > where's the webview entrance? using latest v2.1.1867
> 
> http://localhost:4567/api/v1/webview

thanks. some how "http://127.0.0.1:4567/api/v1/Webview" doesn't work here. gotta use "http://localhost"

Issue #1349, opened 2025-04-19
WebUI Login Page to support Password Managers


soleuniverse101:
### Describe your suggested feature

To login to the web app, currently we need to fill out the credentials inside an alert box. This unfortunately isn't able to be autofilled by password managers (at least mine). This is particularly painful on mobile since I need to manually copy the credentials from my password manager and it closes the alert box each time.

It would be nice if here was a proper login page.

I run the latest version of the server (v1.1.1) and of the WebUI (r1689).

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have updated to the **[latest version](https://github.com/suwayomi/suwayomi-server/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.

cpiber:
Suwayomi currently does not have user management, so this cannot really be done. I assume you have set up basic authentication on your webserver in the reverse proxy, which would explain the dialog option. While basic authentication is the simplest, nginx supports several authentication schemes, e.g. one where you provide a login page: https://developer.okta.com/blog/2018/08/28/nginx-auth-request

Robonau:
> I assume you have set up basic authentication on your webserver in the reverse proxy.

no, basic auth is an option in Suwayomi itself, they more than likely are not using a reverse proxy

cpiber:
@soleuniverse101 Adding a login page wouldn't be too difficult, I think I'll work on that in a few days.

Until then: Did you know that you can use an URL of the form `https://user:password@domain.com/` to automatically transmit the login credentials? You could store that as a bookmark, or put it inside your password manager

Robonau:
> Until then: Did you know that you can use an URL of the form `https://user:password@domain.com/` to automatically transmit the login credentials? You could store that as a bookmark, or put it inside your password manager

iirc that format hasn't been supported in the major browsers for a long time (2017 for chromium)
https://stackoverflow.com/questions/45039335/bypass-blocking-of-subresource-requests-whose-urls-contain-embedded-credentials

for security reasons

cpiber:
That is for page links. It should work for the url bar. It works in both latest Firefox and Chromium on Linux

Robonau:
it doesn't work cause browsers don't auto set the auth header for requests when you use the url
[...]

PR #1498, opened 2025-07-05
[#1349] Basic Cookie Authentication

cpiber:
image

Basic login page and session management.
If enabled, any unauthorized request redirects to the login page, with a parameter that allows us to redirect back to the previous page on success.

Currently there is no session handler attached. Thus, all sessions are stored in memory and not persisted.

I'm opening this as a draft for two reasons:
- Injecting data into the page is not very clean. I'm currently just doing a string replace, and I want to know if you're fine with that. There seems to be support for templating, but that seems kind of overkill for this small page, but I'll change it if wanted (note that this needs an additional dependency)
- In #1349, @Syer10 mentioned having an enum for the authentication methods. I have currently implemented it as two booleans, as I'm not sure how to properly migrate the config file for users that have already enabled basic authentication. Same goes for backups.


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/backup/proto/models/BackupServerSettings.kt
```diff
@@ -48,6 +48,7 @@ data class BackupServerSettings(
     @ProtoNumber(29) override var basicAuthEnabled: Boolean,
     @ProtoNumber(30) override var basicAuthUsername: String,
     @ProtoNumber(31) override var basicAuthPassword: String,
+    @ProtoNumber(56) override var cookieAuthEnabled: Boolean,
```
cpiber: Please also check that this is fine. I'm assuming that the number here is important for deserialization of backups, and that renumbering would cause old backups to fail to deserialize properly, is that right?

Syer10:
KeePassXC! Love to see it

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/backup/proto/models/BackupServerSettings.kt
Syer10: Correct, whenever a new setting is added, it needs to use the next number available.

server/src/main/kotlin/suwayomi/tachidesk/server/JavalinSetup.kt
```diff
@@ -111,8 +117,41 @@ object JavalinSetup {
...
+                val redirect = ctx.queryParam("redirect") ?: "/"
+                // NOTE: We currently have no session handler attached.
+                // Thus, all sessions are stored in memory and not persisted.
+                // Furthermore, default session timeout appears to be 30m
+                ctx.header("Location", redirect)
+                ctx.sessionAttribute("logged-in", username)
+                throw RedirectResponse(HttpStatus.SEE_OTHER)
+            }
+
+            var page = String(this::class.java.getResourceAsStream("/static/login.html")!!.readAllBytes(), StandardCharsets.UTF_8)
```
Syer10:
```suggestion
            var page = this::class.java.getResourceAsStream("/static/login.html")!!.use { it.readAllBytes() }.toString(Charsets.UTF_8)
```

server/src/main/kotlin/suwayomi/tachidesk/server/JavalinSetup.kt
```diff
@@ -111,8 +117,41 @@ object JavalinSetup {
                 }
             }
 
+        app.get("/login.html") { ctx ->
+            var page = String(this::class.java.getResourceAsStream("/static/login.html")!!.readAllBytes(), StandardCharsets.UTF_8)
```
Syer10:
```suggestion
            var page = this::class.java.getResourceAsStream("/static/login.html")!!.use { it.readAllBytes() }.toString(Charsets.UTF_8)
```

Syer10 (commented):
I am fine with the String replacements as thats probably the best way unless we go full templating library, which isn't worth the work for just this.

What I would suggest for the basic auth options, is to deprecate them, but still support them. We don't remove them, but they are only checked if we see `NONE` being used for the new server.authMode option.
For the new options, we add new username/password setting and server.authMode enum(as mentioned previously). This username/password combo is what is used primarily, and in the logs if we detect the old basic auth settings being used, we spit out a warning.

```kotlin
AuthMode {
    NONE,
    BASIC_AUTH,
    SIMPLE_LOGIN,
    // ACCOUNT for when user accounts are added
}
```

cpiber:
I managed to add a migration path to translate the original setting to the new authMode enum. This also works for updates from WebUI, so no incompatibility there. I've tried testing the backup export, but I have not managed to get it to restore *any* of my settings, so I can't tell how those two interact in the backup. Ideally, new backups wouldn't include basicAuthEnabled at all, but I have not found a way to then still deserialize it

Syer10:
Looks good! Can you do the migration with the username/password as well? authUsername and authPassword for naming would work

cpiber:
That was actually a bit easier, since migrating simply means adopting the value under a new name, so I've changed the backup serialization to just read the same ProtoNumber into the new value

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/backup/proto/models/BackupServerSettings.kt
```diff
@@ -45,9 +46,13 @@ data class BackupServerSettings(
...
     // Authentication
+    @ProtoNumber(56) override var authMode: AuthMode,
     @ProtoNumber(29) override var basicAuthEnabled: Boolean,
-    @ProtoNumber(30) override var basicAuthUsername: String,
-    @ProtoNumber(31) override var basicAuthPassword: String,
+    @ProtoNumber(30) override var authUsername: String,
+    @ProtoNumber(31) override var authPassword: String,
+    // deprecated, not serialized
+    override var basicAuthUsername: String,
+    override var basicAuthPassword: String,
```
Syer10: Even without a number, they will be serialized, probably to the next available number, give them a really high number that we will likely never hit, also make them nullable if you can.

Syer10 (changes requested):

Syer10 (approved):

Syer10:
I will merge this once the launcher and docker container are ready

Syer10:
I need to do some testing with the docker container still, I am going to merge PRs and cancel the docker build, then release the docker build when I am done.

Issue #1497, opened 2025-07-05
Cubari: remote storage account support in extension via webview login not working


AwkwardPeak7:
### Steps to reproduce

1. Get latest preview server with webview implementations
2. Get cubari extension
3. login to remote storage account (5apps or other) on cubari via webview
4. have something in you history or pinned series on your cubari remote storage account
5. open popular/latest section of extension

### Expected behavior

6. shows history in latest section or pinned series in popular section

### Actual behavior

6. error
```
ApolloError: Cannot read Json element because of unexpected end of the input at path: $
JSON input: 
```
[server.log](https://github.com/user-attachments/files/21078058/server.1.log)

### Suwayomi-Server version

v2.0.1826 (via docker)

### Server operating system

Arch Linux (host)

### Server Desktop Environment

Kde Plasma (host)

### Server JVM version

the one used in the docker container

### Used client name

WebUI

### Client version

v20250703.01 (r2643)

### Used web browser

Firefox 140

### Client operating system

Arch Linux

### Other details

started the server and got logs like so:
```sh
$ docker-compose pull
$ xhost +
$ docker-compose up | tee server.1.log
```

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have tried the troubleshooting guide described in [README.md](https://github.com/Suwayomi/Suwayomi-Server?tab=readme-ov-file#troubleshooting-and-support)
- [x] I have updated to the **[latest version](https://github.com/suwayomi/suwayomi-server/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.
- [x] I understand that **Suwayomi does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

Syer10:
@cpiber It seems like its something to do with the background webview as Cubari uses it extensivly https://github.com/keiyoushi/extensions-source/blob/main/src/all/cubari/src/eu/kanade/tachiyomi/extension/all/cubari/RemoteStorageUtils.kt

cpiber:
The extension does not really handle running into a timeout. I don't see anything interesting in your logs - the page loads normally, but no activity happens.

Expected logs:
```
19:27:53.126 [Thread-0] INFO android.util.Log -- [VERBOSE] KcefWebViewProvider: KcefWebViewProvider: initialize
[...]

PR #1500, opened 2025-07-06
[#1497] WebView: Localstorage

cpiber:
Enables local storage for WebView.

This currently has some problems though: For some reason, the background WebView only sees a part of the storage: `{"storageOnce":"1","proxyHistory":"true"}`. The interactive webview always gets access to the full one, even after multiple reloads and a restart of the server. `window.location.origin` match. Not sure what else to check.


cpiber:
Nevermind, found the problem, which is a problem in itself: Cubari uses `loadUrl` with headers, which we do by creating a dedicated context. This already caused problems with Cookies before (where we couldn't read them), and apparently it does the same with localStorage. We'll  need to figure out how else to handle headers and post data, or if it's just `KCEFResourceRequestHandler`

Syer10 (approved):

Issue #534, opened 2023-04-05
[Meta] Tachidesk-Server 1.x


AriaMoradi:
There are some early design decisions in the app that we can't change or is too complex or too substantial/fundamental to change with smaller incremental patches; Having these changes would call for a "rework" of the app.

Also as a desktop and (self-hosted) server application, Tachidesk is not just merely a remake of Tachiyomi anymore, it's a different solution to different problems and needs, so we should create a new vison as the only common ground we really have with Tachiyomi is support for its Extensions and backups. 

The changes are up for debate.

## Changes in the organization
- Drop the name Tachidesk, rebrand as Suwayomi and Suwayomi-Xxx completely 

## Changes for Tachidesk-Server 1.x
- Changes in API system
    - Move from REST API to GrapghQL, remove Javalin (Supported by @martinek and @Syer10)
    - Support both REST and GraphQL?
    - Support for proper Authorization and Authentication
    - Better Error and request result reporting (+ l10n and i18n)
- Database and data model changes
    - Support for multi-user
    - Hot-Changeable settings, move away from `com.typesafe.config`
    - Change ORM: Expoed has worked for us so far but if we are going do redo the database structure why not? 
    - Allow the user to use PostgreSQL instead of H2 for better performance
- Support for more Komga-like features
    - Better Local Manga Source (or a separate system)
    - Allow for loading and saving manga 
    - Change the terminology: Book instead of Manga or Entry
- Misc.
   - Rename the base package from `suwayomi.tachidesk` to `org.suwayomi`
    
## Changes for Tachidesk-WebUI
- Desktop client via ElectronJS or Tauri
    -  Use desktop features: Notifications
    - Cloudflare bypass solution
- WebView based Android client (ref: [super-productivity-android](https://github.com/johannesjo/super-productivity-android))
    -  Notification support
    - Cloudflare bypass solution
    - Image caching
- iOS client
    - ???

Syer10:
> Move from REST API to GraphQL

GraphQL is a powerful way query a website, it allows many different queries, with support for filtering, paging, and a bunch of other things. GraphQL allows you to build database-like queries with all the data at your fingertips, performantly.

> remove Javalin

50/50 for me, Javalin supports everything we need on the graphql side, and it would support JWT authorization, but so would Ktor which is what I would prefer.

> Support both REST and GraphQL?

I think it might be good to have both for a while, maybe permanently move to GraphQL once its battle tested.

[...]

PR #1505, opened 2025-07-07
[#1496] Image conversion

### LINKED ISSUE #1496: Converting images on download

Many sources use modern image formats like webp to achieve great compression. However, older devices may not support these formats. So it would be nice to be able to configure a conversion step to jpeg or png, which are widely supported.

This could be either on every download, or a bit more sophisticated to enable for just particular formats (e.g. convert webp, but not jpeg).

### Other details

This should work before CBZ archiving.

### Acknowledgements

cpiber:
- Optional conversion on download
  - This conversion is only done on actual download (via queue), not on cache
  - Configurable per image mime type, with a possible default conversion
- Query parameter to convert in-memory on request
  - If page already downloaded (or local), converts that in-memory
  - Otherwise downloads the page without conversion, then converts in-memory

The main reason for this design decision is that `Page` is responsible both for interactive requests and downloads, so supporting both download conversion and user-provided conversion is more complex. It also makes reading without downloads slower, as conversion takes noticeable time.

The current configuration for download conversions is done via an object. The config library converts this to `server.downloadConversions.default = "..."` and `server.downloadConversions["image/webp"] = "..."`, hence the special case in `ConfigManager`. Maybe we can find a better format, which is better supported?

Closes 1496


Syer10:
~I agree that the config format is getting in out way a bit. I would be open to alternatives.~

~I have had good experiences with TOML in the past, and there is a support TOML library for Kotlin here https://github.com/orchestr7/ktoml,  though I'm not sure how comments would work with that or other formats. We have also mentioned this before in this issue https://github.com/Suwayomi/Suwayomi-Server/issues/534~


~Do you want to take a look at this? I can look at this in my own time if not. Let me know what your plan is, we can move forward with this PR as it is for now.~

Config format is working good now

cpiber:
> I agree that the config format is getting in out way a bit. I would be open to alternatives.
> 
> I have had good experiences with TOML in the past, and there is a support TOML library for Kotlin here [orchestr7/ktoml](https://github.com/orchestr7/ktoml?rgh-link-date=2025-07-07T21%3A22%3A11Z), though I'm not sure how comments would work with that or other formats.

I was more referring to the format of the conversion mapping. I hadn't thought of completely rewriting the config system.

> Do you want to take a look at this? I can look at this in my own time if not. Let me know what your plan is, we can move forward with this PR as it is for now.

For now, I don't think I want to work on migrating/rewriting the config system. That might be interesting for the future, but at the moment I don't have that much time and my vacation is coming up, where I want to reduce contributions

Syer10:
> I was more referring to the format of the conversion mapping. I hadn't thought of completely rewriting the config system

> For now, I don't think I want to work on migrating/rewriting the config system. That might be interesting for the future, but at the moment I don't have that much time and my vacation is coming up, where I want to reduce contributions

Got it, no pressure. 😄

I'll do a in depth review later, I also need to figure out how to handle this in the launcher and docker container before merging.

cpiber:
We could also simplify what can be configured.

I currently have it set up as a map (object) from input mime type to output mime type, but I think realistically users will want to convert their entire library to some format, or not at all. Maybe we could have a whitelist of formats to be converted (e.g. convert "image/jpeg" and "image/png" to "image/webp"), that should cover most usecases

Syer10:
I've found a good way to rewrite our Config usage to be much simplier, it will be easier to use once I am done and we can keep the configuration the way it is, and even improve it.

Syer10:
Now that I have finished my config file improvements, I am fine with keeping the nested configuration. Would it be possible to add options? Like say the compression level. For example something like
```hocon
server.downloadConversions["image/webp"] = {
    target = "image/jpg"
    compression = 90 # I have no clue what you would actually need to put here
}
```

It should be easy to change the map value to a data class for the options.

cpiber:
I had actually planned options originally, don't know why I forgot about it, will do!

cpiber:
[ImageWriteParam](https://docs.oracle.com/javase/8/docs/api/javax/imageio/ImageWriteParam.html) only supports quality and tiling. Tiling does not seem useful (~~and I believe is only sensible for JPEG~~ was thinking about progressive mode, but tiling still does not seem useful), so quality is the only setting for now.

I've also added an example to the reference config, and updated config manager to copy that over. With `config.entrySet()`, only non-null entries are listed (recursively). But an empty object (the default for conversions) is essentially null, so it was never detected. Instead manually iterate the first two levels (the expected depth) and use that as basis, since that also includes empty objects.

Note that this format would also work, since any JSON is allowed:

```
server.downloadConversions = {
  "image/webp": {
    target = "image/jpeg"
  }
}
```

Syer10 (approved):
Currently working on the launcher and then will work on the docker container.

Issue #489, opened 2023-02-06
[Feature Request] Flatpak/snap/Appimage support


dnet890:
---

## What feature should be added to Tachidesk?
deb support is fine. But, it will be wonderful to cover multiple distributions. So, Flatpak/snap/appimage will be great

##  Why/Project's Benefit/Existing Problem
Since, SteamOS growing exponentially. it will be wonderful if we can get this application on different packaging to cover multiple distributions.

djgerman10:
I can't install it, can you explain how to do it, it gives me these errors
![Captura de pantalla 2023-07-28 20:17:14](https://github.com/Suwayomi/Tachidesk-Server/assets/48584512/35a4ae04-f759-433f-8c6a-d260d2ed73d9)

Syer10:
AppImage support has been added in the latest preview

PR #1519, opened 2025-07-12
Add AppImage bundle

KamaleiZestri:
Related to #489.

Add AppImage as a bundle option.

### DIFF
--- a/.github/workflows/build_push.yml
+++ b/.github/workflows/build_push.yml
@@ -112,16 +112,25 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


Robonau:
if its a build option it should probably also be added to the workflows (here and the preview repo) no?

KamaleiZestri:
> if its a build option it should probably also be added to the workflows (here and the preview repo) no?

You are correct, I misspoke. Corrected it to "bundle" option instead of "build" option.

Syer10:
I'll add the workflows to the PR and then merge it

PR #1522, opened 2025-07-15
Localize WebView and Login pages

cpiber:
I'm not 100% happy with this, since it requires the user to distinguish between translated strings in HTML and JS/CSS Strings, but without parsing the HTML I don't think we're going to get anything better (JTE?)

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/global/controller/WebViewController.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/global/controller/WebViewController.kt
@@ -12,8 +12,6 @@ import io.javalin.http.HttpStatus
 import io.javalin.websocket.WsConfig
 import suwayomi.tachidesk.global.impl.WebView
 import suwayomi.tachidesk.i18n.LocalizationHelper
-import suwayomi.tachidesk.i18n.MR
-import suwayomi.tachidesk.server.JavalinSetup
 import suwayomi.tachidesk.server.util.handler
 import suwayomi.tachidesk.server.util.queryParam
 import suwayomi.tachidesk.server.util.withOperation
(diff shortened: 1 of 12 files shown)


Syer10:
JTE seems like a good library, especially since you can write Kotlin HTML templates. At this point it may be worth the implementation.

cpiber:
For simplicity, I'm always using generated rendering for now, we could set up hot-reload on `:server:run` (default, but doesn't work for shadow jar), but I don't think we'll be developing many pages, so this should be fine

Also I saw only later on that there's a kotlin extension for `.kte` (needs additional libraries), maybe we want to switch to that?

Syer10:
Can you switch the jte files to kte files? It would be better to use the Kotlin templating instead.

Syer10:
Lol just saw your edit. Yes it would be better to use the Kotlin version

server/src/main/kotlin/suwayomi/tachidesk/global/controller/WebViewController.kt
```diff
@@ -11,21 +11,31 @@ import io.javalin.http.ContentType
...
                     description("Opens and browses WebView")
                 }
             },
-            behaviorOf = { ctx ->
+            behaviorOf = { ctx, lang ->
+                val locale: Locale = LocalizationHelper.ctxToLocale(ctx, lang)
                 ctx.contentType(ContentType.TEXT_HTML)
-                ctx.result(javaClass.getResourceAsStream("/webview.html")!!)
+                ctx.render(
+                    "Webview.jte",
```
Syer10, line 34: Are the filenames here still good to be `.jte`? Or do they need to be switched to `.kte`

cpiber: I tested it and seems to be fine. Docs also don't say anything about different extensions.

Switching to kte extension fails with:

```
You're missing the 'jte-kotlin' dependency in your project. Add the dependency:

pom.xml:
<dependency>
    <groupId>gg.jte</groupId>
    <artifactId>jte-kotlin</artifactId>
    <version>2.2.1</version>
</dependency>

build.gradle or build.gradle.kts:
implementation("gg.jte:jte-kotlin:2.2.1")
```

So I assume now it's trying to compile the template at runtime, instead of using the generated class

Syer10 (approved):

PR #1524, opened 2025-07-15
Basic JWT implementation

cpiber:
As discussed on discord

Instead of the transparent server-side login, switch to a JWT implementation. This requires cooperation from the UI.

Very heavily based on #623


server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    @OptIn(ExperimentalEncodingApi::class)
+    fun generateSecret(): String {
+        val keyBytes = ByteArray(32)
+        SecureRandom().nextBytes(keyBytes)
+
+        val secretKey = SecretKeySpec(keyBytes, ALGORITHM)
+
+        return Base64.encode(secretKey.encoded)
+    }
```
Syer10: This key needs to be generated only once, and then save it to be reused. We can save it into a SharedPreferences like we do in TrackerPreferences

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlin.time.Duration.Companion.hours
+
+object Jwt {
+    private val logger = KotlinLogging.logger {}
+
+    private const val ALGORITHM = "HmacSHA256"
+    private val accessTokenExpiry = 1.hours
+    private val refreshTokenExpiry = 60.days
+    private const val ISSUER = "tachidesk"
+    private const val AUDIENCE = "" // todo audience
```
Syer10: This TODO should probably be investigated(I had not gotten to it yet)

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlin.time.Duration.Companion.days
+import kotlin.time.Duration.Companion.hours
+
+object Jwt {
+    private val logger = KotlinLogging.logger {}
+
+    private const val ALGORITHM = "HmacSHA256"
+    private val accessTokenExpiry = 1.hours
+    private val refreshTokenExpiry = 60.days
+    private const val ISSUER = "tachidesk"
```
Syer10:
```suggestion
    private const val ISSUER = "suwayomi-server"
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlin.io.encoding.ExperimentalEncodingApi
+import kotlin.time.Duration.Companion.days
+import kotlin.time.Duration.Companion.hours
+
+object Jwt {
+    private val logger = KotlinLogging.logger {}
+
+    private const val ALGORITHM = "HmacSHA256"
+    private val accessTokenExpiry = 1.hours
+    private val refreshTokenExpiry = 60.days
```
Syer10: Maybe make it user-configurable?

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
```diff
@@ -0,0 +1,97 @@
...
+            refreshToken = refreshToken,
+        )
+    }
+
+    fun refreshJwt(refreshToken: String): JwtTokens {
+        val jwt = verifier.verify(refreshToken)
+        require(jwt.getClaim("token_type").asString() == "refresh") {
+            "Cannot use access token to refresh"
+        }
+        return generateJwt()
```
Syer10:
```suggestion
    fun refreshJwt(refreshToken: String): String {
        val jwt = verifier.verify(refreshToken)
        require(jwt.getClaim("token_type").asString() == "refresh") {
            "Cannot use access token to refresh"
        }
        return createAccessToken()
```
Lets only refresh the access token, I want the refresh token to only be given once per login.
You will need to change the refreshToken mutation as well

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
cpiber: From a quick search, this should usually be the public url of the API, but this is not really known here. We could make it a configurable setting? Another option is also just some generic string like "tenant-xyz" or "service-A", so the setting could really be any string, maybe defaulting to "suwayomi-server".

https://mojoauth.com/blog/lets-understand-jwt-audience-aud-claim/

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
cpiber: Makes sense, I think that's the typical usecase

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
Syer10: That works for me

cpiber:
I'm not sure if how I'm parsing the duration is acceptable for you, seemed easier to do it like this than hack it somewhere into `ConfigModule`

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -69,6 +70,33 @@ class ServerConfig(
...
+                .drop(1)
+                .distinctUntilChanged()
+                .filter { it != parseValue(thisRef.overridableConfig.getConfig().getValue<ServerConfig, String>(thisRef, property)) }
+                .onEach { GlobalConfigManager.updateValue("$moduleName.${property.name}", (it as T).toString() as Any) }
+                .launchIn(mutableConfigValueScope)
+
+            return stateFlow
+        }
+    }
+
```
Syer10: You don't actually need this, you can register a custom Config type like I have done with MutableStateFlowType(search it and its usages), just simpler since its a basic type.

server/src/main/kotlin/suwayomi/tachidesk/global/impl/util/Jwt.kt
```diff
@@ -0,0 +1,126 @@
...
+                preferenceStore.edit().putString(PREF_KEY, Base64.Default.encode(k)).apply()
+                k
+            }
+
+        val secretKey = SecretKeySpec(keyBytes, ALGORITHM)
+
+        return Base64.encode(secretKey.encoded)
+    }
+
+    private val algorithm: Algorithm = Algorithm.HMAC256(generateSecret()) // todo store secret
```
Syer10:
```suggestion
    private val algorithm: Algorithm = Algorithm.HMAC256(generateSecret())
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
cpiber: But we still want to have a MutableStateFlow, right? So it would need to match a `MutableStateFlow<Duration>`. I'll look into that tomorrow

Syer10: Nope, it just needs to match Duration, the way I have MutableStateFlow is just a wrapper for the internal type deserialization. So you only need to register a Duration custom type

cpiber:
I also noticed that GraphQL in the entirety and the WebSockets weren't protected.

There's a minor problem with the WebView though: Currently no UI supports it, but at some point there will probably be a button to open it in a new tab. Then, the token needs to somehow be transferred to allow opening the page. I currently assume that it is transmitted in the `token` query parameter, but that [seems to be rather unideal for security](https://portswigger.net/kb/issues/00500700_session-token-in-url). So is setting a JS-readable cookie imo.

Something that should work is creating a standard form and having that [open in a new tab](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form#target), thus opening WebView as a post request. Not ideal, but would avoid putting the token in the browser history...

Syer10:
Great, I was going to check if you had copied the gql protection today. 

We could shorten the default access token expiry time to something like 5 mins, that would minimize the risk.

We could also put the refresh token in the cookies and have WebUI and other pages share it.

cpiber:
> We could shorten the default access token expiry time to something like 5 mins, that would minimize the risk.

True, but that also means that the WebUI has to refresh quite often. But I think 5 minutes is quite common (some recommendations I see are 15mins).

> We could also put the refresh token in the cookies and have WebUI and other pages share it.

Actually, I don't think that's needed. The token is only checken on connect, so for pages like WebView which only use sockets, they only need the token once, briefly. This could be considered a security problem in itself though. I'm not too familiar with GraphQL, can you send quries/mutations over the socket, or just subscribe? If it's just a subscription, we just need to be careful to not expose any actual configuration over the sockets.

Otherwise we'd need to implement some protocol where specific messages may be sent to authenticate (i.e. inside the connection, not when establishing).

Also, a JS-accessible cookie is barely better than passing it via URL to my understanding, so I'd avoid that.

Syer10:
The issue with those solutions is that /api/v1/weview will be inaccessable on its own. I would like people to be able to just open api/v1/webview and not run into any accessability issues as long as they have a valid token.

Syer10:
> The token is only checken on connect, so for pages like WebView which only use sockets, they only need the token once, briefly. This could be considered a security problem in itself though. I'm not too familiar with GraphQL, can you send quries/mutations over the socket, or just subscribe? If it's just a subscription, we just need to be careful to not expose any actual configuration over the sockets.

I think you can just subscribe? I've never actually seen anything else done. But GQL is very complicated so I could be wrong.

> True, but that also means that the WebUI has to refresh quite often. But I think 5 minutes is quite common (some recommendations I see are 15mins).

Usually these things use an interceptor type system so if one request fails it refreshes the token and tries again. It shouldn't impact the user at all.

cpiber:
> The issue with those solutions is that /api/v1/weview will be inaccessable on its own. I would like people to be able to just open api/v1/webview and not run into any accessability issues as long as they have a valid token.

I guess that makes sense. So you'd have the user login over WebUI, then independently open `/api/v1/webview`, and this should just work? That only leaves cookies and localstorage then.

To support the use-case where WebUI/... is hosted on a separate port from the server, as brought up in discord, I think cookies are the only viable solution, since those can be set for specific domains (the UI would need to set it specifically to the server's domain, not sure if that works with ports).

I'll test this for a bit.

> Usually these things use an interceptor type system so if one request fails it refreshes the token and tries again. It shouldn't impact the user at all.

Yes, but it's still requests that need to be sent, and a (perhaps) noticeable delay every time the token needs to be refreshed before retrying the request.

cpiber:
So, basically: If it's just a different port, no problem. The cookie is per-domain, so it applies to the server automatically. The problem is if it's actually a different domain. The browser will reject this, as we may only set cookies for ourselves (to avoid tracking etc). It seems the server could allow this by setting a CSP policy, but I'm honestly not sure where we'd even put that, since we're not actually making a request... The server could set it for itself, if you can figure out how to do that from inside GraphQL, that might work

cpiber:
Also, at least with how it is set up in Suwayomi, the GQL socket only accepts subscription messages, via `ApolloSubscriptionProtocolHandler`

Robonau:
not sure if this is useful

the clients can remake the `/api/v1/webview` UI (im in the middle of doing that in VUI)
but the websocket would need a way to authenticate that isn't an authentication header since browsers cant set that on websockets

same issue for the GQL subscriptions authentication

i think we could just set a cookie on the client to the token
and the server reads that

cpiber:
>  but the websocket would need a way to authenticate that isn't an authentication header since browsers cant set that on websockets

The page had that same problem ;) As it is currently, you can set the subprotocol (second argument of the constructor), see https://github.com/cpiber/Suwayomi-Server/blob/login-api/server/src/main/jte/Webview.kte

> i think we could just set a cookie on the client to the token and the server reads that

I have implemented that as well, with the cookie "suwayomi-server-token"

Syer10:
`suwayomi-server-token` would be a access token then? Does that mean we need to refresh the refresh token in the webview HTML?

The only solution I can think of for the mentioned issue is implementing a simple login dialog in the webview HTML.

Robonau:
> The only solution I can think of for the mentioned issue is implementing a simple login dialog in the webview HTML.

the clients can just deal with the webview UI

cpiber:
> suwayomi-server-token would be a access token then? Does that mean we need to refresh the refresh token in the webview HTML?

Only the access token, yes. But once the connection is established, we don't need the token anymore (everything over the socket). Would of course require that the user opens relatively soon after doing something in WebUI. If that is not acceptable, then yes, or just let the UI deal with it, as @Robonau suggests



> > but the websocket would need a way to authenticate that isn't an authentication header since browsers cant set that on websockets
> 
> The page had that same problem ;) As it is currently, you can set the subprotocol (second argument of the constructor), see [cpiber/Suwayomi-Server@`login-api`/server/src/main/jte/Webview.kte](https://github.com/cpiber/Suwayomi-Server/blob/login-api/server/src/main/jte/Webview.kte?rgh-link-date=2025-07-17T17%3A54%3A45Z)
> 
> > i think we could just set a cookie on the client to the token and the server reads that
> 
> I have implemented that as well, with the cookie "suwayomi-server-token"

My bad, that was not quite true. It seems like cookies aren't sent for WebSocket requests, so the *only* way is to pass it via the protocol or the URL (as a `token` query parameter).

Robonau:
> My bad, that was not quite true. It seems like cookies aren't sent for WebSocket requests, so the _only_ way is to pass it via the protocol or the URL (as a `token` query parameter).

in browser dev tools it sure seems like cookies are sent just fine for websockets
image

cpiber:
> in browser dev tools it sure seems like cookies are sent just fine for websockets

Sorry, should have checked that better. I used the cookie as a fallback to the protocol header, but not setting that still gives a value, not `null`, so the cookie wasn't actually used on the server side. Pushed a working check now

Syer10:
I am feature freezing Suwayomi-Server for the next week so that we have time to squash more bugs before the release I hope to get out at the end of July.

I would like to get this tested before its merged, could we get a test implementation in WebUI to verify its working how we want it @cpiber @schroda ?

Also this is still marked as draft.

cpiber:
Makes sense.

I kept it as a draft because I felt not everything was fully discussed, and as you said, there's no UI implemention. I'm now fully on vacation though, so I probably won't get much done (another reason why I did not feel comfortable marking it as ready).

schroda:
I can test it on the weekend or so

cpiber:
Since we now have a working implementation in WebUI (not yet merged), I think the design of this is sound, so I'm marking it as ready for review.

Only thing left to finalize is WebView. I see two options:

1. Remove WebView HTML endpoint and require UIs to implement the functionality like with other features. The endpoint was never in a stable release, so no need to further support it imo (or leave it as a POC, but not officially supported).
2. Still allow the UI to pass a token (via post or get parameter, or cookie?), but if no token is provided, display a simple login page. I think that should be fairly doable, even if it makes the page a bit more complicated.

Of course there's also the possibility of mixing the approach, even if we go with 2 the UIs can still re-implement the functionality in a page that better fits into the UI's design.

cpiber:
#1560 should probably be merged before this, I'll adjust their APIs to use authorization as introduced here

Also, I'm planning allowing basic auth as fallback for OPDS as discussed on discord as a separate PR after this is done

Syer10 (approved):

Syer10:
Once all the new changes are implemented and I setup the launcher/docker configuration, I'll merge it

cpiber:
That should be an up-to-date master, including new endpoints from #1560.

However, we still haven't fully resolved the question with Webview. Since the last release, WebUI displays buttons which open the Webview in a new tab, so this should probably work conveniently (or we get the UIs to embed or reimplement).

Since the login is only needed briefly to start the connection, I think a cookie or query param with the auth token (no refresh token!) would be fine, and if not supplied we can still implement a tiny login prompt in Webview (without refresh it's quite simple, we can just hardcode the GQL query).

Syer10:
> Since the login is only needed briefly to start the connection, I think a cookie or query param with the auth token (no refresh token!) would be fine, and if not supplied we can still implement a tiny login prompt in Webview (without refresh it's quite simple, we can just hardcode the GQL query).

This seems like the best solution for this issue

cpiber:
I'll see about implementing a login. The cookie and paramter should both already work, but I'll check again a bit later

cpiber:
New commit adds a very basic login dialog. Like WebUI, it attempts to connect with given parameters (includes cookie), and if that fails, it prompts the user to login. It also sets the cookie itself, so the next refresh will not prompt again, until the token expires.

The code for handling that is a bit ugly, with some nested callbacks, I can clean that up a bit if you want, but since it's only used once I left it like that...

I also noticed that the dialogs don't really match the design of the rest of the page, which tries to match default WebUI (dark mode?). This also applies to the copy dialog from #1593, so we can adjust that in a new PR? or leave it as it is

Syer10 (commented):
Was the basic auth alternative for the OPDS endpoints implememented? I don't see it.

cpiber:
> Also, I'm planning allowing basic auth as fallback for OPDS as discussed on discord as a separate PR after this is done

I was planning a separate PR, but I can implemented it in this if you want (but tomorrow)

Syer10:
Separate PR works 👍

Syer10 (approved):

PR #1526, opened 2025-07-16
Fix downloads on errors when converting image

Syer10:


Syer10:
@cpiber

cpiber (approved):
Looks good to me, is there any particular situation where the previous code caused problems? Doesn't seem like it changes much in the logic

Edit: Ah no, I see, the try was only for creating the output stream, not the conversion itself, got it, makes sense

Syer10:
Yes, in the discord(suwayomi-server channel) someone reported a exception with writer.write that errored the download

PR #1542, opened 2025-07-24
page load fix patch

Chiru-Dey:
in my recent pr, some pages didnot load. this patch fixes it


Syer10:
This fix is too complicated and not worth the code that was added.

We will implement an alternate fix later.

Chiru-Dey:
ok thank you

Chiru-Dey:
Tip: the images don't load because url has json metadata appended to it, making the url too long for database

PR #1560, opened 2025-08-01
feat(sync/koreader): Add KOReader reading progress synchronization

zeedif:
This pull request introduces a comprehensive integration with KOReader Sync Server, enabling two-way synchronization of reading progress between Suwayomi-Server and KOReader devices. This allows users to maintain a consistent reading experience across platforms.

#### Key Features

*   **Two-Way Synchronization:** Reading progress (last page read and read status) is pushed to the KOReader Sync Server when updated in Suwayomi and pulled when a chapter is accessed, ensuring the latest progress is always available.
*   **Configurable Sync Strategy:** Users can define how conflicts are handled via the `koreaderSyncStrategy` setting, with options to prioritize the latest timestamp (`LATEST`), always trust the KOReader server (`KOSYNC`), or always trust Suwayomi-Server (`SUWAYOMI`).
*   **GraphQL API for Sync Management:** A new set of GraphQL queries and mutations has been added to manage the KOReader Sync connection seamlessly:
    *   `koSyncStatus`: Checks the current login status.
    *   `connectKoSyncAccount`: A simplified flow that attempts to log in and, if the user doesn't exist, automatically registers them on the sync server.
    *   `logoutKoSyncAccount`: Clears local credentials to disconnect from the service.
*   **OPDS Integration:** Chapter metadata feeds now automatically pull the latest progress from the KOReader Sync Server, providing up-to-date reading status to OPDS clients.
*   **Settings and Backup:** All new KOReader Sync settings are fully integrated into the server configuration, GraphQL settings API, and the backup/restore system.

#### Technical Implementation

*   A new `KoreaderSyncService` has been implemented to encapsulate all logic related to authentication, hashing, and communication with the sync server.
*   The `Chapter` table has been extended with a `koreader_hash` column to uniquely identify chapter files (CBZs) on the sync server. A database migration (`M0051`) is included.
*   CBZ files are now hashed on download completion using KOReader's specific algorithm to ensure document compatibility.
*   Progress updates via the API (`ChapterMutation`) and OPDS (`OpdsEntryBuilder`) now trigger sync operations.
*   Configuration options for the feature have been refactored to use enums (`KoreaderSyncStrategy`, `KoreaderSyncChecksumMethod`) for improved type safety and robustness.

#### For Frontend (WebUI) Developers

To ensure the reader opens on the correct, synchronized page, a small adjustment is required in the WebUI's GraphQL query for fetching chapter pages. The `fetchChapterPages` mutation now returns the updated progress fields on the `chapter` object after performing a sync check.

**Before:**

```graphql
mutation GET_CHAPTER_PAGES_FETCH($input: FetchChapterPagesInput!) {
  fetchChapterPages(input: $input) {
    chapter {
      id
      pageCount
      __typename
    }
    pages
    __typename
  }
}
```

**After:**

The query should be updated to request `lastPageRead` and other relevant progress fields.

```graphql
mutation GET_CHAPTER_PAGES_FETCH($input: FetchChapterPagesInput!) {
  fetchChapterPages(input: $input) {
    chapter {
      id
      pageCount
      lastPageRead  # <-- Added: The synced last page read
      lastReadAt    # <-- Recommended: The timestamp of the last read
      isRead        # <-- Recommended: The updated read status
      __typename
    }
    pages
    __typename
  }
}
```

By making this change, the WebUI will automatically receive the most up-to-date reading progress when a user opens a chapter, allowing the reader to jump to the correct page seamlessly.


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/sync/KoreaderSyncService.kt
```diff
@@ -0,0 +1,359 @@
...
+    ): AuthResult {
+        val request =
+            buildRequest("${serverConfig.koreaderSyncServerUrl.value.removeSuffix("/")}/users/auth") {
+                get()
+                addHeader("x-auth-user", username)
+                addHeader("x-auth-key", userkey)
+            }
+
+        return try {
+            network.client.newCall(request).execute().use { response ->
```
Syer10: This needs to use .await() instead of .execute(), make sure to change each instance

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/sync/KoreaderSyncService.kt
```diff
@@ -0,0 +1,359 @@
...
+                KoSyncConnectPayload(false, registerResult.message ?: "Registration failed.", null)
+            }
+        }
+
+        return KoSyncConnectPayload(false, authResult.message ?: "Authentication failed.", null)
+    }
+
+    suspend fun logout() {
+        GlobalConfigManager.updateValue("server.koreaderSyncUsername", "")
+        GlobalConfigManager.updateValue("server.koreaderSyncUserkey", "")
```
Syer10: This needs to set the values in serverConfig, not GlobalConfigManager
```suggestion
        serverConfig.koreaderSyncUsername.value = ""
        serverConfig.koreaderSyncUserkey.value = ""
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/sync/KoreaderSyncService.kt
```diff
@@ -0,0 +1,359 @@
...
+            GlobalConfigManager.updateValue("server.koreaderSyncUserkey", userkey)
+            return KoSyncConnectPayload(true, "Login successful.", username)
+        }
+
+        if (authResult.isUserNotFoundError) {
+            logger.info { "[KOSYNC CONNECT] Authorization failed, attempting to register new user." }
+            val registerResult = register(username, userkey)
+            return if (registerResult.success) {
+                GlobalConfigManager.updateValue("server.koreaderSyncUsername", username)
+                GlobalConfigManager.updateValue("server.koreaderSyncUserkey", userkey)
```
Syer10:
```suggestion
                serverConfig.koreaderSyncUsername.value = username
                serverConfig.koreaderSyncUserkey.value = userkey
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/sync/KoreaderSyncService.kt
```diff
@@ -0,0 +1,359 @@
...
+    suspend fun connect(
+        username: String,
+        password: String,
+    ): KoSyncConnectPayload {
+        val userkey = Hash.md5(password)
+        val authResult = authorize(username, userkey)
+
+        if (authResult.success) {
+            GlobalConfigManager.updateValue("server.koreaderSyncUsername", username)
+            GlobalConfigManager.updateValue("server.koreaderSyncUserkey", userkey)
```
Syer10:
```suggestion
           serverConfig.koreaderSyncUsername.value = username
           serverConfig.koreaderSyncUserkey.value = userkey
```

Syer10 (changes requested):
A quick review, will do more later.

This will likely have to be merged in a week or 2, as I will be unavailable for the next week.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
```diff
@@ -105,27 +107,6 @@ abstract class ChaptersFilesProvider<Type : FileType>(
         scope: CoroutineScope,
         step: suspend (DownloadChapter?, Boolean) -> Unit,
     ): Boolean {
-        val existingDownloadPageCount =
```
Syer10: Looks like you accidentally reverted one of the recent commits 😄

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
```diff
@@ -189,6 +191,25 @@ abstract class ChaptersFilesProvider<Type : FileType>(
...
+                    }
+                }
+            }
+        }
+
+        transaction {
+            ChapterTable.update({ ChapterTable.id eq chapterId }) {
+                it[ChapterTable.pageCount] = getImageCount()
+            }
+        }
```
Syer10: This can be removed, as we already do this in `ChapterForDownload.getChapterDownloadReady`

Syer10 (approved):

Syer10:
I will merge after I add the settings to the Launcher and the Docker container

zeedif:
@Syer10, before this gets merged, I had a few ideas for improving the KOReader sync while working on an OPDS implementation for Readest.

1. I was thinking of adding `prompt` (to confirm changes) and `disable` (to pause sync without logging out). I'd also rename the strategies to be more intuitive: `prompt` | `silent` (replaces LATEST) | `send` (replaces SUWAYOMI) | `receive` (replaces KOSYNC) | `disable`.

2. What do you think about a customizable `device` name? It might be more of a frontend thing, but it would be great for differentiating clients like Sorayomi or multiple WebUIs on the same server.

3. Also, does the server handle any other file types besides CBZ? I'm asking because KOReader makes a distinction between paginated files (like CBZ) and reflowable ones (like EPUB), where the progress format changes from a page number to an XPATH-like string. My current implementation isn't prepared to handle the XPATH format yet.

4. For reflowable files, since my PR can't parse the XPath, we'd need to fall back to comparing the percentage. This also helps with edge cases where checksums might match (especially with filename hashing) but page counts differ, making direct page comparison impossible. To prevent false conflicts when percentages are nearly identical (e.g., 31.8% vs 31.9%), I implemented a configurable "tolerance" in Readest (koreaderSyncPercentageTolerance). It treats progress as synced if the difference is within a small, user-defined margin.

Let me know what you think.

Syer10:
> 1. I was thinking of adding `prompt` (to confirm changes) and `disable` (to pause sync without logging out). I'd also rename the strategies to be more intuitive: `prompt` | `silent` (replaces LATEST) | `send` (replaces SUWAYOMI) | `receive` (replaces KOSYNC) | `disable`.

Seems fine to me, I don't have any problem with that. Seems like a good solution.

> 2\. What do you think about a customizable `device` name? It might be more of a frontend thing, but it would be great for differentiating clients like Sorayomi or multiple WebUIs on the same server

I don't know about this one, what is the idea behind differentiating clients? Suwayomi doesn't need to know what client is making changes as they all use the same api to both read and make changes.

> 3\. Also, does the server handle any other file types besides CBZ? I'm asking because KOReader makes a distinction between paginated files (like CBZ) and reflowable ones (like EPUB), where the progress format changes from a page number to an XPATH-like string. My current implementation isn't prepared to handle the XPATH format yet.

No plans to support anything other then CBZ outside local source. We want the downloaded files to be simple and match Tachiyomi/Mihon. 

> 4\. For reflowable files, since my PR can't parse the XPath, we'd need to fall back to comparing the percentage. This also helps with edge cases where checksums might match (especially with filename hashing) but page counts differ, making direct page comparison impossible. To prevent false conflicts when percentages are nearly identical (e.g., 31.8% vs 31.9%), I implemented a configurable "tolerance" in Readest (koreaderSyncPercentageTolerance). It treats progress as synced if the difference is within a small, user-defined margin.

That would work for me.

zeedif:
I've just pushed the changes incorporating the new sync strategy and tolerance settings.

Since I'm not deeply familiar with the web UI's caching behavior, what I do is avoid updating the local database when the sync strategy is `PROMPT` so as to:

*   When the client calls `fetchChapterPages`, the `SyncResult` payload will now include a boolean `isConflict`.
*   If `isConflict` is true (indicating the `PROMPT` strategy found a difference), the payload provides the updated remote page number, but the server **does not** update its own database record for the chapter.
*   The web UI can then compare this remote page number with its cached value. If they differ, it can display a prompt to the user asking if they want to accept the remote progress.
*   If the user accepts, the client can then explicitly call the `updateChapter` mutation to set the new `lastPageRead`, which will then be persisted.

This delegates the "prompt" logic to the client, preventing unwanted database writes on the server for this specific strategy.

---

#### Regarding `koreaderSyncPercentageTolerance`

This setting avoids sync churn when the progress difference between local and remote is negligible (e.g., rounding differences between devices).

In the Readest UI, I've implemented this as a slider. Instead of having the user input a tiny decimal like `0.01`, the slider operates on a logarithmic scale that represents decimal precision. The tolerance value is calculated as **10<sup>-N</sup>**, where `N` is the slider's value.

The slider for `N` ranges from **0 to 15**. This upper limit of 15 was chosen because the KOReader sync service itself stores the progress percentage with a precision of 15 decimal places, so any greater precision would be ignored by the service anyway.

For example:
*   A slider value of **2** corresponds to a tolerance of 10<sup>-2</sup> = **0.01** (or 1%). Any progress difference less than 1% will be ignored.
*   A slider value of **4** corresponds to a tolerance of 10<sup>-4</sup> = **0.0001** (or 0.01%).

This offers more intuitive and granular control over the very small values this setting requires.

---

#### A Note on OPDS Behavior

Due to the stateless nature of OPDS feeds, the `PROMPT` strategy has a slight behavior change in that context. When an OPDS client requests a chapter's metadata (which is when the progress sync check occurs), there is no mechanism to "prompt" the user. Therefore, for OPDS clients, PROMPT will behave like SILENT: if a conflict is detected, the server will silently accept the most recent progress and update the local database.

Although instead it could offer two separate acquisition links in the chapter metadata feed (one for local progress and one for remote) so that the OPDS client can decide, for now this seems like a reasonable solution.

Let me know what you think. Looking forward to your review

Syer10 (approved):
Seems good to me. I'l look at the launcher and docker container soon

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt
```diff
@@ -248,9 +263,28 @@ class ChapterMutation {
         val paramsMap = input.toParams()
 
         return future {
-            asDataFetcherResult {
-                val chapter = getChapterDownloadReadyById(chapterId)
+            var chapter = getChapterDownloadReadyById(chapterId)
```
schroda: Mutations need to be wrapped in `asDataFetcherResult` to handle partial successful mutation responses.
Otherwise, in case a request includes multiple mutations, one failing mutation would lead to all mutations to not return any data in the request

schroda (changes requested):

server/src/main/resources/server-reference.conf
```diff
@@ -85,3 +85,12 @@ server.opdsMarkAsReadOnDownload = false
...
 server.opdsChapterSortOrder = "DESC" # "ASC", "DESC"
+
+# Koreader Sync
+server.koreaderSyncEnabled = false
+server.koreaderSyncServerUrl = "http://localhost:17200"
+server.koreaderSyncUsername = ""
+server.koreaderSyncUserkey = ""
+server.koreaderSyncDeviceId = ""
+server.koreaderSyncChecksumMethod = "binary" # "binary" or "filename"
+server.koreaderSyncStrategy = "latest" # "latest", "suwayomi", "kosync"
```
schroda: This looks to be outdated, `latest` does not exist in the enum.
It probably also should be `disable` by default, otherwise, the koreader sync will kick in and fail for every user, regardless of them using koreader or not?

server/src/main/resources/server-reference.conf
```diff
@@ -85,3 +85,12 @@ server.opdsMarkAsReadOnDownload = false
 server.opdsShowOnlyUnreadChapters = false
 server.opdsShowOnlyDownloadedChapters = false
 server.opdsChapterSortOrder = "DESC" # "ASC", "DESC"
+
+# Koreader Sync
+server.koreaderSyncEnabled = false
```
schroda: This does not get used anywhere

server/src/main/resources/server-reference.conf
zeedif: The sync logic is gated by the koreaderSyncUsername and koreaderSyncUserkey fields not being empty, so the strategy's default value doesn't affect users who haven't configured them. Plus, the connect mutation already sets the strategy to prompt on a successful login. So, the default value mainly serves as a safeguard for users who might set their username and userkey directly via environment variables, bypassing the login flow.

schroda: > The sync logic is gated by the koreaderSyncUsername and koreaderSyncUserkey fields not being empty, so the strategy's default value doesn't affect users who haven't configured them.

I see. Looks like I missed that.

>  Plus, the connect mutation already sets the strategy to prompt on a successful login

What's the reasoning behind that? I would assume that this will get reported as a bug in the future.

If a user changed the sync strategy and at some point logged out and back in again, this would just randomly change the previously set sync strategy without the user realizing that it happened.

Additionally, it looks like the connect mutation doesn't inform the client about it mutating this setting. Which would lead to an out-of-date cache on the client.

schroda: > Additionally, it looks like the connect mutation doesn't inform the client about it mutating this setting. Which would lead to an out-of-date cache on the client.

I just realized that the connect mutation mutates the server settings but does not return the SettingType.
Which means that the client has to manually update its cache (username and password) instead of it happening automatically through the returned type

zeedif:
Okay, I've implemented the suggestions. The `connectKoSyncAccount` mutation no longer overwrites the sync strategy, and the default in `server-reference.conf` is now correctly set to `DISABLED`.

Building on that, I've improved the conflict handling for both OPDS and GraphQL clients:
- **OPDS:** When a conflict occurs, the feed now provides two separate links: one for local progress and one for remote, showing the device name (e.g., "Continue Reading Online (Synced from KOReader)").
- **GraphQL:** The `fetchChapterPages` mutation now returns a `syncConflict` object with the `deviceName` and `remotePage`, so the UI can build a more informative prompt like "Sync reading progress from 'KOReader'?".

This should address all the points from the discussion. Let me know what you think

schroda:
Looks good.
I'd like to have the koreader sync mutations return the SettingsType in the response, to make cache handling on the client easier (automatic in the case of apollo client). Like mentioned here https://github.com/Suwayomi/Suwayomi-Server/pull/1560#discussion_r2271891685

server/src/main/resources/server-reference.conf
```diff
@@ -85,3 +85,11 @@ server.opdsMarkAsReadOnDownload = false
...
 server.opdsShowOnlyDownloadedChapters = false
 server.opdsChapterSortOrder = "DESC" # "ASC", "DESC"
+
+# Koreader Sync
+server.koreaderSyncServerUrl = "http://localhost:17200"
+server.koreaderSyncUsername = ""
+server.koreaderSyncUserkey = ""
+server.koreaderSyncDeviceId = ""
+server.koreaderSyncChecksumMethod = "binary" # "binary" or "filename"
+server.koreaderSyncStrategy = "disabled" # "prompt", "silent", "send", "receive", "disabled"
```
Syer10, line 95: This is missing the server.koreaderSyncPercentageTolerance configuration

Syer10 (changes requested):

Syer10 (approved):

schroda (approved):

PR #1612, opened 2025-08-22
Feature/graphql settings add jwt settings

schroda:


server/src/main/kotlin/suwayomi/tachidesk/graphql/server/primitives/DurationAsString.kt
```diff
@@ -0,0 +1,122 @@
...
+        .name("Duration")
+        .description("An ISO-8601 encoded duration string")
+        .coercing(GraphqlDurationAsStringCoercing())
+        .build()
+
+private class GraphqlDurationAsStringCoercing : Coercing<Duration, String> {
+    private fun toStringImpl(input: Any): String =
+        when (input) {
+            is Duration -> input.toIsoString()
+            is String -> Duration.parse(input).toIsoString()
```
Syer10, line 28: Everywhere else we use Duration.toString(), so its probably best to standardize on that. Unless there is a drawback to that.

```suggestion
            is Duration -> input.toString()
            is String -> Duration.parse(input).toString()
```

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/server/primitives/DurationAsString.kt
schroda: While kotlin Duration seems to support parsing these types of strings, they are not valid ISO 8601 representations.
So there is no guarantee that other libraries will be able to work with these kinds of strings.
The library used in the webUI is dayjs, which does not understand the format created by toString e.g., 60d; it requires it to be of proper ISO 8601 representation which would be P60D.

While the toString result (e.g., 60d) is easier to work with in the config for humans, I think the api should return a proper ISO 8601 representation.

Syer10 (approved):

PR #1614, opened 2025-08-24
Feature/streamline settings

schroda:
Currently, there is a lot of overhead with the settings.
They have to be updated in the 
- ServerConfig
- server-reference.conf files
- GraphQL setting types
- GraphQL setting mutation (validation and updating)
- Handling of settings in exporting/importing backups

This PR has the goal to make the ServerConfig the single source of truth for settings.

- [x] Streamlining GraphQL setting mutation
  - [x] Validating settings
  - [x] Updating settings
- [x] Generating the server-reference.conf files
- [x] Generating the graphql SettingsType
- [x] Backups
  - [x] Exporting
  - [x] Importing
- [x] Up-to-date generated files are used on run/build

Additionally, this adds validating values when reading them from the config. In case a setting is invalid, it will get converted to a valid value, or if this is not possible, fall back to the default setting.

----
- New module `server-config-generate` which contains the generation code
- New module `server-config` which contains all setting related code which is shared between  `server-main` and `server-config-generate`
- Generated files are created in a build folder and are not part of the src tree anymore


server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -0,0 +1,721 @@
...
+    )
+
+    val basicAuthEnabled: MutableStateFlow<Boolean> by MigratedConfigValue(
+        protoNumber = 29,
+        defaultValue = false,
+        group = SettingGroup.AUTH,
+        deprecated =
+            SettingsRegistry.SettingDeprecated(
+                replaceWith = "authMode",
+                message = "Removed - prefer authUsername",
```
cpiber:
```suggestion
                message = "Removed - prefer authMode",
```

Syer10 (approved):

blokyk:
The wiki's [page on config] still has a link to the now deleted reference config (which in my opinion is very useful), so would it be possible to host a version of it on a gist or something? or somehow link to a build artifact or some instance of the default settings?

in the meantime, it could at least link to [the last setting file static version](https://github.com/Suwayomi/Suwayomi-Server/blob/42fda173f920bfe9de369dedc8cfa4f6d93c7afd/server/src/main/resources/server-reference.conf)

Issue #1226, opened 2025-01-09
[Feature Request] Support for other databases


Tsabo:
## What feature should be added to Suwayomi?
The ability to use a different database other than H2. 

##  Why/Project's Benefit/Existing Problem
Performance, third-party querying outside the API and GraphQL, and backup strategies. 

It appears that you are using JetBrains Exposed for the data layer which does provide support for other databases:
https://jetbrains.github.io/Exposed/working-with-database.html#mysql

[DBManager.kt#L24](https://github.com/Suwayomi/Suwayomi-Server/blob/master/server/src/main/kotlin/suwayomi/tachidesk/server/database/DBManager.kt#L24)

```java
Database.connect(
    "jdbc:mysql://localhost:3306/test",
    driver = "com.mysql.cj.jdbc.Driver",
    user = "user",
    password = "password"
)
```

I'm happy to take a look. I'm a .NET developer by day. Java isn't my thing, so I have to work through some build issues before I could do anything.

Robonau:
the bot is just gonna auto close it unless that first line is deleted

Syer10:
I would be open to a PR adding support for other databases, PostgreSQL should be compatible with how we use H2. This would need to be tested with multiple different GraphQL queries, specifically each filtering option, since we do use some custom queries in H2 like the DistinctOn function.

Robonau:
syer has said [PostgreSQL should be possible](https://discord.com/channels/801021177333940224/855025275258929162/1197639864217239562) in the discord before

Robonau:
though an external DB isn't gonna be better for performance
the network overhead alone should make any gains from a faster DB irrelevant
3rd party queries and backups are still valid tho i guess

Syer10:
There is also the issue of migrations, I am not sure how that has to be handled. Maybe the plain SQL we use will be compatible, or maybe we will need migrations for each database type.

Robonau:
i mean for postgres and other SQL compatable DBs migrations in plain SQL should work fine

Tsabo:
I'm happy that this stirred up some interest. I'll spend the weekend getting my environment set up (necessary requirements, etc..) and playing around with it. I make no promises, but I am motivated for this to be implemented.

Robonau:
if u run in to any issues or need help with yr environment or whatnot you are probably best asking in the discord, stuff like that will get lost on github

Syer10:
Closing since we have PostgreSQL now

PR #1617, opened 2025-08-25
Support PostgreSQL Databases

Syer10:
Related to https://github.com/Suwayomi/Suwayomi-Server/issues/1226


server/src/main/kotlin/suwayomi/tachidesk/server/database/DBManager.kt
```diff
@@ -13,24 +13,43 @@ import io.github.oshai.kotlinlogging.KotlinLogging
...
-                    preserveKeywordCasing = false
-                },
-        )
+        val dbConfig =
+            DatabaseConfig {
+                useNestedTransactions = true
+                @OptIn(ExperimentalKeywordApi::class)
+                preserveKeywordCasing = false
+            }
+        when (serverConfig.databaseType.value) {
```
schroda: What about reconnecting to the database on setting changes?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SettingsMutation.kt
```diff
@@ -133,6 +133,12 @@ class SettingsMutation {
         updateSetting(settings.ip, serverConfig.ip)
         updateSetting(settings.port, serverConfig.port)
 
+        // database
+        updateSetting(settings.databaseType, serverConfig.databaseType)
+        updateSetting(settings.databaseUrl, serverConfig.databaseUrl)
```
schroda: Should we validate that the database exists?

Syer10: I don't think there is an easy way to check, I feel it would be better to leave it to the user.

server/src/main/kotlin/suwayomi/tachidesk/server/database/DBManager.kt
Syer10: Added 👍

schroda (approved):

PR #1624, opened 2025-09-02
refactor(kosync): introduce differentiated sync strategies

zeedif:
This PR refactors the KOReader sync strategy to provide more granular control over conflict resolution.

Previously, a single strategy setting was applied universally. Now, users can define separate strategies for two distinct scenarios:
1.  **Forward Sync:** When the remote progress is newer than the local progress.
2.  **Backward Sync:** When the remote progress is older than the local progress.

This is achieved by deprecating the old `koreaderSyncStrategy` and introducing two new settings: `koreaderSyncStrategyForward` and `koreaderSyncStrategyBackward`.

The `SILENT` option has been removed from the strategy enum because its behavior is now implicitly defined by the user's choices. For example:
- Setting `Forward` to `KEEP_REMOTE` and `Backward` to `KEEP_LOCAL` perfectly replicates the old `SILENT` behavior (always sync to the most recent timestamp).

All previous functionalities are retained and new, more complex combinations are now possible, giving users greater customization. Existing user settings are automatically migrated to the new system for a smooth transition.

### DIFF
--- a/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -654,26 +654,8 @@ class ServerConfig(
         },
     )
 
-    val koreaderSyncStrategyForward: MutableStateFlow<KoreaderSyncConflictStrategy> by EnumSetting(
-        protoNumber = 65,
-        group = SettingGroup.KOREADER_SYNC,
-        defaultValue = KoreaderSyncConflictStrategy.PROMPT,
-        enumClass = KoreaderSyncConflictStrategy::class,
-        typeInfo = SettingsRegistry.PartialTypeInfo(imports = listOf("suwayomi.tachidesk.graphql.types.KoreaderSyncConflictStrategy")),
-        description = "Strategy to apply when remote progress is newer than local.",
-    )
-
-    val koreaderSyncStrategyBackward: MutableStateFlow<KoreaderSyncConflictStrategy> by EnumSetting(
-        protoNumber = 66,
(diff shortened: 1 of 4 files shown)


server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
Syer10, line 1: New settings need to be added to the bottom of the serverconfig and given the next proto number slot, they can't be added in the middle.

Syer10 (changes requested):

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -599,12 +600,58 @@ class ServerConfig(
         typeInfo = SettingsRegistry.PartialTypeInfo(imports = listOf("suwayomi.tachidesk.graphql.types.KoreaderSyncChecksumMethod")),
     )
 
-    val koreaderSyncStrategy: MutableStateFlow<KoreaderSyncStrategy> by EnumSetting(
+    @Deprecated("Use koreaderSyncStrategyForward and koreaderSyncStrategyBackward instead")
+    val koreaderSyncStrategy: MutableStateFlow<KoreaderSyncLegacyStrategy> by MigratedConfigValue(
```
schroda, line 604: I just noticed that I've missed https://github.com/Suwayomi/Suwayomi-Server/blob/master/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt#L310 in the settings cleanup, so you'd have to add the migration logic here as well for now.
Otherwise, the current setting gets lost on a server start.

Syer10 (approved):

PR #1627, opened 2025-09-02
Optimize Download Queue

Syer10:
Don't store a Manga and Chapter in memory in the download queue.

The old api websocket is probably going to get slower with this PR, but its deprecated anyway.


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/model/OldDownloadStatus.kt
schroda, line 1: Was the file name change intentional?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -195,11 +191,17 @@ object DownloadManager {
         }
 
         if (immediate) {
-            val status = getStatus()
```
schroda, line 198: I think the reason to get the status outside the launched scope was to make sure that the status is from the point of time the function got called. Otherwise, it might have changed in the meantime and some changes get dropped.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/model/OldDownloadStatus.kt
Syer10: Nope, fixed.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
Syer10: Fixed

schroda (approved):

PR #1658, opened 2025-09-19
Add server-side subpath support for WebUI

renjfk:
Enables server-side configuration for hosting WebUI under non-root URL paths (e.g., /manga) with automatic client injection.

## Changes
- Add webUISubpath configuration setting with regex validation
- Create temporary WebUI directory for subpath serving  
- Inject subpath config into index.html for client detection
- Support all WebUI flavors with subpath compatibility
- Update browser opening to use subpath URLs

Related to Suwayomi/Suwayomi-WebUI#174

### DIFF
--- a/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -715,14 +715,6 @@ class ServerConfig(
         requiresRestart = true,
     )
 
-    val webUISubpath: MutableStateFlow<String> by StringSetting(
-        protoNumber = 75,
-        group = SettingGroup.WEB_UI,
-        defaultValue = "",
-        pattern = "^(/[a-zA-Z0-9._-]+)*$".toRegex(),
-        description = "Serve WebUI under a subpath (e.g., /manga). Leave empty for root path. Must start with / if specified.",
-    )
-
     val databaseType: MutableStateFlow<DatabaseType> by EnumSetting(
         protoNumber = 69,
         group = SettingGroup.DATABASE,
@@ -766,6 +758,16 @@ class ServerConfig(
         typeInfo = SettingsRegistry.PartialTypeInfo(imports = listOf("suwayomi.tachidesk.graphql.types.KoreaderSyncConflictStrategy")),
         description = "Strategy to apply when remote progress is older than local.",
     )
+
+    val webUISubpath: MutableStateFlow<String> by StringSetting(
+        protoNumber = 75,
+        group = SettingGroup.WEB_UI,
+        defaultValue = "",
+        pattern = "^(/[a-zA-Z0-9._-]+)*$".toRegex(),
+        description = "Serve WebUI under a subpath (e.g., /manga). Leave empty for root path. Must start with / if specified.",
+        requiresRestart = true,
+    )
+
 
     /** ****************************************************************** **/
     /**                                                                    **/
--- a/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -715,6 +715,14 @@ class ServerConfig(
         requiresRestart = true,
     )
 
+    val webUISubpath: MutableStateFlow<String> by StringSetting(
+        protoNumber = 75,
+        group = SettingGroup.WEB_UI,
+        defaultValue = "",
+        pattern = "^(/[a-zA-Z0-9._-]+)*$".toRegex(),
+        description = "Serve WebUI under a subpath (e.g., /manga). Leave empty for root path. Must start with / if specified.",
+    )
+
     val databaseType: MutableStateFlow<DatabaseType> by EnumSetting(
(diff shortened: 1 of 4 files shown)


server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -715,6 +715,14 @@ class ServerConfig(
         requiresRestart = true,
     )
 
+    val webUISubpath: MutableStateFlow<String> by StringSetting(
```
schroda: Settings are sorted by `protoNumber`. So this should be after `koreaderSyncStrategyBackward`

schroda: This is currently marked as a setting that can be changed during runtime, but is not implemented as such.
If it is not changeable during runtime, it needs to set `requiresRestart = true`. Otherwise, the javalin app would need to handle the sub path change.

server/src/main/kotlin/suwayomi/tachidesk/server/JavalinSetup.kt
```diff
@@ -65,18 +65,66 @@ object JavalinSetup {
                 val templateEngine = TemplateEngine.createPrecompiled(ContentType.Html)
                 config.fileRenderer(JavalinJte(templateEngine))
                 if (serverConfig.webUIEnabled.value) {
-                    val serveWebUI = {
-                        config.spaRoot.addFile("/", applicationDirs.webUIRoot + "/index.html", Location.EXTERNAL)
-                    }
-                    WebInterfaceManager.setServeWebUI(serveWebUI)
```
schroda, line 71: This breaks the automatic update logic of the webUI files

renjfk: Actually it doesn't because it works on a copy not original dir

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
renjfk: Btw does `protoNumber` have any special meaning? So if I squeeze new setting into middle by adjusting the rest, would that break anything?

Syer10: Proto number is the unique number assigned to a setting in the backup. To keep backup compatibility we cannot change the numbers.

server/src/main/kotlin/suwayomi/tachidesk/server/JavalinSetup.kt
renjfk: Oh wait a second, it's missing the `suwayomi.tachidesk.server.util.WebInterfaceManager#setServeWebUI` call 🤦

schroda: After looking again, the temp folder also needs to be updated with the newly updated files (https://github.com/Suwayomi/Suwayomi-Server/blob/808e0ecae71db4a18e1c9cd4d7b595814faf94f2/server/src/main/kotlin/suwayomi/tachidesk/server/util/WebInterfaceManager.kt#L612).

Syer10 (changes requested):
We return relative paths to endpoints in multiple places. Please verify using a find all that all `api/v1` endpoints are changed to include the subpath.

Unless this is something that clients also need to manually handle.

renjfk:
> We return relative paths to endpoints in multiple places. Please verify using a find all that all `api/v1` endpoints are changed to include the subpath.
> 
> Unless this is something that clients also need to manually handle.

That's a very good point. Currently that's handled on the client side with WebUI (https://github.com/Suwayomi/Suwayomi-WebUI/pull/1011/files#diff-7548c07a15bf41d897bd85546249170862b95a2e37590e82ec5a7faa6ba26707R75) but I'm not sure what would be the best solution considering other clients 🤔

Syer10:
I've given it some more thought, its probably best that we leave it for the clients to handle. Clients will need to manually add support if the way they handle the URL breaks with this change. I think the Suwayomi extension will work out of the box as long as the user inputs the subpath.

schroda (approved):

server/src/main/kotlin/suwayomi/tachidesk/server/JavalinSetup.kt
```diff
@@ -65,18 +65,79 @@ object JavalinSetup {
...
-                    config.staticFiles.add(applicationDirs.webUIRoot, Location.EXTERNAL)
-                    serveWebUI()
+                    // Helper function to create a servable WebUI directory with subpath injection
+                    fun createServableWebUIRoot(): String =
+                        if (subpath.isNotBlank()) {
+                            val tempWebUIRoot = WebInterfaceManager.createServableWebUIDirectory()
+
+                            // Inject subpath configuration
+                            val indexHtmlPath = "$tempWebUIRoot/index.html"
+                            val indexHtmlFile = java.io.File(indexHtmlPath)
```
Syer10: Import File

Syer10 (changes requested):
Should be fine after this

Syer10 (approved):

PR #1659, opened 2025-09-20
Use new Kitsu search URL

cpiber:
The algolia URL changed recently, so now searching always returns "Remote host terminated the handshake"


cpiber:
As mentioned on mihonapp/mihon#2517, it's my network that blocks the old URL. Still, an update to the new URL is probably due to not run into problems later on

PR #1671, opened 2025-09-27
Use graphql directive for auth handling

schroda:
Is cleaner and easier to use (see also https://www.graphql-java.com/documentation/directives/#adding-behaviour)

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/directives/RequireAuth.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/directives/RequireAuth.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) Contributors to the Suwayomi project
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+package suwayomi.tachidesk.graphql.directives
+
+import com.expediagroup.graphql.generator.annotations.GraphQLDirective
+import graphql.introspection.Introspection.DirectiveLocation
+
+@GraphQLDirective(
+    name = "requireAuth",
+    description = "Requires user authentication",
+    locations = [
+        DirectiveLocation.FIELD_DEFINITION,
+        DirectiveLocation.OBJECT,
+    ],
+)
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class RequireAuth
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/directives/RequireAuthDirectiveWiring.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/directives/RequireAuthDirectiveWiring.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) Contributors to the Suwayomi project
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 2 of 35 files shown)


cpiber:
Could this still be used to get the user id if #623 goes through?

schroda:
It still reads the user from the graphql context, so in that regard nothing changes. If that is what you mean?

cpiber:
I mean, for that PR they actually need the user details, to query for the appropriate user, e.g. https://github.com/Syer10/Tachidesk/blob/932f5861f0823cb0c5b13aa3f6fa0c6380b935eb/server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/MangaQuery.kt#L248

So, the question is if this attribute still allows the code to get access to that ID, or if this would then require going back to what we have at the moment

schroda:
Yes, nothing changes there.

cpiber:
Is this PR then not counter-productive, if #623 needs to effectively revert it to work? My understanding is that this is a goal, even if it's not high priority

schroda:
I'm confused. Nothing changes, the user/user id is in the graphql context and can be accessed from anywhere you have access to the context.
This PR only streamlines the auth check a little by using the annotation instead of having to use the data fetching env arg, get the user and call `requireUser` everywhere.
The auth check is still done the exact same way as before, just in a single location instead of in every query, mutation, etc.

Currently:
```
fun myMutation/Query/Subscription(
        dataFetchingEnvironment: DataFetchingEnvironment,
        input: Input,
    ): CompletableFuture<RestoreBackupPayload> {
        dataFetchingEnvironment.getAttribute(Attribute.TachideskUser).requireUser()
        ....
}
```

With this change

```
@RequireAuth
fun myMutation/Query/Subscription(
        input: Input,
    ): CompletableFuture<RestoreBackupPayload> {
        ....
}
```

And in the future when adding different permissions, roles the annotation can be updated to include them as well e.g.,

```
@RequireAuth(roles = ["admin", ...], permissions = [...])
```

cpiber:
I mean, to get the ID, one still has to call `val userId = dataFetchingEnvironment.getAttribute(Attribute.TachideskUser).requireUser()`, so the annotation is redundant then

Syer10:
We will still need to use dataFetchingEnviornment requireUser to get the user id in each query/mutation, is it possible for that to be simplified along with this?

schroda:
Yes, if the userId is needed to do stuff then it needs to be retrieved from the context. I think there is no way around this and the only way to make things available in the fetchers and data loaders.

I think it's still a cleaner way of handling authentication, even if you have to get the user id from the context.
It's easier to see, read and define. And at the point of time you are inside the query/mutation you know the user is authorized for it.

From what I see there are at least a few queries where `requireUser` is called in the query itself to "authenticate" the user and then again in the data loader to actually get the user id (makes the `requireUser` function a little confusing imo).
With this, you would just add the annotation to the query and access the user id in the data loader from the context.

schroda:
Ok, after looking a bit, it is possible to pass the userId as an argument to the data fetcher. However, you would still need to read it from the context in a data loader

schroda:
[This ](https://github.com/Suwayomi/Suwayomi-Server/pull/1671/files#diff-c677c42714ce7b3b1f7f3dacf49ef0db1f16bfff896620ad4ea71c78b39ef989R32) would add the userId as an argument which than could be used like a normaler parameter in the data fetcher. `userId` would then be a reserved parameter.

```
    @RequireAuth
    fun restoreBackup(
        @GraphQLIgnore
        userId: Int,
        input: RestoreBackupInput,
    ): CompletableFuture<RestoreBackupPayload> {
```

Syer10:
That looks great. I think it would be good if an error is thrown if someone tries to pass it manually

Syer10 (approved):

PR #1678, opened 2025-09-28
REAL-ESRGAN Support

FadedSociety:
Added Real ESRGAN support for suwayomi using https://github.com/FadedSociety/Real-ESRGAN-Service/releases/tag/RELEASE refer to Robo and Syer10 on suwayomi discord about this

PR #1684, opened 2025-09-29
Remote Image Processing

FadedSociety:


server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -763,15 +760,12 @@ class ServerConfig(
     )
 
     val webUISubpath: MutableStateFlow<String> by StringSetting(
-        protoNumber = 75,
+        protoNumber = 76,
         group = SettingGroup.WEB_UI,
         defaultValue = "",
-        pattern = "^(/[a-zA-Z0-9._-]+)*$".toRegex(),
```
Robonau, line 830: why are you now changing webUISubpath config?
and what happened to the conversion config?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
```diff
@@ -142,20 +145,49 @@ object Page {
...
     ): Pair<InputStream, String> {
-        val imageExtension = MimeUtils.guessExtensionFromMimeType(image.second) ?: image.second.removePrefix("image/")
+        var currentImage = image
+        var currentMimeType = image.second
+
+        // Step 1: Check if HTTP upscaling is configured
+        val conversions = serverConfig.downloadConversions.value
+        val defaultConversion = conversions["default"]
+        val imageType = image.second
+        val conversion = conversions[imageType] ?: defaultConversion
```
cpiber: Better than what you had in #1678, but I think it would make sense to allow both upscaling and converting, i.e. first upscale if desired, then convert (as mentioned on discord)

Robonau: its not really specifically for upscaling, a converter server could also do the conversion to webp or whatever

my use-case for this feature would be just to have access to all the options when converting to webp that the current conversion system hides

dejavui:
You should create a new branch, not use the master branch to create pull requests (⌐■_■)

FadedSociety (commented):
👍

schroda (approved):

docs/Configuring-Suwayomi‐Server.md
```diff
@@ -78,7 +78,7 @@ server.downloadConversions = {}
 - `server.excludeEntryWithUnreadChapters = true` controls if Suwayomi will download new chapters for titles with unread chapters (requires `server.autoDownloadNewChapters`).
 - `server.autoDownloadNewChaptersLimit = 0` sets how many chapters should be downloaded at most, `0` to disable the limit; if the limit is reached, new chapters will not be downloaded (requires `server.autoDownloadNewChapters`).
 - `server.autoDownloadIgnoreReUploads = false` controls if Suwayomi will re-download re-uploads on update (requires `server.autoDownloadNewChapters`).
-- `server.downloadConversions = {}` configures optional image conversions for all downloads. This is an [JSON object](https://en.wikipedia.org/wiki/JSON#Syntax), with the source image [mime type](https://en.wikipedia.org/wiki/Media_type) as the key and an object with the target mime type and options as value.  
+- `server.downloadConversions = {}` configures optional image conversions for all downloads. This is an [JSON object](https://en.wikipedia.org/wiki/JSON#Syntax), with the source image [mime type](https://en.wikipedia.org/wiki/Media_type) as the key and an object with the target mime type or url and options as value.
```
cpiber, line 81: Maybe we should document what the `url` actually does.
- That the URL represents an external converter (and e.g. upscaler)
- What parameters such an API should expect and what it is expected to return

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
```diff
@@ -250,7 +251,9 @@ abstract class ChaptersFilesProvider<Type : FileType>(
             val targetConversion = conversion ?: defaultConversion ?: return@forEach
 
             val (targetMime) = targetConversion
-            val requiresConversion = imageType != targetMime && targetMime != "none"
+            val requiresConversion =
+                ConversionUtil.isHttpPostProcess(targetConversion) ||
+                    imageType != targetMime && targetMime != "none"
```
cpiber: Aren't you missing the `ConversionUtil.imageHttpPostProcess` below in `convertPage`?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
```diff
@@ -134,28 +145,71 @@ object Page {
...
-        val imageExtension = MimeUtils.guessExtensionFromMimeType(image.second) ?: image.second.removePrefix("image/")
+        var currentImage = image.first
+        var currentMimeType = image.second
+
+        val conversions = serverConfig.serveConversions.value
+        val defaultConversion = conversions["default"]
+        val conversion = conversions[currentMimeType] ?: defaultConversion
+
+        // Apply HTTP post-process if configured (complementary with format conversion)
+        if (!isDownloader && conversion != null && ConversionUtil.isHttpPostProcess(conversion)) {
```
cpiber: I'm not quite sure I understand your intensions here, does this not mean non-url conversions are ignored on serve?

Together with the other comment is seems like on download we only support regular convert, and on serve only url-based. If so, the documentation is misleading.

cpiber (changes requested):
Please note that I haven't had time to compile the thing and play with it, so my observations are purely from reading the code, so if I missed something feel free to ignore my comments. I'll look into it more closely in the evening.

cpiber (changes requested):
To amend my previous review:

- When downloading, we first see `Page.convertImageResponse(isDownloader = true)` for all pages individually, then `ChapterFilesProvider.maybeConvertPages` at the end
- When serving directly from source or from download, we see `Page.convertImageResponse(isDownloader = false)` for all requested pages

So my observation was correct, when serving the page the current implementation ignores all non-http image conversion except those explicitly requested via the `format` query parameter. On the other hand, downloads get converted separately by ChapterFilesProvider.
I see a simple solution: Move the conversion logic to `Page`, and use either `serveConversions` or `downloadConversions` based on `isDownload`. No need for the post-conversion then. I have originally implemented it split like this, but this was only because I forgot to clean it up, I see no reason for architecting it this way. Note also that the current implementation in `Page` does not even consider the parameter, which could be fixed by merging the implementation.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
Syer10: Fixed

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
Syer10: Fixed

docs/Configuring-Suwayomi‐Server.md
Syer10: I improved the documentation and added a curl example

cpiber:
Config:
```
server.downloadConversions = { "image/webp" : { target : "image/jpeg", compressionLevel : 0.7 }}
 # default: {} ; map input mime type to conversion information, or "default" for others
# server.downloadConversions."image/webp" = {
#   target = "image/jpeg"   # image type to convert to, can also be a url to an external server
#   compressionLevel = 0.8  # quality in range [0,1], leave away to use default compression
# }
server.serveConversions = {}
```

at runtime fails with

```
22:21:03.845 [DefaultDispatcher-worker-16] WARN suwayomi.tachidesk.manga.impl.download.Downloader source(6084907896154116083) - downloadChapter(501 - 1072 | state= Queued, tries= 2, progress= 0.049583334)) -- failed due to
java.util.NoSuchElementException: Key E is missing in the map.
        at kotlin.collections.MapsKt__MapWithDefaultKt.getOrImplicitDefaultNullable(MapWithDefault.kt:24)
        at kotlin.collections.MapsKt__MapsKt.getValue(Maps.kt:370)
        at io.github.config4k.readers.SelectReader.getReader(SelectReader.kt:53)
        at io.github.config4k.readers.ArbitraryTypeReaderKt.extractWithParameters(ArbitraryTypeReader.kt:35)
        at io.github.config4k.readers.ArbitraryTypeReader$1.invoke(ArbitraryTypeReader.kt:16)
        at io.github.config4k.readers.ArbitraryTypeReader$1.invoke(ArbitraryTypeReader.kt:15)
        at io.github.config4k.readers.Reader$getValue$1.invoke(Reader.kt:15)
        at io.github.config4k.readers.Reader$getValue$1.invoke(Reader.kt:14)
        at io.github.config4k.readers.MapReader$1.invoke(MapReader.kt:18)
        at io.github.config4k.readers.MapReader$1.invoke(MapReader.kt:12)
        at io.github.config4k.readers.Reader$getValue$1.invoke(Reader.kt:15)
        at io.github.config4k.readers.Reader$getValue$1.invoke(Reader.kt:14)
        at suwayomi.tachidesk.server.util.MutableStateFlowType.parse(MutableStateFlowType.kt:24)
        at io.github.config4k.readers.SelectReader$getReader$1.invoke(SelectReader.kt:34)
        at io.github.config4k.readers.SelectReader$getReader$1.invoke(SelectReader.kt:34)
        at io.github.config4k.readers.Reader$getValue$1.invoke(Reader.kt:15)
        at io.github.config4k.readers.Reader$getValue$1.invoke(Reader.kt:14)
        at suwayomi.tachidesk.server.ServerConfig.getDownloadConversions(ServerConfig.kt:5425)
        at suwayomi.tachidesk.manga.impl.Page.getPageImageDownload(Page.kt:192)
        at suwayomi.tachidesk.manga.impl.Page$getPageImageDownload$1.invokeSuspend(Page.kt)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:829)
```

That's a config that worked before, and I'm not sure what it's expecting (replacing `:` by `=` like the comment suggests also does not work)

Empty config works as expected

Syer10:
Found the issue, the headers field needed to be an Array instead of a List

cpiber (approved):
Out of scope for this PR, but I think there's a race condition during server startup. When I kill the server while stuff is in the download queue, then on resume I occasionally see some unconverted images in the download, possibly because the downloader starts before the config has finished parsing?

Image

PR #1698, opened 2025-10-04
Use canonical path for static files directory

schroda:
On mac the temp system folder is a symlink which jetty does not allow by default due to security reasons. This caused the webui files to not get served on mac.


Syer10 (approved):

cpiber:
Does this work on the first start? Because on the first start, the directory probably doesn't exist, so `canonicalPath` might not work? I haven't tested it, but the [docs](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/File.html#getCanonicalPath()) say this is implementation defined:

> The canonical form of the pathname of a nonexistent file or directory may be different from the canonical form of the same pathname after the file or directory is created.

schroda:
It works on windows.

Issue #1404, opened 2025-05-23
Enhance graphql manga query


schroda:
### Describe your suggested feature

It should be possible to filter manga for
- [ ] source name/display name
- chapters
  - [ ] downloaded
  - [ ] unread
  - [ ] read
  - [ ]  bookmarked
  - [ ] duplicated
- [ ] has missing chapters
- tracking
  - [ ] is tracked
  - [ ] is tracked by
  
  
and sort by
- chapters
  - [ ] unread
  - [ ] total
  - [ ] recently read
  - [ ] latest fetched
  - [ ] latest uploaded

required to be able to perform the library filterting and sorting completly on the server (https://github.com/Suwayomi/Suwayomi-WebUI/issues/952)

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have updated to the **[latest version](https://github.com/suwayomi/suwayomi-server/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.

Robonau:
some extra stuff VUI does:
- filter
  - [ ] tracker status
- sort
  - [ ] alphabetical Artist
  - [ ] alphabetical Author
  - [ ] random (maybe with seed?)

note: latest fetched is already sort-able its LAST_FETCHED_AT in gql

PR #1719, opened 2025-10-13
Add chapter stats to MangaTable with GraphQL sort/filter support

everpcpc:
## Summary

- Add chaptersLastReadAt, chaptersLastUploadDate, chaptersUnreadCount, chaptersTotalCount to MangaTable
- Support sorting and filtering these fields in GraphQL MangaQuery
- Auto-update stats when chapters are fetched or modified
- Add database migration M0053_AddMangaChapterStatsColumns

## Related

- part of: https://github.com/Suwayomi/Suwayomi-Server/issues/1404


Syer10:
This should be done using joins and other database queries. It shouldn't be done by adding the columns to the manga table directly

everpcpc:
Sorting with join queries can be very complex, especially for operations like count. The query efficiency will be very low and may greatly impact performance.

Robonau:
write both
test the speed difference 
and post findings

if the performance difference is substantial enough syer may change his mind

Syer10:
Some test code I wrote a while ago to provide an example for the join kind of implementation.

https://github.com/Syer10/Tachidesk/commit/a88ebfbe9f24b8fd07003aca41a47740ed98dedd


And for asCursor I won't mind if there are `@GraphqlIgnore` fields in the MangaType to handle it.

PR #1731, opened 2025-10-22
OPDS: Offer CBZ in older mimetype

cpiber:
- Offer CBZ in older mimetype  
  Perhaps a cleaner implementation would be to offer both mimetypes, but that would probably show the button twice for newer clients? I'd assume that all clients support the older mimetype
- Include length when offering CBZ download
- Disable compression on CBZ endpoint

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt
@@ -358,7 +358,7 @@ object OpdsEntryBuilder {
                     "/api/v1/chapter/${chapter.id}/download?markAsRead=${serverConfig.opdsMarkAsReadOnDownload.value}",
                     serverConfig.opdsCbzMimetype.value.mediaType,
                     MR.strings.opds_linktitle_download_cbz.localized(locale),
-                    length = cbzFileSize!!,
+                    length = cbzFileSize,
                 ),
             )
         }
--- a/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt
@@ -356,7 +356,7 @@ object OpdsEntryBuilder {
                 OpdsLinkXml(
                     OpdsConstants.LINK_REL_ACQUISITION_OPEN_ACCESS,
                     "/api/v1/chapter/${chapter.id}/download?markAsRead=${serverConfig.opdsMarkAsReadOnDownload.value}",
-                    OpdsConstants.TYPE_CBZ,
+                    serverConfig.opdsCbzMimetype.value.mediaType,
                     MR.strings.opds_linktitle_download_cbz.localized(locale),
                     length = cbzFileSize!!,
                 ),
--- a/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt
@@ -336,6 +336,14 @@ object OpdsEntryBuilder {
             }
 
         val entryTitle = "$titlePrefix ${chapter.name}"
+        val cbzFileSize =
+            if (chapter.downloaded) {
+                withContext(Dispatchers.IO) {
+                    runCatching { ChapterDownloadHelper.getArchiveStreamWithSize(manga.id, chapter.id).second }.getOrNull()
+                }
+            } else {
+                null
+            }
 
         val links = mutableListOf<OpdsLinkXml>()
         chapter.url?.let {
@@ -350,6 +358,7 @@ object OpdsEntryBuilder {
                     "/api/v1/chapter/${chapter.id}/download?markAsRead=${serverConfig.opdsMarkAsReadOnDownload.value}",
                     OpdsConstants.TYPE_CBZ,
                     MR.strings.opds_linktitle_download_cbz.localized(locale),
+                    length = cbzFileSize!!,
                 ),
             )
         }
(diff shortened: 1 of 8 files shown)


Syer10:
@zeedif could you take a look at this?

zeedif:
@cpiber, thank you for looking into compatibility. I've been reviewing the changes, and it seems there might be a small mix-up between the CBZ and CBR formats.

*   **CBZ** files are ZIP archives, and their MIME type is `application/vnd.comicbook+zip` (modern) or `application/x-cbz` (legacy).
*   **CBR** files are RAR archives, and their MIME type is `application/vnd.comicbook-rar` (modern) or `application/x-cbr` (legacy).

This PR changes the MIME type for CBZ downloads to `application/x-cbr`, which incorrectly tells clients that the file is a RAR archive when it's actually a ZIP. This will likely cause issues for readers that determine the file type based on the mimetype instead of the extension.

I understand the underlying problem is that some older readers only recognize the `x-` types while newer ones only recognize the `vnd.` types. As far as I know, the `x-` types are deprecated, and the ones starting with `vnd.` (like `application/vnd.comicbook+zip`) are the ones officially registered by IANA. You can check the official registry on the IANA site:
- `vnd.comicbook+zip` for CBZ: [https://www.iana.org/assignments/media-types/application/vnd.comicbook+zip](https://www.iana.org/assignments/media-types/application/vnd.comicbook+zip)
- `vnd.comicbook-rar` for CBR: [https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar](https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar)

So, instead of trying to support non-standard configurations, the ideal path forward would be for clients to update to the modern standard. I'll leave the final decision on whether to support this for legacy compatibility to the maintainers.

cpiber:
According to sources I found, `application/x-cbr` is the mimetype for any comic book archive (e.g. https://mime.wcode.net/cbz), while `application/x-cbz` is specifically for cbz. Thus, I used the former for increased compatibility.

And you are right, the x- are deprecated, but some older software refuses downloads since it does not know the new ones.
We may also introduce a setting to allow the user to switch the mimetypes (boolean for modern or compatibility).

zeedif:
A configurable setting for backward compatibility sounds like a good idea.

That said, I’m still skeptical about using `application/x-cbr` for CBZ files. That site doesn't seem to be an official reference like the IANA registry, and technically, `x-cbr` implies a RAR archive, not a ZIP. I can't test this on older devices myself, as my Kobo and Kindle handle the standard MIME types without any issues.

Instead of a boolean for `modern/legacy`, what do you think about using an enum? We could offer choices like:

*   `application/vnd.comicbook+zip`
*   `application/x-cbz`
*   `application/x-cbr`

That said, I have no issues with including the `length` attribute on the acquisition link or disabling HTTP compression.

Even though OPDS 1.2 doesn't explicitly mention the `length` attribute, it inherits it from the Atom specification ([RFC 4287, section 4.2.7.6](https://datatracker.ietf.org/doc/html/rfc4287#section-4.2.7.6)):

> The "length" attribute indicates an advisory length of the linked content in octets; it is a hint about the content length of the representation returned when the IRI in the href attribute is mapped to a URI and dereferenced. Note that the length attribute does not override the actual content length of the representation as reported by the underlying protocol. Link elements MAY have a length attribute.

This could allow some clients to show the file size to the user before starting a download, although that depends on the reader.

Disabling HTTP compression for the CBZ endpoint also makes sense to me. I doubt the GZIP compression layer provides any benefit and likely just wastes CPU, but I'm not sure if it would have any impact on clients. So, it would just be a matter of fixing the MIME type issue.

cpiber:
> Even though OPDS 1.2 doesn't explicitly mention the length attribute, it inherits it from the Atom specification ([RFC 4287, section 4.2.7.6](https://datatracker.ietf.org/doc/html/rfc4287#section-4.2.7.6))

Honstely, I added it because Calibre adds it, and it seemed like it might help.

> Disabling HTTP compression for the CBZ endpoint also makes sense to me. I doubt the GZIP compression layer provides any benefit and likely just wastes CPU, but I'm not sure if it would have any impact on clients.

Also, compression implies a different `Content-Length`, which we manually set to the size of the offered file. I think that's why my reader (KyBook) just closed the connection during download, it probably didn't expect any more data due to an incorrect length header (implying that gzip compression made the file bigger).

zeedif:
If disabling compression fixes the issue, that’s great. Personally, I don’t really take Calibre as a reference since it often does things outside the EPUB standard (like using non-standard metadata attributes instead of documented alternatives for series, series index, rating, etc.) or OPDS (for example, paginating feeds with OpenSearch instead of using feed-level links: first, last, next, and previous). But at least the *length* part is implicitly covered in the documentation.

Syer10:
I'm confused, does `KyBook` not recognize `LEGACY`(`application/x-cbz`)? Why are we offering the Cbz as both a zip and rar even though its a zip.

cpiber:
It does, I just didn't learn of that mimetype initially. According to several sources, `application/x-cbr` is valid for any comic book archive (.cbz, .cbr, .cba, .cbt, .cb7), so I've left it in case there are other readers that don't like `application/x-cbz`

Syer10 (approved):

Syer10:
I see. I guess we can leave it then.

server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt
```diff
@@ -348,8 +356,9 @@ object OpdsEntryBuilder {
                 OpdsLinkXml(
                     OpdsConstants.LINK_REL_ACQUISITION_OPEN_ACCESS,
                     "/api/v1/chapter/${chapter.id}/download?markAsRead=${serverConfig.opdsMarkAsReadOnDownload.value}",
-                    OpdsConstants.TYPE_CBZ,
+                    serverConfig.opdsCbzMimetype.value.mediaType,
                     MR.strings.opds_linktitle_download_cbz.localized(locale),
+                    length = cbzFileSize!!,
```
Syer10: We probably don't need to enforce non-null here.
```suggestion
                    length = cbzFileSize,
```

PR #1736, opened 2025-10-24
Add --change-stack-guard-on-fork=disable to fix stack smashing

cpiber:
This should fix frequent stack smashing messages. In my testing, these occured when 2 or more instances of CEF browser were open at the same time. Most easily reproducible by opening the webview endpoint and using an extension that requires webview.

See chromiumembedded/cef#3912

I would argue that this is a bug in JCEF, since they are missing the `NO_STACK_PROTECTOR` that CEF requires, but it might also be related to our usage.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
@@ -527,7 +527,13 @@ fun applicationSetup() {
                 }
                 appHandler(
                     KCEF.AppHandler(
-                        arrayOf("--disable-gpu", "--off-screen-rendering-enabled", "--disable-dev-shm-usage", "--enable-widevine-cdm"),
+                        arrayOf(
+                            "--disable-gpu",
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -527,7 +527,13 @@ fun applicationSetup() {
...
                 appHandler(
                     KCEF.AppHandler(
-                        arrayOf("--disable-gpu", "--off-screen-rendering-enabled", "--disable-dev-shm-usage", "--enable-widevine-cdm"),
+                        arrayOf(
+                            "--disable-gpu",
+                            "--off-screen-rendering-enabled",
+                            "--disable-dev-shm-usage",
+                            "--enable-widevine-cdm",
+                            "--change-stack-guard-on-fork=disable",
+                        ),
```
Syer10, line 540: Maybe add comments referencing any issues so that we know why they were added.

Syer10 (approved):

Issue #1596, opened 2025-08-18
Suwayomi sometimes doesn't gracefully exit on PC shutdown


Akonyl:
### Steps to reproduce

1. Run Suwayomi
2. Update manga, read, etc, typical usage (unknown what exactly causes the issue)
3. Close all open windows (including browser which the WebUI is running from)
4. Leave Suwayomi server running and shutdown the PC

### Expected behavior

The PC should shut down gracefully without throwing exceptions or requiring user input.

### Actual behavior

 Occasionally (~50-75% chance) the shutdown process will halt, with a java exception being thrown from Suwayomi's javaw.exe.

### Suwayomi-Server version

2.1.1867

### Server operating system

Windows 10

### Server JVM version

21.0.8.0

### Used client name

WebUI

### Client version

r2717

### Used web browser

Firefox 141

### Client operating system

Windows 10

### Other details

Obviously a minor bug, but I'm writing it up because on the previous version I used (2.0.1727, used since May), it didn't happen. Could also be related to heavier usage now, but I immediately started seeing it upon switching to the new version.

I assume this isn't very common as if I was running Suwayomi server on an actual server, rather than the same desktop as the webui reader, I wouldn't be shutting it off.

Sorry for the phone pic, but I can't really take a screenshot of it because of the circumstance.

<img width="517" height="474" alt="Image" src="https://github.com/user-attachments/assets/f542212b-3f8e-44e0-8720-c0b8a95688ea" />

> The exception Breakpoint
> A breakpoint has been reached.
> (0x80000003) occurred in the application at location
> 0x00007FF84CE83B07.
> 
> Click on OK to terminate the program
> Click on CANCEL to debug the program

Clicking both OK and Cancel on the window both do the same thing (just closing the window and letting the PC shut down normally), Cancel doesn't give me any more useful debug information. 

Digging into it, I found "debug.log" in the bin folder, which I have attached. It seems like each time, the output looks something like:

[debug.log](https://github.com/user-attachments/files/21845155/debug.log)

[0818/030125.185:ERROR:network_service_instance_impl.cc(599)] Network service crashed, restarting service.
[0818/030125.505:ERROR:gpu_process_host.cc(989)] GPU process exited unexpectedly: exit_code=1073807364
[0818/030125.505:WARNING:gpu_process_host.cc(1360)] The GPU process has crashed 1 time(s)
[...]

PR #1738, opened 2025-10-24
[#1596] Dispose KCEF on shutdown

cpiber:
Properly dispose KCEF on shutdown. Tried to reboot ~5 times and didn't see any errors, but I also could not reproduce the original issue.

@yeezylife, @JAK1047 please test if this solves the issue or if something else is responsible

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
@@ -32,7 +32,6 @@ import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import org.bouncycastle.jce.provider.BouncyCastleProvider
 import org.cef.network.CefCookieManager
 import org.koin.core.context.startKoin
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -538,4 +540,15 @@ fun applicationSetup() {
             onError = { it?.printStackTrace() },
         )
     }
+
+    Runtime.getRuntime().addShutdownHook(
+        thread(start = false) {
+            runBlocking {
+                val logger = KotlinLogging.logger("KCEF")
+                logger.debug { "Shutting down KCEF" }
+                KCEF.dispose()
```
Syer10: Is there a disposeBlocking version of this function that we can use instead of wrapping this in runBlocking?

cpiber: There is, but it just wraps the call in `runBlocking` itself:

https://github.com/DatL4g/KCEF/blob/0665269b7d6a91b0ee187f4432bb5be5ca41a112/kcef/src/main/kotlin/dev/datlag/kcef/KCEF.kt#L294-L297

Syer10: Let's use that. Less code on our end.

Syer10 (approved):

PR #1757, opened 2025-10-31
Optionally allow basic auth fallback for chapter page endpoint

schroda:
Requesting a chapter page while having a non basic auth mode (e.g. ui login) enabled caused the basic auth prompt.
However, this is only supposed to be happen for opds


Syer10:
Add the parameter here https://github.com/Suwayomi/Suwayomi-Server/blob/c9575449da69211ea3511827c37197e09d968694/server/src/main/kotlin/suwayomi/tachidesk/opds/impl/OpdsEntryBuilder.kt#L367-L368

Syer10 (approved):

PR #1758, opened 2025-11-01
Fix/remove koreader-sync credentials from server config

schroda:


Syer10:
We need a migration in https://github.com/Suwayomi/Suwayomi-Server/blob/master/server/src/main/kotlin/suwayomi/tachidesk/server/Migration.kt for this I think

Syer10 (approved):

PR #1759, opened 2025-11-01
Prevent running webui auto update in parallel to the setup

schroda:


server/src/main/kotlin/suwayomi/tachidesk/server/util/WebInterfaceManager.kt
```diff
@@ -239,6 +242,10 @@ object WebInterfaceManager {
     private fun scheduleWebUIUpdateCheck() {
         HAScheduler.descheduleCron(currentUpdateTaskId)
 
+        if (!isSetupComplete) {
+            return
+        }
+
```
Syer10: Wouldn't this mean that it never happens.

Would it be better to add a lock instead? Like

```kotlin
val setupLock = Mutex(locked = true)

// in scheduling function, suspends until unlocked
setupLock.withLock {}

// after setup is finished
setupLock.unlock()
```

schroda: The setup will update to the latest version. So in case it's running, the automated update should just be skipped

schroda: Ok, I just noticed, I added the guard at the wrong place

Syer10: The scheduler for the next job happens at the end of this function, so this means that will also be skipped won't it?

Syer10 (approved):

PR #1784, opened 2025-11-12
Add IReader extension support to Suwayomi-Server

kazemcodes:
- Integrated IReader source-api (38 files) for novel/light novel reading
- Created separate database tables for IReader extensions and sources
- Implemented extension management (install, update, uninstall)
- Added REST API endpoints under /api/v1/ireader/
- Full parallel operation with existing Tachiyomi functionality
- Zero breaking changes to existing code
- Comprehensive documentation included


Syer10:
Interesting, your are the dev of IReader? I'm not against novel support for Suwayomi.

A few notes to start with.
- We do not support the REST api anymore, endpoints should be added under the GraphQL api in the GraphQL package
- I saw quite a bit of duplicate code, especially with the extensions installer. There should be a way to deduplicate this code so that functions can be shared between Manga and Novel extension installers.
- I saw a few expect functions and classes. Since we are a single platform program, there should be no expect/actual, just implementations.

kazemcodes:
yes, I'm dev of IReader, I will try to fix the issue you mentioned,
by the way thanks for creating base implementation

kazemcodes:
I complete all changes you mentioned

> Interesting, your are the dev of IReader? I'm not against novel support for Suwayomi.
> 
> A few notes to start with.
> 
> * We do not support the REST api anymore, endpoints should be added under the GraphQL api in the GraphQL package
> * I saw quite a bit of duplicate code, especially with the extensions installer. There should be a way to deduplicate this code so that functions can be shared between Manga and Novel extension installers.
> * I saw a few expect functions and classes. Since we are a single platform program, there should be no expect/actual, just implementations.

Syer10:
Its a start. I will do a more in-depth review when I have time.

The GraphQL parts are not complete but I can also tell you what's needed later.

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderExtensionMutation.kt
```diff
@@ -0,0 +1,217 @@
...
+
+    data class UpdateAllIReaderExtensionsPayload(
+        val clientMutationId: String?,
+        @GraphQLDescription("Number of extensions queued for update")
+        val updatedCount: Int,
+    )
+
+    @RequireAuth
+    @GraphQLDescription("Update all IReader extensions that have updates available")
+    fun updateAllIReaderExtensions(
```
Syer10: This function is not needed, the client can loop over UpdateExtension themselves

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderExtensionQuery.kt
```diff
@@ -0,0 +1,165 @@
...
+import suwayomi.tachidesk.graphql.directives.RequireAuth
+import suwayomi.tachidesk.graphql.types.IReaderExtensionType
+import suwayomi.tachidesk.graphql.types.IReaderSourceType
+import suwayomi.tachidesk.manga.impl.IReaderSource
+import suwayomi.tachidesk.manga.impl.extension.ireader.IReaderExtension
+import suwayomi.tachidesk.manga.impl.extension.ireader.IReaderExtensionsList
+import suwayomi.tachidesk.server.JavalinSetup.future
+import java.util.concurrent.CompletableFuture
+
+class IReaderExtensionQuery {
```
Syer10, line 18: This implementation is wrong, it should mimic the ExtensionQuery and ExtensionMutation classes.

Any query that changes the database needs to be a mutation, and (almost) all database tables should have a dedicated Query class.

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderNovelQuery.kt
```diff
@@ -0,0 +1,115 @@
...
+import suwayomi.tachidesk.graphql.directives.RequireAuth
+import suwayomi.tachidesk.graphql.types.IReaderChapterType
+import suwayomi.tachidesk.graphql.types.IReaderNovelType
+import suwayomi.tachidesk.graphql.types.IReaderNovelsPageType
+import suwayomi.tachidesk.graphql.types.IReaderPageType
+import suwayomi.tachidesk.manga.impl.IReaderNovel
+import suwayomi.tachidesk.server.JavalinSetup.future
+import java.util.concurrent.CompletableFuture
+
+class IReaderNovelQuery {
```
Syer10, line 46: This is a group of mutations that should be split up into multiple several files, try to mimic the Manga side when creating these variants.

Syer10 (changes requested):
Initial round of review

kazemcodes:
I made some changes, when you had time check them out

server/src/main/kotlin/ireader/core/http/DefaultUserAgent.kt
```diff
@@ -0,0 +1,5 @@
+package ireader.core.http
+
+const val DEFAULT_USER_AGENT =
```
Syer10: Is this shared to the extension API? If not there is no need to use a contant for the UA since we can use flaresolverr to get a updated one. See NetworkHelper.userAgent

server/src/main/kotlin/ireader/core/prefs/PreferenceStore.kt
```diff
@@ -0,0 +1,220 @@
...
+import java.util.prefs.Preferences
+
+/**
+ * Implementation of IReader's PreferenceStore
+ * Must be in ireader.core.prefs package to match extension expectations
+ */
+class PreferenceStoreImpl(
+    private val packageName: String,
+) : PreferenceStore {
+    private val prefs = Preferences.userRoot().node("suwayomi/ireader/$packageName")
```
Syer10: Instead of storing in the JVM Preferences lets use our implementation of SharedPreferences in AndroidCompat. For example
```suggestion
    private val prefs = Injekt.get<Application>().getSharedPreferences("suwayomi/ireader/$packageName", Context.MODE_PRIVATE)
```

server/src/main/kotlin/ireader/core/source/model/PageListEmpty.kt
```diff
@@ -0,0 +1,5 @@
+
+
+package ireader.core.source.model
```
Syer10:
```suggestion
package ireader.core.source.model
```

server/src/main/kotlin/org/ireader/coreapi/prefs/PreferenceStore.kt
```diff
@@ -0,0 +1,208 @@
...
+import kotlinx.serialization.modules.SerializersModule
+import java.util.prefs.Preferences
+
+/**
+ * Simple implementation of IReader's PreferenceStore using Java Preferences API
+ */
+class PreferenceStoreImpl(
+    private val packageName: String,
+) : PreferenceStore {
+    private val prefs = Preferences.userRoot().node("suwayomi/ireader/$packageName")
```
Syer10: Duplicate implementation?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderChapterMutation.kt
```diff
@@ -0,0 +1,88 @@
...
+import suwayomi.tachidesk.server.JavalinSetup.future
+import java.util.concurrent.CompletableFuture
+
+class IReaderChapterMutation {
+    data class FetchIReaderChaptersInput(
+        val clientMutationId: String? = null,
+        @GraphQLDescription("Source ID")
+        val sourceId: Long,
+        @GraphQLDescription("Novel URL from the source")
+        val novelUrl: String,
```
Syer10: Why are we using a novelUrl instead of a database id?

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderChapterMutation.kt
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+    }
+
+    data class FetchIReaderChapterContentInput(
+        val clientMutationId: String? = null,
+        @GraphQLDescription("Source ID")
+        val sourceId: Long,
+        @GraphQLDescription("Chapter URL from the source")
+        val chapterUrl: String,
```
Syer10: Why are we using a chapter url instead of a database id?

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderNovelQuery.kt
Syer10: This is not mimicing the MangaQuery class.

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderSourceQuery.kt
```diff
@@ -0,0 +1,75 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+package suwayomi.tachidesk.graphql.queries
+
+import com.expediagroup.graphql.generator.annotations.GraphQLDescription
+import suwayomi.tachidesk.graphql.directives.RequireAuth
+import suwayomi.tachidesk.graphql.types.IReaderSourceType
+import suwayomi.tachidesk.manga.impl.IReaderSource
+
+class IReaderSourceQuery {
```
Syer10, line 46: This is not mimicing the SourceQuery class

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/IReaderNovelTypes.kt
```diff
@@ -0,0 +1,79 @@
...
+
+import com.expediagroup.graphql.generator.annotations.GraphQLDescription
+import ireader.core.source.model.ChapterInfo
+import ireader.core.source.model.MangaInfo
+import ireader.core.source.model.MangasPageInfo
+import ireader.core.source.model.Page
+import ireader.core.source.model.Text
+
+@GraphQLDescription("Novel information from IReader source")
+data class IReaderNovelType(
```
Syer10: We need data loaders to get related types. Take a look at the `MangaType.chapters()` function and the `ChaptersForMangaDataLoader` class.

This needs to be done for every type that has relations. These are some of them that are at the top of my head
Extensions -> Source
Source -> Extension
Source -> Manga
Manga -> Source
Manga -> Chapter
Chapter -> Manga

Syer10 (changes requested):
Round 2

KolbyML:
@kazemcodes this is cool, hope you continue

Syer10:
Looking at the recent options, I think you got hit by the dex2jar update. You can probably remove a lot of that code by adding this to the dex2jar runner.
https://github.com/Suwayomi/Suwayomi-Server/blob/817589f71081cfdccd4031799225aa0807546f47/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/PackageTools.kt#L70

Unless you are referencing another dex2jar issue.

kazemcodes:
> Looking at the recent options, I think you got hit by the dex2jar update. You can probably remove a lot of that code by adding this to the dex2jar runner.
> 
> https://github.com/Suwayomi/Suwayomi-Server/blob/817589f71081cfdccd4031799225aa0807546f47/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/PackageTools.kt#L70
> 
> Unless you are referencing another dex2jar issue.

there is another problem with dex2jar 
this is an answer from claude opus 4.5 model
"The .dontSanitizeNames(true) is already present. The stackmap frame issue is a fundamental problem with how dex2jar converts Dalvik bytecode to JVM bytecode - it doesn't always generate correct stackmap frames for complex control flow (especially Kotlin coroutines)."

kazemcodes:
I think the issue is with coroutines current versions, there seems to be some problem with it, not sure, it works fine before with any issue without flag, but after I updated the my source-api module in order to create even better sources, this bug happens here ,
I decided to recode the source in something better than orginal tachi sources which handles cloudflare, ratelimit, and other stuff better

Syer10:
Try making sure the extensions don't include any dependencies in its apks. Use the Android Studio APK explorer. For example when I implemented Apollo in the Suwayomi extension it pulled in other dependencies that conflicted with the apps dependencies.
https://github.com/Suwayomi/tachiyomi-extension/blob/769c2d3b2412fc3ddb50d5a479f418215dedfd6d/src/all/tachidesk/build.gradle#L30

Otherwise that verify error sounds familiar, but I forget where.

kazemcodes:
> Try making sure the extensions don't include any dependencies in its apks. Use the Android Studio APK explorer. For example when I implemented Apollo in the Suwayomi extension it pulled in other dependencies that conflicted with the apps dependencies.
> https://github.com/Suwayomi/tachiyomi-extension/blob/769c2d3b2412fc3ddb50d5a479f418215dedfd6d/src/all/tachidesk/build.gradle#L30
> 
> Otherwise that verify error sounds familiar, but I forget where.

I will try to find a solution for this problem

Syer10:
I think I remember where I encountered that error before.
Keiyoushi once tested enabling proguard for the extensions but it caused the verify exception that you are talking about in Suwayomi, so it was decided not to do it and leave proguard off.

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderNovelQuery.kt
kazemcodes: can you check this file again, does it now meets you expectation now?

kazemcodes:
> I think I remember where I encountered that error before. Keiyoushi once tested enabling proguard for the extensions but it caused the verify exception that you are talking about in Suwayomi, so it was decided not to do it and leave proguard off.

I tried multiple ways to fix this issue, and after trying every thing,I decided to not use dex2jar at all, directly using .jar files that main ireader-extensions repo create ,
the problem wasn't even in proguard rule as extensions is not using proguard rules at all, jar are faster to use in jvm

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderChapterMutation.kt
kazemcodes: The current design uses chapterUrl because this mutation is intended for source-level browsing - fetching chapter content directly from IReader sources without requiring the chapter to be persisted in the database first. This allows users to preview content before adding novels to their library.

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderChapterMutation.kt
kazemcodes: The current design uses novelUrl because this mutation is intended for source-level browsing - fetching novel content directly from IReader sources without requiring the novel to be persisted in the database first. This allows users to preview content before adding novels to their library.

kazemcodes:
can you recheck the new file changes, I almost resolved all things you mentioned

local.properties
Syer10, line 1: This file shouldn't be added to the repo.

gradle/libs.versions.toml
```diff
@@ -165,7 +169,23 @@ ktlint = { module = "com.pinterest.ktlint:ktlint-cli", version.ref = "ktlint" }
...
+ktor-core = { module = "io.ktor:ktor-client-core", version.ref = "ktor" }
+ktor-core-android = { module = "io.ktor:ktor-client-android", version.ref = "ktor" }
+ktor-core-cio = { module = "io.ktor:ktor-client-cio", version.ref = "ktor" }
+ktor-okhttp = { module = "io.ktor:ktor-client-okhttp", version.ref = "ktor" }
+ktor-client-js = { module = "io.ktor:ktor-client-js", version.ref = "ktor" }
+ktor-contentNegotiation = { module = "io.ktor:ktor-client-content-negotiation", version.ref = "ktor" }
+ktor-contentNegotiation-gson = { module = "io.ktor:ktor-serialization-gson", version.ref = "ktor" }
+ktor-contentNegotiation-kotlinx = { module = "io.ktor:ktor-serialization-kotlinx-json", version.ref = "ktor" }
+ktor-contentNegotiation-jackson = { module = "io.ktor:ktor-serialization-jackson", version = "3.1.2" }
+stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib-common", version.ref = "kotlin" }
```
Syer10: No need to add the std lib

gradle/libs.versions.toml
```diff
@@ -165,7 +169,23 @@ ktlint = { module = "com.pinterest.ktlint:ktlint-cli", version.ref = "ktlint" }
 
 # moko
 moko = { module = "dev.icerock.moko:resources", version.ref = "moko" }
-
+# Ksoup - Kotlin Multiplatform HTML parser (works on iOS, Android, Desktop)
+ksoup = { module = "com.fleeksoft.ksoup:ksoup", version.ref = "ksoup" }
+ksoup-network = { module = "com.fleeksoft.ksoup:ksoup-network", version.ref = "ksoup" }
+androidx-datastore-core = { module = "androidx.datastore:datastore-core", version.ref = "datastore" }
+androidx-datastore-preferences-core = { module = "androidx.datastore:datastore-preferences-core", version.ref = "datastore" }
```
Syer10: Do we need datastore?

server/src/main/kotlin/ireader/core/http/CookieManager.kt
```diff
@@ -0,0 +1,99 @@
...
+import okhttp3.CookieJar
+import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+
+/**
+ * A cookie manager for saving cookies from network responses, allowing to send them in succesive
+ * requests. There's an in-memory copy of the cookies through [cache] and a persistent copy
+ * through [store] so that they are restored when the application is restarted.
+ */
+class CookieManager(private val store: CookieStore) : CookieJar {
```
Syer10: This should be using the NetworkHelper cookie store so we have a centralized place for cookies that's shared with Webview and Extensions

server/src/main/kotlin/ireader/core/http/CookieSynchronizer.kt
```diff
@@ -0,0 +1,19 @@
+package ireader.core.http
+
+/**
+ * Desktop implementation of cookie synchronization
+ * Desktop doesn't have WebView, so this is a no-op implementation
+ */
+ class CookieSynchronizer {
```
Syer10: No need for this? Our Webview already shares cookies with extensions.

server/src/main/kotlin/ireader/core/http/OkHttpExtension.kt
```diff
@@ -0,0 +1,11 @@
+package ireader.core.http
+
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.okhttp.OkHttpEngine
+import okhttp3.OkHttpClient
+
+/**
+ * Extension to get OkHttpClient from Ktor HttpClient on Desktop.
+ */
+val HttpClient.okhttp: OkHttpClient
```
Syer10, line 10: Do we use CIO or OkHttp for Ktor? And how is this used?

server/src/main/kotlin/ireader/core/http/WebViewManger.kt
```diff
@@ -0,0 +1,37 @@
+package ireader.core.http
+
+import com.fleeksoft.ksoup.nodes.Document
+
+/**
+ * Desktop implementation of WebViewManager
+ * Currently a stub - could be enhanced with JavaFX WebView or JCEF
+ */
+ class WebViewManger {
```
Syer10, line 9: We have a Webview implementation, stubbing the standard Android Webview classes. So just copy the android implementation?

gradle/libs.versions.toml
```diff
@@ -165,7 +169,23 @@ ktlint = { module = "com.pinterest.ktlint:ktlint-cli", version.ref = "ktlint" }
...
+ktor-core-android = { module = "io.ktor:ktor-client-android", version.ref = "ktor" }
+ktor-core-cio = { module = "io.ktor:ktor-client-cio", version.ref = "ktor" }
+ktor-okhttp = { module = "io.ktor:ktor-client-okhttp", version.ref = "ktor" }
+ktor-client-js = { module = "io.ktor:ktor-client-js", version.ref = "ktor" }
+ktor-contentNegotiation = { module = "io.ktor:ktor-client-content-negotiation", version.ref = "ktor" }
+ktor-contentNegotiation-gson = { module = "io.ktor:ktor-serialization-gson", version.ref = "ktor" }
+ktor-contentNegotiation-kotlinx = { module = "io.ktor:ktor-serialization-kotlinx-json", version.ref = "ktor" }
+ktor-contentNegotiation-jackson = { module = "io.ktor:ktor-serialization-jackson", version = "3.1.2" }
+stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib-common", version.ref = "kotlin" }
+kermit = "co.touchlab:kermit:2.0.8"
```
Syer10, line 184: I see the `ireader.core.log` package, do we need kermit for extensions?

server/src/main/kotlin/ireader/core/source/DeepLinkSource.kt
```diff
@@ -0,0 +1,11 @@
+
+
+package ireader.core.source
```
Syer10, line 3:
```suggestion
package ireader.core.source
```

server/src/main/kotlin/ireader/core/source/Dependencies.kt
```diff
@@ -0,0 +1,11 @@
+
+
+package ireader.core.source
```
Syer10, line 3:
```suggestion
package ireader.core.source
```

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderSourceMutation.kt
```diff
@@ -0,0 +1,186 @@
...
+            }
+        }
+    }
+
+    data class FetchIReaderChaptersInput(
+        val clientMutationId: String? = null,
+        @GraphQLDescription("Source ID (required if using novelUrl)")
+        val source: Long? = null,
+        @GraphQLDescription("Novel URL/key from the source (provide this OR novelId)")
+        val novelUrl: String? = null,
```
Syer10, line 179: This and the sourceId should be removed, at this point we should be relying only on the novelId. Use that to get the sourceId.

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderSourceMutation.kt
```diff
@@ -0,0 +1,186 @@
...
+    )
+
+    data class FetchIReaderChaptersPayload(
+        val clientMutationId: String?,
+        val chapters: List<IReaderChapterType>,
+    )
+
+    @RequireAuth
+    @GraphQLDescription("Fetch chapters for a novel from an IReader source")
+    fun fetchIReaderChapters(input: FetchIReaderChaptersInput): CompletableFuture<DataFetcherResult<FetchIReaderChaptersPayload?>> {
```
Syer10, line 191: This should be moved to a `IReaderChapterMutation` class

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderSourceMutation.kt
```diff
@@ -0,0 +1,186 @@
...
+            }
+        }
+    }
+
+    data class FetchIReaderChapterContentInput(
+        val clientMutationId: String? = null,
+        @GraphQLDescription("Source ID (required if using chapterUrl)")
+        val source: Long? = null,
+        @GraphQLDescription("Chapter URL/key from the source (provide this OR chapterId)")
+        val chapterUrl: String? = null,
```
Syer10, line 335: This and the sourceId should be removed, at this point we should be relying only on the chapterId. Use that to get the sourceId.

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderSourceMutation.kt
```diff
@@ -0,0 +1,186 @@
...
+    )
+
+    data class FetchIReaderChapterContentPayload(
+        val clientMutationId: String?,
+        val pages: List<IReaderPageType>,
+    )
+
+    @RequireAuth
+    @GraphQLDescription("Fetch content/pages for a chapter from an IReader source")
+    fun fetchIReaderChapterContent(input: FetchIReaderChapterContentInput): CompletableFuture<DataFetcherResult<FetchIReaderChapterContentPayload?>> {
```
Syer10, line 347: This should be moved to a `IReaderChapterMutation` class

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderExtensionQuery.kt
Syer10: This  class still deviates quite a bit from the ExtensionQuery class. It doesn't implement proper sorting, filtering, and adds a bunch of useless queries. Change it to mimic the ExtensionQuery class

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderSourcePreferencesQuery.kt
```diff
@@ -0,0 +1,46 @@
...
+data class IReaderSourcePreference(
+    val key: String,
+    val title: String,
+    val summary: String?,
+    val defaultValue: String?,
+    val currentValue: String?,
+    val valueType: String,
+)
+
+class IReaderSourcePreferencesQuery {
```
Syer10, line 47: This should just be a DataLoader value on the Source class, not its own query

server/src/main/kotlin/suwayomi/tachidesk/graphql/subscriptions/IReaderExtensionSubscription.kt
```diff
@@ -0,0 +1,43 @@
...
+object IReaderExtensionStatusManager {
+    private val _statusFlow = MutableSharedFlow<IReaderExtensionStatus>(replay = 1)
+    val statusFlow = _statusFlow.asSharedFlow()
+
+    suspend fun emitStatus(status: IReaderExtensionStatus) {
+        _statusFlow.emit(status)
+    }
+}
+
+class IReaderExtensionSubscription {
```
Syer10, line 39: Is there a need for this? I don't think it takes that long to install a extension in Suwayomi.

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/IReaderNovelTypes.kt
```diff
@@ -0,0 +1,167 @@
...
+        genre = mangaInfo.genres,
+        status = mangaInfo.status,
+        inLibrary = false,
+        inLibraryAt = 0,
+        lastFetchedAt = null,
+        chaptersLastFetchedAt = null,
+    )
+
+    fun source(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<IReaderSourceType?> =
+        dataFetchingEnvironment.getValueFromDataLoader<Long, IReaderSourceType?>("IReaderSourceDataLoader", sourceId)
```
Syer10, line 84: This needs a chapters data loader

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderSourceMutation.kt
```diff
@@ -0,0 +1,186 @@
...
+                            IReaderNovel.getPopularNovels(sourceId, page)
+                        }
+                        FetchIReaderNovelType.LATEST -> {
+                            IReaderNovel.getLatestNovels(sourceId, page)
+                        }
+                    }
+
+                FetchIReaderNovelsPayload(
+                    clientMutationId = clientMutationId,
+                    novels = novelsPage.mangas.map { IReaderNovelType(it, sourceId) },
```
Syer10: These novels need to be in the database at this point. We cannot serve data that is not saved. They also need database ids so that we can use them later.

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/IReaderSourceMutation.kt
```diff
@@ -0,0 +1,186 @@
...
+                            novel[IReaderNovelTable.sourceReference] to novel[IReaderNovelTable.url]
+                        }
+                        novelUrl != null && source != null -> {
+                            require(novelUrl.isNotBlank()) { "Novel URL cannot be empty" }
+                            source to novelUrl
+                        }
+                        else -> throw IllegalArgumentException("Either novelId OR (novelUrl + source) must be provided")
+                    }
+
+                val chapters = IReaderNovel.getChapterList(resolvedSourceId, resolvedNovelUrl)
```
Syer10: These chapters need to be in the database at this point. We cannot serve data that is not saved. They also need database ids so that we can use them later.

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/IReaderNovelQuery.kt
Syer10: Much better

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/IReaderNovelTypes.kt
```diff
@@ -0,0 +1,167 @@
...
+    val scanlator: String,
+) {
+    constructor(chapterInfo: ChapterInfo) : this(
+        name = chapterInfo.name,
+        key = chapterInfo.key,
+        dateUpload = chapterInfo.dateUpload,
+        number = chapterInfo.number,
+        scanlator = chapterInfo.scanlator,
+    )
+}
```
Syer10, line 176: Chapters should be in the database

server/src/main/kotlin/suwayomi/tachidesk/manga/MangaAPI.kt
```diff
@@ -37,6 +40,34 @@ object MangaAPI {
...
+                get("{sourceId}/popular/{page}", IReaderNovelController.popular)
+                get("{sourceId}/latest/{page}", IReaderNovelController.latest)
+                get("{sourceId}/search", IReaderNovelController.search)
+
+                // Novel details and chapters
+                get("{sourceId}/details", IReaderNovelController.details)
+                get("{sourceId}/chapters", IReaderNovelController.chapters)
+                get("{sourceId}/chapter", IReaderNovelController.chapterContent)
+            }
+        }
```
Syer10: Honestly I think we should remove all these. Adding new endpoints to the deprecated REST API isn't a good idea. It just makes the actual IReader implementation confusing.

Syer10 (changes requested):

gradle/libs.versions.toml
kazemcodes: No

kazemcodes: I will remove it

kazemcodes:
Thanks for review, I will fix the issue you mentioned later

kazemcodes:
Any more issues?

Syer10:
There is still a bunch of comments that need to be addressed from my previous review

sohilsayed:
Really looking forward for this 
Keep up the good work @kazemcodes

1Selxo:
yo this is hype really waiting for LN support and maybe anime if possible it'd be the ultimate all in 1 app

Syer10:
I've been looking deeper at this PR and at this moment I don't think I can merge this. The IReader Source API is just too much as it tries to be all encompassing.

If it was mostly interfaces that could be implemented by the app its implemented in that would be better. But as it is its just too much as it itself tries to do all the implementations.

The Source API needs to slim down before I would consider merging this again. Open a new PR if it ever gets to that point.

kazemcodes:
which part you didnt liked the most?

PR #1794, opened 2025-11-24
Feature/improve server config non privacy safe setting handling

schroda:


AndroidCompat/Config/src/main/java/xyz/nulldev/ts/config/ConfigManager.kt
```diff
@@ -178,6 +178,23 @@ open class ConfigManager {
         userConfigFile.writeText(newUserConfigDoc.render())
         getUserConfig().entrySet().forEach { internalConfig = internalConfig.withValue(it.key, it.value) }
     }
+
+    fun getRedactedConfig(nonPrivacySafeKeys: List<String>): Config {
+        val entries =
+            config.entrySet().associate { entry ->
+                val key = entry.key
+                val value =
+                    if (nonPrivacySafeKeys.contains(key.split(".")[1])) {
```
Syer10: Would this work? It should prevent potential issues.
```suggestion
                    if (nonPrivacySafeKeys.any { key.split(".").getOrNull(1) == it }) {
```

Syer10 (approved):

PR #1869, opened 2026-01-23
Prevent subpath injection with disabled setting

schroda:


renjfk:
This seems to have broken the subpath feature completely.

schroda:
> This seems to have broken the subpath feature completely.

What does not work? It didn't change anything in case you're using a subpath.

renjfk:
My bad, it was due to misconfiguration. It was actually working by sheer luck but now that you have removed the subpath block I had opportunity to correct it. 😌

PR #1879, opened 2026-01-27
feat(opds): Add option to skip metadata feed for direct access

zeedif:
This PR adds a configuration option, `server.opdsSkipChapterMetadataFeed`, to allow the chapter list to handle everything the metadata feed used to do. Instead of having to tap into each chapter to find the download or stream links, they are now available directly in the main list. This is a big quality-of-life improvement for KOReader users or anyone using automated downloaders.

### 1. Optimization and Logic Cleanup
- **`ChapterForDownload.kt`**: I've extracted some of the internal logic (`updateChapterPersistence` and `refreshChapterPageList`) to make it reusable. This allows the server to verify local files and update database page counts in bulk much more efficiently, without relying on the full download-ready flow.

### 2. Making the Chapter List "Smarter"
- **`ChapterRepository.kt`**: When the "skip metadata" flag is on, the repository now enriches the chapter list asynchronously. It checks the download status, fixes DB mismatches on the fly, and calculates the **CBZ file size**. This way, the list has all the info it needs (like the "Downloaded" status and size) without needing a sub-feed.
- **`OpdsEntryBuilder.kt`**: I moved the logic that generates download (CBZ) and stream (OPDS-PSE) links so it can be injected directly into the chapter list entries. I also moved the visual state indicators (like the "Downloaded" icon) to the list titles.

### 3. KoSync Integration
- I’ve integrated **KoSync** directly into the chapter list. It now checks for remote progress while generating the list.
- Since we lose the "Metadata" screen where a user could manually resolve a conflict, I’ve made it non-blocking: it applies updates if the strategy says so, but if it's set to `PROMPT`, it just stays quiet and uses the local progress to keep the feed moving.

### 4. Configuration and DTOs
- Added the new setting to `ServerConfig.kt` (ProtoNumber 86).
- Updated the DTOs (`OpdsChapterListAcqEntry` and `OpdsChapterMetadataAcqEntry`) to carry the `downloaded` status and `cbzFileSize` info.
- Added a note in `Configuring-Suwayomi-Server.md` about how this affects syncing.

## Why do this?
It saves a lot of navigation. Instead of *Chapter List -> Metadata Feed -> Download*, it’s just *Chapter List -> Download*. It makes the whole OPDS integration feel more responsive and works much better with tools that try to index or bulk-download chapters.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
@@ -71,13 +71,46 @@ object ChapterDownloadHelper {
                     .select(ChapterTable.columns + MangaTable.columns)
                     .where { ChapterTable.id eq chapterId }
                     .firstOrNull() ?: throw IllegalArgumentException("ChapterId $chapterId not found")
+
             val chapter = ChapterTable.toDataClass(row)
-            val mangaTitle = row[MangaTable.title]
+            val mangaTitle = row[MangaTable.title].trim()
 
-            val scanlatorPart = chapter.scanlator?.let { "[$it] " } ?: ""
-            val fileName = "$mangaTitle - $scanlatorPart${chapter.name}.cbz"
+            val scanlatorName = chapter.scanlator?.trim()?.takeIf { it.isNotEmpty() }
+            val chapterName = chapter.name.trim().takeIf { it.isNotEmpty() }
 
-            Pair(chapter, fileName)
+            val fileName =
+                buildString {
+                    append(mangaTitle)
+                    append(" - ")
+
+                    if (chapterName != null) {
+                        append(chapterName)
+                    } else if (chapter.chapterNumber >= 0f) {
+                        // chapterNumber is stored as Float, drop .0 for whole numbers.
+                        val formatNumber =
+                            if (chapter.chapterNumber % 1 == 0f) {
+                                chapter.chapterNumber.toInt().toString()
+                            } else {
+                                chapter.chapterNumber.toString()
+                            }
+                        append("#$formatNumber")
+                    } else {
+                        // Fallback when neither name nor valid chapter number exists
+                        append("#${chapter.index}")
+                    }
+
+                    if (scanlatorName != null) {
+                        append(" [")
+                        append(scanlatorName)
+                        append("]")
+                    }
+                    append(".cbz")
+                }
+
+            // Sanitize filename for OS compatibility
+            val safeFileName = fileName.replace(Regex("[\\\\/:*?\"<>|]"), "_")
+
+            Pair(chapter, safeFileName)
         }
 
     fun getCbzForDownload(
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
@@ -64,6 +64,11 @@ object ChapterDownloadHelper {
         chapterId: Int,
     ): Pair<InputStream, Long> = provider(mangaId, chapterId).getAsArchiveStream()
 
+    fun getChapterArchiveSize(
+        mangaId: Int,
+        chapterId: Int,
+    ): Long = provider(mangaId, chapterId).getArchiveSize()
+
     private fun getChapterWithCbzFileName(chapterId: Int): Pair<ChapterDataClass, String> =
         transaction {
             val row =
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
@@ -71,46 +71,13 @@ object ChapterDownloadHelper {
                     .select(ChapterTable.columns + MangaTable.columns)
                     .where { ChapterTable.id eq chapterId }
                     .firstOrNull() ?: throw IllegalArgumentException("ChapterId $chapterId not found")
-
             val chapter = ChapterTable.toDataClass(row)
-            val mangaTitle = row[MangaTable.title].trim()
+            val mangaTitle = row[MangaTable.title]
 
-            val scanlatorName = chapter.scanlator?.trim()?.takeIf { it.isNotEmpty() }
-            val chapterName = chapter.name.trim().takeIf { it.isNotEmpty() }
+            val scanlatorPart = chapter.scanlator?.let { "[$it] " } ?: ""
+            val fileName = "$mangaTitle - $scanlatorPart${chapter.name}.cbz"
 
-            val fileName =
-                buildString {
-                    append(mangaTitle)
-                    append(" - ")
-
-                    if (chapterName != null) {
-                        append(chapterName)
-                    } else if (chapter.chapterNumber >= 0f) {
-                        // chapterNumber is stored as Float, drop .0 for whole numbers.
-                        val formatNumber =
-                            if (chapter.chapterNumber % 1 == 0f) {
-                                chapter.chapterNumber.toInt().toString()
-                            } else {
-                                chapter.chapterNumber.toString()
-                            }
-                        append("#$formatNumber")
-                    } else {
-                        // Fallback when neither name nor valid chapter number exists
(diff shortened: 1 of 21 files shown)


zeedif:
Updated the PR to incorporate some improvements I found during my daily use:

- Added a fallback for empty chapter titles (to "Chapter X" or "Oneshot") because some extensions don't provide names and it looked broken in the feed.
- Source names now include their language code (e.g., `Source (EN)`) to avoid confusion, especially with multi-language extensions where it was hard to tell which one to select.
- Fixed a bug in the CBZ filename builder that left an awkward space between the scanlator brackets and the `.cbz` extension when the chapter name was missing.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
```diff
@@ -71,13 +71,46 @@ object ChapterDownloadHelper {
...
+                        append(scanlatorName)
+                        append("]")
+                    }
+                    append(".cbz")
+                }
+
+            // Sanitize filename for OS compatibility
+            val safeFileName = fileName.replace(Regex("[\\\\/:*?\"<>|]"), "_")
+
+            Pair(chapter, safeFileName)
```
Syer10: What is the point of these changes? I would rather not change how CBZs are named on the disk.

zeedif: Actually, I noticed some issues with the old logic while testing the downloads. 

If a manga didn't have a `chapter.name`, they ended up named like `"mangaTitle - scanlatorPart .cbz"`, or just `"mangaTitle - .cbz"` if there was no scanlator. This was a bit risky because multiple chapters could overwrite each other with the same name, and it also left an ugly empty space before the `.cbz` extension.

To fix this, I trimmed the strings and added a fallback: if `chapter.name` is missing, it now uses the `chapterNumber` so every file stays unique. Since we are now guaranteed to have a chapter identifier, I moved the scanlator name to the end in brackets `[scanlator]` to make it look cleaner.

Lastly, I added a simple regex to replace invalid characters (`/`, `:`, `*`, `?`, `"`, `<`, `>`, `|`, and `\`) with underscores. This prevents the download from failing if any manga, chapter, or scanlator names contain characters that the OS doesn't support. I'm not entirely sure if this sanitization is strictly necessary since browsers usually clean up invalid characters during download anyway, but I figured it was safer to have it in place.

zeedif: For example, the InManga (ES) extension has no chapter.scalator, while Hitomi doesn't save a chapter.name (since each manga only has a single chapter). Since I don't know how every extension behaves, I proposed this fallback logic, but I'm completely fine with reverting or changing any part of it if you prefer—my main goal is just to prevent that awkward empty space before the extension when the chapter name is missing.

zeedif: Also, I remembered why I swapped the order of the chapter name and the scanlator: many chapters are just named 'Chapter X'. Putting the scanlator first breaks the sequential sorting on disk when a manga has releases from multiple different scanlators. Placing the chapter name first keeps the files sorted in order, leaving the scanlator at the end purely as a reference.

Syer10: Can you split ChapterForDownload.kt and ChapterDownloadHelper.kt changes to another PR, or are they needed here?

zeedif: I've reverted the `ChapterDownloadHelper.kt` filename changes as requested. Regarding `ChapterForDownload.kt`, I'm currently looking into reverting it, but there's a performance catch. The reason I extracted `updateChapterPersistence` and `refreshChapterPageList` was to avoid the N+1 query problem in the OPDS chapter list.

If I just call `getChapterDownloadReady()` directly in a loop for all 50 chapters in the feed, it triggers redundant DB `SELECT` queries for every single chapter. By extracting the internal logic, I can do the disk checks and update the DB in parallel using the chapter data we already fetched in bulk.

zeedif: Alternatively, I can make skipMetadata non-functional for now (keeping the traditional feed) and open a separate, dedicated PR for the entire skipMetadata implementation and its optimization logic. Let me know which approach you prefer.

server/src/main/kotlin/suwayomi/tachidesk/opds/repository/ChapterRepository.kt
```diff
@@ -38,69 +51,163 @@ object ChapterRepository {
...
+                                        }.getOrDefault(0)
+                                }
+                            }
+                        }
+
+                        // Calculate CBZ size if downloaded
+                        val cbzFileSize =
+                            if (isDownloaded) {
+                                runCatching {
+                                    ChapterDownloadHelper.getArchiveStreamWithSize(entry.mangaId, entry.id).second
```
Syer10: Why is this not using `ChapterDownloadHelper.getCbzMetadataForDownload`? Seems like a better fit.

Also if you still use this, you need to close the unused stream
```suggestion
                                    ChapterDownloadHelper.getArchiveStreamWithSize(entry.mangaId, entry.id).also { it.first.close() }.second
```

server/src/main/kotlin/suwayomi/tachidesk/opds/repository/ChapterRepository.kt
```diff
@@ -38,69 +51,163 @@ object ChapterRepository {
...
+            lastPageRead = chapterDataClass.lastPageRead,
+            lastReadAt = chapterDataClass.lastReadAt,
+            sourceOrder = chapterDataClass.index,
+            downloaded = chapterDataClass.downloaded,
+            pageCount = chapterDataClass.pageCount,
+            url = chapterDataClass.realUrl,
+            cbzFileSize =
+                if (chapterDataClass.downloaded) {
+                    withContext(Dispatchers.IO) {
+                        runCatching { ChapterDownloadHelper.getArchiveStreamWithSize(mangaId, chapterDataClass.id).second }.getOrNull()
```
Syer10: Same as the other review.

```suggestion
                        runCatching { ChapterDownloadHelper.getArchiveStreamWithSize(mangaId, chapterDataClass.id).also { it.first.close() }.second }.getOrNull()
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/chapter/ChapterForDownload.kt
```diff
@@ -31,6 +31,87 @@ import suwayomi.tachidesk.manga.model.table.PageTable
...
+            needsUpdate = true
+        }
+        needsUpdate
+    }
+}
+
+suspend fun refreshChapterPageList(
+    mangaId: Int,
+    chapterId: Int,
+): Int {
```
Syer10, line 75: We need to be able to pass the original chapterEntry from ChapterForDownload here, so add a nullable parameter so it only grabs it when we need it.

Syer10 (changes requested):

Syer10 (commented):
Also add a changelog entry

zeedif:
I've just pushed a couple of commits applying your suggestions. I avoided `getCbzMetadataForDownload` because it executes a database transaction to retrieve the manga/chapter titles for the filename. Doing this in a loop across the feed's chapters would reintroduce the N+1 DB query issue I'm trying to fix with this extraction.

However, you are absolutely right about the unclosed stream leak. Instead of opening a stream just to close it immediately with `.also { it.first.close() }`, I added a small `getChapterArchiveSize` helper in `ChapterDownloadHelper.kt` that delegates directly to `provider().getArchiveSize()`. This gets the size without opening unnecessary streams. I kept `getArchiveStreamWithSize` intact because it is still actively required by `KoreaderSyncService` (which needs the actual bytes to compute the binary file hash).

I also added the nullable existingChapterEntry: ResultRow? = null parameter to refreshChapterPageList so we can reuse the row we already have in ChapterForDownload and save an extra DB query.
 
Finally, I added the entry to CHANGELOG.md as requested. Let me know if anything else is needed.

Syer10 (approved):

PR #1909, opened 2026-02-16
fix: support for `POST` requests on `CloudflareInterceptor`

D-Brox:
Works with Flaresolverr. Required for Kagane.

Sadly, Byparr is not a drop-in replacement for everything, it just ignores the `cmd` and interprets anything it receives as a GET request.

### DIFF
--- a/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
+++ b/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
@@ -208,9 +208,10 @@ object CFClearance {
                                                 },
                                             returnOnlyCookies = onlyCookies,
                                             maxTimeout = timeout.inWholeMilliseconds.toInt(),
-                                            postData = if (originalRequest.method == "POST") {
-                                                originalRequest.body?.let { encodeToString(it) }
-                                            } else null,
+                                            postData = if (originalRequest.method == "POST" && originalRequest.body is FormBody) {
+                                                Buffer().also { (originalRequest.body as FormBody).writeTo(it) }
+                                                    .readUtf8()
+                                            } else null
                                         ),
                                     ).toRequestBody(jsonMediaType),
                         ),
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -208,6 +208,7 @@ object CFClearance {
                                                 },
                                             returnOnlyCookies = onlyCookies,
                                             maxTimeout = timeout.inWholeMilliseconds.toInt(),
+                                            postData = if (originalRequest.method == "POST") originalRequest.body.toJSONString() else null,
```
Syer10: I would prefer to use another json library for the toJSONString function, like Jackson or Kotlinx. We don't use fastjson2 anywhere else.

Syer10 (changes requested):

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -208,6 +208,9 @@ object CFClearance {
                                                 },
                                             returnOnlyCookies = onlyCookies,
                                             maxTimeout = timeout.inWholeMilliseconds.toInt(),
+                                            postData = if (originalRequest.method == "POST") {
+                                                originalRequest.body?.let { encodeToString(it) }
+                                            } else null,
```
Syer10: I checked the FlareSolverr API and it expects only FormBody [postData](https://github.com/FlareSolverr/FlareSolverr?tab=readme-ov-file#-requestpost)

```suggestion
                                            postData = if (originalRequest.method == "POST" && originalRequest.body is FormBody) {
                                                Buffer().also { (originalRequest.body as FormBody).writeTo(it) }
                                                    .readUtf8()
                                            } else null
```

This seems to be the best way to match the API spec. Be sure to test it to make sure it works with Kagane.

Syer10 (changes requested):

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
D-Brox: Kaganes post request has no data in the body, so it doesn't matter.
I tested the endpoint on the network devtools with and without data, it'll just ignore.

I'll build and test locally tho

D-Brox: It's working

Syer10 (approved):
Just run the linter(`./gradlew ktlintFormat`) and it should be good.

D-Brox:
`originalRequest.body is FormBody` broke it after all, I'll fix it in another PR.
My tests probably didn't fail because I had the cookie cached

PR #1915, opened 2026-02-20
fix: default body to empty string if not present for FlareSolverr `POST` requests

D-Brox:


Syer10 (commented):
Flaresolverr doesn't support json based on its API spec, and converting json to forms is not something we should do.

I don't think that this will solve any issues.

Syer10:
image
I see. So a postBody is needed for request.post.

For this I would just pass a empty string as it should accept that if there is no actual body.

Syer10 (changes requested):
Just verify that its tested and I'll give it a merge.

Syer10 (approved):

D-Brox:
I have tested by copying the jar to a new docker instance (without volumes so it doesn't use my existing cookies), Kagane is working.

Json post requests from other extensions may still fail, since it would require for flaresolverr to implements other content-type post requests.

PR #1916, opened 2026-02-21
fix: don't resuse invalidated `cf_clearance` cookie on `CloudFlareInterceptor`

D-Brox:
Old invalidated `cf_clearance` cookies can cause http code 403 on Kagane requests and invalid responses on other sources:

image

Filtering this cookie only in the FlareSolverr request let's it replace the cookie when needed, otherwise it tries to use the old one.

This PR also dedups the sent cookies.

### DIFF
--- a/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
+++ b/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
@@ -205,7 +205,8 @@ object CFClearance {
                                             session = serverConfig.flareSolverrSessionName.value,
                                             sessionTtlMinutes = serverConfig.flareSolverrSessionTtl.value,
                                             cookies =
-                                                network.cookieStore.get(originalRequest.url)
+                                                network.cookieStore
+                                                    .get(originalRequest.url)
                                                     .groupBy { it.name }
                                                     .filter { it.key !in CloudflareInterceptor.COOKIE_NAMES }
                                                     .values
@@ -280,7 +281,8 @@ object CFClearance {
                     }
             logger.trace { "New cookies\n${cookies.joinToString("; ")}" }
             val finalCookies =
-                network.cookieStore.get(originalRequest.url)
+                network.cookieStore
+                    .get(originalRequest.url)
                     .groupBy { it.name }
                     .values
                     .map { cookies -> cookies.maxBy { it.expiresAt } }
--- a/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
+++ b/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
@@ -103,7 +103,7 @@ class CloudflareInterceptor(
     companion object {
         private val ERROR_CODES = listOf(403, 503)
         private val SERVER_CHECK = arrayOf("cloudflare-nginx", "cloudflare")
-        private val COOKIE_NAMES = listOf("cf_clearance")
+        val COOKIE_NAMES = listOf("cf_clearance")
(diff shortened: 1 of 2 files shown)


D-Brox:
The issue only affects flaresolverr, byparr works fine without this because it ignores the cookie

Syer10 (commented):
What are the options for fixing this in the cookie store itself? Why is it doing this, is the same cookie coming from different subdomains?

D-Brox:
@Syer10 The `cf_clearance` token sometimes gets invalidated by cloudflare before it's cookie expiration time.

The detection may not be trivial because it could either appear as a cloudflare challenge html page or a few possible http error codes.

This doesn't affect Byparr because it just ignores the cookie, while Flaresolverr always tries to use the one provided in the request (if it's not provided it'll just solve the new challenge instead). 

Tbh, the easiest way to solve this is filtering.

Syer10:
If we are getting a new cookie with the same host, it should be replacing the existing cookie in the cookie store, if it's not that's a bug with the cookie store that needs to be fixed in the cookie store, not in the cloudflare interceptor.

You can try reproducing the issue and then looking at your cookiestore.xml in the data folder. Share the cookie elements here.

D-Brox:
The expiration/replacement works correctly, but the issue is precisely that we are not getting a new cookie from Flaresolverr (at least until it expires).

Since suwayomi sends a `cf_clearance` cookie in the Flaresolverr request, Flaresolverr doesn't even try to obtain a new one, even if cloudflare invalidated the one we sent (the issue is technically on their side, but we can just not send that cookie).

Byparr doesn't have that issue because it just ignores the `cf_clearance` cookie in the request and gets a new one, as it should.

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -205,9 +205,15 @@ object CFClearance {
...
-                                                },
+                                                network.cookieStore
+                                                    .get(originalRequest.url)
+                                                    .groupBy { it.name }
+                                                    .filter { it.key !in CloudflareInterceptor.COOKIE_NAMES }
+                                                    .values
+                                                    .map { cookies ->
+                                                        val cookie = cookies.maxBy { it.expiresAt }
+                                                        FlareSolverCookie(cookie.name, cookie.value)
+                                                    },
```
Syer10, line 213: It was my understanding that the maxBy was the fix, if its just the cookie name filter it can be simplified to this.

```suggestion
                                                network.cookieStore
                                                    .get(originalRequest.url)
                                                    .filter { it.name !in CloudflareInterceptor.COOKIE_NAMES }
                                                    .map { cookie ->
                                                        FlareSolverCookie(cookie.name, cookie.value)
                                                    },
```

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -275,9 +281,14 @@ object CFClearance {
...
-                    "${it.name}=${it.value}"
-                }
+                network.cookieStore
+                    .get(originalRequest.url)
+                    .groupBy { it.name }
+                    .values
+                    .map { cookies -> cookies.maxBy { it.expiresAt } }
+                    .joinToString("; ", postfix = "; ") {
+                        "${it.name}=${it.value}"
+                    }
```
Syer10: This change can be reverted if the maxBy doesn't do anything

Syer10 (changes requested):

D-Brox:
IIRC the `CookieStore` may have an issue with duplicated cookies, but it's been so long since I added this I may not be remembering it well.

Syer10 (approved):
No problem, if the cookie store does have issues we will need to fix it there 😄

Syer10 (changes requested):
Almost forgot, add it to the changelog.md. You will need to merge/rebase for that

Syer10 (approved):

PR #1932, opened 2026-03-06
Detect non-standard bot challenges behind Cloudflare

manilkadev3-max:
Some sites behind Cloudflare use non-standard bot challenges (vShield/BalooPow) that return HTTP 200 instead of 403/503. The interceptor skips these since it only checks for 403/503, so FlareSolverr never gets called and extensions fail to parse the challenge page.

This peeks at the response body when it's HTTP 200 from a Cloudflare server and checks for known challenge markers. If found, it routes through FlareSolverr like a normal Cloudflare challenge.

Fixes issues with Iken-based extensions (e.g. Vortex Scans) throwing `Unable to retrieve NEXT data`.

### DIFF
--- a/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
+++ b/server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.sync.withLock
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonNames
 import okhttp3.Cookie
 import okhttp3.FormBody
 import okhttp3.HttpUrl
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -76,7 +89,8 @@ class CloudflareInterceptor(
                     if (!isImage) {
                         logger.debug { "Falling back to FlareSolverr response" }
 
-                        setUserAgent(flareResponse.solution.userAgent)
+                        // Save cookies from FlareSolverr so subsequent requests can bypass
+                        CFClearance.requestWithFlareSolverr(flareResponse, setUserAgent, originalRequest)
```
Syer10: Why are we requesting with flaresolverr again?

Syer10 (changes requested):

server/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
manilkadev3-max: Good catch — the naming was confusing. Refactored: extracted a saveSolutionCookies() method that only persists cookies and sets the user agent (no request building). The fallback path now calls that instead. requestWithFlareSolverr() delegates to it internally to avoid duplication.

Why we need to save cookies in the fallback path (where the original code only set the user agent): for non-standard challenges like vShield, FlareSolverr returns solution cookies (e.g. _2__vShield_v) but reports "Challenge not detected!". Without persisting those cookies, every subsequent request would trigger another FlareSolverr call.

PR #1933, opened 2026-03-07
fix: truncate filenames by bytes instead of characters to avoid File name too long

Akiaki0324:
Truncate filenames by bytes instead of characters in `buildValidFilename`


AndroidCompat/src/main/java/xyz/nulldev/androidcompat/util/SafePath.kt
```diff
@@ -27,11 +30,41 @@ object SafePath {
...
             }
         }
-        // Even though vfat allows 255 UCS-2 chars, we might eventually write to
-        // ext4 through a FUSE layer, so use that limit minus 15 reserved characters.
-        return sb.toString().take(240)
+
+        return truncateFilename(sb.toString())
+    }
+
+    private fun truncateFilename(filename: String): String {
```
Syer10, line 37: Mihon added a smilar function recently, could we pull from that instead? https://github.com/mihonapp/mihon/blob/f6b2684323569ef0eb23e143cc5d65d7cc1aae3c/core/common/src/main/kotlin/eu/kanade/tachiyomi/util/storage/DiskUtil.kt#L171

Syer10 (changes requested):

AndroidCompat/src/main/java/xyz/nulldev/androidcompat/util/SafePath.kt
Akiaki0324: While Mihon's code does the same thing, mine is likely more memory-efficient when dealing with exceptionally long filename inputs.

Syer10 (approved):

Syer10 (changes requested):
Rebase and add a [CHANGELOG.md](https://github.com/Suwayomi/Suwayomi-Server/blob/master/CHANGELOG.md) entry.

Section Downloads

Akiaki0324:
> Rebase and add a [CHANGELOG.md](https://github.com/Suwayomi/Suwayomi-Server/blob/master/CHANGELOG.md) entry.
> 
> Section Downloads

Done

Syer10 (approved):

PR #1944, opened 2026-03-31
filter scanalator fix for unread count and resume button

Chiru-Dey:
Filtering Scanalator now:-
i) Assigns correct scanalator unread chapter.
ii) shows correct unread count.

Tried to good GraphQL API documentation.

Note: if it is not correctly showing these, click on 'Reload the source" button.


Syer10 (changes requested):
If specific code is added for excluded scanaltors it won't be using the manga meta table.

It would be its own coloums or tables modeled on how Mihon handles excluded scanlators.

Using manga meta for storing something that changes behaviour of the server is not a good idea, it's meant for client only data that the client needs to store.

Chiru-Dey:
i tried to do this. pls check and test throughly. it is not compatible with current webui version.

schroda:
Why is it a new table instead of a column in the manga table?

What about all the other places the filtered scanlators need to potentially be considered?
- manga type:
  - downloadCount
  - unreadCount
  - bookmarkCount
  - hasDuplicateChapters
  - lastReadChapter
  - latestReadChapter
  - latestFetchedChapter
  - latestUploadedChapter
  - highestNumberedChapter
- during auto downloads

Should there be a flag that can be passed to optionally ignore the excluded chapters in the queries?

Syer10:
Its a new table so that we can have multiple filtered scanlators per manga while keeping it usable in sql.

Issue #1185, opened 2024-12-06
[Feature Request] editable information on entries


xdeiimoss:
## What feature should be added to Suwayomi?
Forks like TachiSY or J2K/Yokai have the option to edit information on entries (description, title, status, author or artist names, etc)
The same thing should be possible in Suwayomi.
Currently, if you do have edited entries and you update your library in Suwayomi, it will remove all of the edits and refresh it to the ones found on the site, which should be fixed

Robonau:
similar to #1009

Issue #1009, opened 2024-08-19
[Feature Request] Downloading metadata for local manga


Hunter-G01:
Request for the development of several features: metadata downloading, web-based metadata editing, and manga cover modification.

The metadata downloading feature ought to mirror what Jellyfin provides. Due to restrictions with current plugins, numerous websites are unsupported, leading to a deficiency in metadata when employing third-party tools for downloads. Manually inputting this information is excessively cumbersome.

Rather than altering the local code, there should be an in-interface manual editing function for metadata adjustments.

For manga covers, there should be an option to select any single image or any size to be utilized as the cover image.

Robonau:
why would the metadata follow what jellyfin does and not what other manga/comic media library's do?

Hunter-G01:
I'm just using Jellyfin as an analogy, because in terms of its metadata download feature, it's the most concise, convenient, and easy-to-use among all the web platforms I've ever used. If other comic media libraries have better solutions, they can also be taken into consideration. Although online comics already have the feature to download metadata, the most critical point is to add the feature to obtain metadata for local comics.

RA341:
Something like audiobookshelf would be pretty nice.  

See image. We can search for the book title, and then it would autofill the metadata.

![image](https://github.com/user-attachments/assets/62ed100b-6756-42cc-a291-a9b2c3c28052)

Issue #978, opened 2024-06-25
[Feature Request] Option to rename manga in the library and download folder


KiRa937:
## What feature should be added to Suwayomi?
It should be possible to rename manga in the library as well as its download directory. The naming should be transferable during migration.

##  Why/Project's Benefit/Existing Problem
It should help to avoid problematic names in the paths (names too long, unreadable symbols e.t.c.). Also to make it possible to organize manga naming regardless of the titles provided by sources: fully translated, romanji written, kanji written, copyright-avoiding names (hello, "Pirate King").

daviddavo:
+1 to this request. I also have the same manga in multiple languages or from multiple sources and I'd like to be able to differentiate them.

Robonau:
this probably wont be added as it is not a feature of mihon.

if it were to be added it would likely fail in some aspects, like not being in backups 
(due to backups needing to be compatible with mihon)

since migration is done entirely on the client side it wouldn't be an issue for that to function correctly

this is really just a more minimal version of #1009 that is asking for all of the metadata to be editable

Aeres-u99:
Sometimes, post download the filename might contain special characters like `#` which can break some of the workflows, (For instance in my case, I download as cbz -> convert to pdf and then use k2pdfopt to convert it for kindle format)
Having an automatic renaming of some sorts will help.

PR #1946, opened 2026-04-04
feat: add local manga metadata editing

HaroonChelsea:
## Summary

Adds the ability to edit manga metadata (title, author, artist, description, genre, status) via a new `updateMangaDetails` GraphQL mutation. For local source manga, changes are persisted to `details.json`.

Includes `LocalMangaDetailsService` with atomic file writes (temp + rename), concurrent access protection (Mutex), and field clearing support (empty string = clear field, empty list = clear genres).

**Part 1 of 3** — local manga editing. See also:
- Part 2: Cover image upload
- Part 3: Metadata providers (AniList, MangaUpdates)

Addresses #1009, #1185, #978

## Test plan

- [x] 9 unit tests for LocalMangaDetailsService (read/merge/write, field clearing, corrupt file)
- [x] Manual test: updateMangaDetails via GraphiQL
- [x] Manual test: details.json written in local manga directory
- [x] Build: `shadowJar` passes
- [x] Lint: `ktlintCheck` passes

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/MangaMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/MangaMutation.kt
@@ -1,5 +1,7 @@
 package suwayomi.tachidesk.graphql.mutations
 
+import eu.kanade.tachiyomi.source.local.LocalSource
+import eu.kanade.tachiyomi.source.local.io.LocalSourceFileSystem
 import graphql.execution.DataFetcherResult
 import org.jetbrains.exposed.sql.LikePattern
 import org.jetbrains.exposed.sql.Op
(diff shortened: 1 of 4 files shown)


xdeiimoss:
Hello. Would this implementation be compatible with backups from tachiyomisy? If not, could it possibly be done?

HaroonChelsea:
> Hello. Would this implementation be compatible with backups from tachiyomisy? If not, could it possibly be done?

TachiyomiSY stores user-edited metadata in separate custom protobuf fields (customTitle, customArtist, customAuthor, customDescription, customGenre, customStatus). These fields aren't currently mapped in Suwayomi's backup model, so metadata you manually edited in SY would be dropped during import into Suwayomi.

**Can it be done?** Yes, it would be a small follow-up. We'd add those proto fields to Suwayomi's BackupManga model and map them in the restore handler to apply via the same metadata editing logic this PR introduces.

PR #1957, opened 2026-04-18
fix: only clear thumbnail cache when URL actually changed

Skyrexolino:
Thumbnails were getting wiped on every library update even when the URL hadn't changed, because fetchManga() calls clearThumbnail() whenever the source returns any non-empty thumbnail URL without checking if it actually differs from the stored one. This could then in turn delete thumbnails and fail to download them if a source isn't available anymore.


schroda:
This behavior is intentional: ee4c852f1b4b6eba9ee4bac6140f5aaf5691fcb8

cpiber:
It is quite common for the URL to be the same, but the content to change.

You'd need to download the cover to a temporary location, and only if that succeeds (and additionally, maybe even only if the file has changed, to avoid cache invalidation), overwrite the existing cover.

PR #1958, opened 2026-04-21
Add Yamtrack as a new tracker service

Breezyslasher:
Adds Yamtrack as a new tracker service, letting users sync manga reading progress to a self-hosted Yamtrack instance alongside the existing trackers (MyAnimeList, AniList, Kitsu, etc.).
Features

    Self-hosted login — user provides their Yamtrack host URL + API token; credentials are verified against /api/v1/statistics/ before being saved.
    Search — queries /api/v1/search/manga/ and returns matches from Yamtrack's upstream providers (MangaUpdates, MAL, AniList, etc.), plus a "manual entry" fallback for items those providers don't index (mirrors Yamtrack's own source=manual flow).
    Bind / Track — first-time tracking POSTs to /media/manga/; the response's tracked flag is checked so already-tracked items skip the add and go straight to update.
    Update — PATCHes status, progress, score, and reading dates to /media/manga/{source}/{media_id}/.
    Refresh — pulls the latest consumption record back into the local track.
    Delete — removes the entry from the user's Yamtrack library.
    Reading dates — supportsReadingDates = true; start/end dates are serialized as ISO yyyy-MM-dd and round-tripped through the consumption record.
    Statuses — Planning, Reading, Paused, Completed, Dropped, mapped to Yamtrack's integer status codes (0–4) per api/helpers.py.
    Scores — 0.0–10.0 in 0.1 increments, submitted only when non-zero.

Implementation notes

    Stable remote IDs are derived by hashing (source, media_id) since Yamtrack identifies entries by a pair rather than a single integer.
    Tracking URLs use the form {base}/media/manga/{source}/{media_id} and are parsed back to (source, media_id) for API calls.
    max_progress = 1 is treated as "unknown" (Yamtrack's placeholder for ongoing manga with no chapter count) and mapped to total_chapters = 0, so the progress picker isn't capped at chapter 1.
    Auth is applied via a dedicated YamtrackInterceptor using Bearer token headers.
    Uses the official Yamtrack logo (android-chrome-512x512.png) as the tracker brand image.

## Summary by Sourcery

Add Yamtrack as a new self‑hosted tracker integration alongside existing tracking services.

New Features:
- Introduce a Yamtrack tracker implementation that supports login via host URL and API token, search, bind, update, refresh, and delete operations for manga tracking.
- Expose Yamtrack as a configurable tracker option in the tracking settings screen.
-

current errors:
 - Needs support for per-tracker custom password field labels to allow Yamtrack to label inputs as host URL and API token instead of a traditional username and password in the WebUI.

Enhancements:
- Map Yamtrack-specific statuses, scores, dates, and remote identifiers into the app’s generic tracking model, including a manual-entry flow and handling of unknown chapter counts.


Breezyslasher:
i did just copy my work from https://github.com/komikku-app/komikku/pull/1601 to make it compatible with Suwayomi-Server

Syer10:
This needs to be merged into Mihon before it will be considered to be merged here.

Breezyslasher:
Closing the pr for now until

> This needs to be merged into Mihon before it will be considered to be merged here.

OK I will close the pr for now until it get merged In to Mihon

Issue #1197, opened 2024-12-11
[Bug] System tray icon - no option on click - Windows 11


AvvAdeh:
Hello!
Nothing dramatic, but I liked the option to be able to easily quit Suwayomi server.
Currently I have to open the task manager and close the **OpenJDK Platform binary** process.

## Device information
- Suwayomi-Server version: v1.1.1-r1649 win64
- Server Operating System: Windows 11
- Server Desktop Environment: Windows 11 default
- Server JVM version: the one bundled with Suwayomi server
- Client Operating System: Windows 11
- Client Web Browser: Edge, with WebUI r1943

## Steps to reproduce
Launch Suwayomi, with the system tray enable activated in the server settings.

### Expected behavior
When clicking on the system tray icon, a contextual menu pops up, allowing to quit the server (and open a suwayomi page in the browser, if I recall correctly).

### Actual behavior
Clicking the icon do nothing (right or left click).

### Possible alternative
Add an option in the settings to shut down Suwayomi server.

Anyway, thank you (a lot!) for this great app :)

Syer10:
The quit option is under the right click menu, are you having trouble getting the right click menu?

AvvAdeh:
Yes, it doesn't work, no menu appears.

I switched from:
**stable** release v1.1.1 r1535 where the right click works,
and **preview** release v1.1.1 r1586 where it wasn't working. (But because the chapter order was reversed I didn't used it regurlarly and waited until several days ago to try another preview version, thus, I didn't noticed the right click didn't worked until now.)

the "latest" preview release where it worked for me was the v1.1.1-r1548, but I didn't tried each versions between r1549 and r1649.

Syer10:
Its likely related to https://github.com/dorkbox/SystemTray/issues/209

AvvAdeh:
Ah!
Thank you for finding the issue and making a follow up in the relevant thread. I wouldn't have thought of searching there :)

I'll let you decide if you want to close this thread/issue here for Suwayomi, I don't know how you'd prefer it to be.

videoclube:
I'm having the same issue on the latest version (2.1.1867) and W11, is there any way to fix it?

PR #1973, opened 2026-05-03
fix: putting native awt to windows, fixing windows tray

nauanelinhares:
## Problem

On Windows, the Dorkbox SystemTray library caused issues with the system tray icon. This fix replaces it with a native AWT-based implementation for Windows.

Before: We are unable to close the Suwayomi with the system tray. We had to open the task manager and finish the task manually.

After: We can close the app by system tray. Plus, double click to open the suwayomi.

image

## Changes

- Introduced a `SystemTrayHandler` interface with `create()` and `remove()` methods
- Added `WindowsAwtSystemTrayHandler`: a native AWT implementation used on Windows, avoiding Dorkbox entirely
- Added `DorkboxSystemTrayHandler`: extracts the existing Dorkbox logic into its own handler for non-Windows platforms
- Updated `SystemTray` object to select the appropriate handler at runtime based on the OS

## Behavior

- **Windows**: uses `java.awt.SystemTray` directly — no Dorkbox dependency involved
- **Linux / macOS**: continues using the Dorkbox-based handler as before


nauanelinhares:
Greetings [Syer10](https://github.com/Syer10), can you review that pull request? I am trying to fix that [issue](https://github.com/Suwayomi/Suwayomi-Server/issues/1197). It's quite old, and it doesnt seems there's been a solution so far

Syer10 (commented):
Doesn't Dorkbox have a AWT option? Maybe we should just add a config option to set the tray type using Dorkbox.

nauanelinhares:
> Doesn't Dorkbox have a AWT option? Maybe we should just add a config option to set the tray type using Dorkbox.

If it was that, the solution would be simple, but i tried [here](https://github.com/Suwayomi/Suwayomi-Server/commit/e380bf25aa1e514c2a2d2762ce0a7942c55dbeb5) and didnt work :(

nauanelinhares (commented):
@Syer10 can you see? 🙏🏾

PR #2027, opened 2026-05-14
Pin CEF version to one known to work with KCEF

cpiber:
Fixes problems like
```
java.lang.ClassNotFoundException: org.cef.callback.CefResourceReadCallback_N
```
and
```
Exception in thread "Thread-584" java.lang.NoSuchMethodError: open
```

Equivalent fix to docker incoming shortly

Let me know if you want a Changelog entry for this or not


Syer10 (approved):

Syer10:
Changelog will be handled by other ways we resolve this later.

Issue #1997, opened 2026-05-09
`CATEGORY.SORT_ORDER` and `CHAPTER.KOREADER_HASH` not found after an update


C0rn3j:
### Steps to reproduce

I have updated after a longer time (months at minimum) without an update, maybe that's not helping?

`docker image ls --filter dangling=true` shows `<untagged>   5c8407a6e927       1.39GB             0B` which could potentially be the image I used previously.

EDIT: Looks like I updated from v2.1.1867.

### Expected behavior

App launches and works fine

### Actual behavior

I can login but loading any tab gives an error.

```
org.jetbrains.exposed.exceptions.ExposedSQLException: org.h2.jdbc.JdbcSQLSyntaxErrorException: Column "CHAPTER.KOREADER_HASH" not found; SQL statement:
SELECT CHAPTER.ID, CHAPTER.URL, CHAPTER."NAME", CHAPTER.DATE_UPLOAD, CHAPTER.CHAPTER_NUMBER, CHAPTER.SCANLATOR, CHAPTER."READ", CHAPTER.BOOKMARK, CHAPTER.LAST_PAGE_READ, CHAPTER.LAST_READ_AT, CHAPTER.FETCHED_AT, CHAPTER.SOURCE_ORDER, CHAPTER.REAL_URL, CHAPTER.IS_DOWNLOADED, CHAPTER.PAGE_COUNT, CHAPTER.MANGA, CHAPTER.KOREADER_HASH FROM CHAPTER INNER JOIN MANGA ON MANGA.ID = CHAPTER.MANGA WHERE MANGA.IN_LIBRARY = ? ORDER BY CHAPTER.FETCHED_AT DESC, CHAPTER.SOURCE_ORDER DESC, CHAPTER.ID ASC [42122-200]
...
Caused by: org.h2.jdbc.JdbcSQLNonTransientException: General error: "java.lang.IllegalStateException: Unable to read the page at position 96170708817765466 [1.4.200/6]"; SQL statement:
SELECT COUNT(*) FROM CHAPTER WHERE (CHAPTER.LAST_READ_AT IS NOT NULL) AND (CHAPTER.LAST_READ_AT <> ?) [50000-200]
        at org.h2.message.DbException.getJdbcSQLException(DbException.java:505)
        at org.h2.message.DbException.getJdbcSQLException(DbException.java:429)
        at org.h2.message.DbException.get(DbException.java:194)
[...]

PR #2029, opened 2026-05-14
Update Troubleshooting

cpiber:
Adds some well-known problems to the Troubleshooting page and how to solve them.

I wanted to add the icon for "Update in WebView", but Github Markdown does not seem to allow that with a dedicated file, no embedding the svg or png...

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -1,19 +1,12 @@
 ## General Troubleshooting
 This guide will try to fix Suwayomi by reseting it to a clean installation state.
 
+> [!WARNING]
+> This will remove all your data, including the library.
+> Make sure you have copied your backups as described above!
+
 - Make sure you have a recent backup of your library or create one in the app (if possible) because we **are going to wipe all Suwayomi data**.
 - Make sure Suwayomi is not running (right click on tray icon and quit or kill it through the way your Operating System provides)
 - Clear all browsing data on your browser if you use Suwayomi from a browser.
-- Delete the Suwayomi data directory located below and re-run the app.
-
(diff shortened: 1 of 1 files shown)


docs/Troubleshooting.md
```diff
@@ -1,19 +1,68 @@
...
+
+At the end, you will find a General section, which is the nuclear option if nothing else works.
+
+For further support, visit the [official Suwayomi Discord server](https://discord.gg/DDZdqZWaHA).  
+In such cases, it will be helpful to have logs ready. You can find them in [The Data Directory](./The-Data-Directory) in the logs directory.
+
+**All steps below assume that you have stopped Suwayomi**.
+
+
+## Broken database
```
Syer10, line 13: This is outdated since we now have a Stable version with the sort order change

cpiber: While that is true, I expect similar errors in the future when the DB is updated, and I've also seen those errors on a corrupted DB. We could change the text to remove the references on the sort order though

Syer10: This is a good example of a corrupted database https://github.com/Suwayomi/Suwayomi-Server/issues/1997

cpiber: Yes, that was my inspiration. And this one says `CHAPTER.KOREADER_HASH` instead of `CATEGORY.ORDER`, otherwise it's the same. I've updated it, and added the text from the inner exception.

Syer10 (approved):

cpiber:
Shit, the `[skip ci]` messed up the updating of the wiki, can you trigger it manually?

Syer10:
Looks like I can't, it doesn't have workflow dispatch.

PR #1856, opened 2026-01-13
Add disable kcef flag to disable kcef webview for platforms which don't support it

KolbyML:
Android and iOS don't support kcef, also apparently riscv doesn't either

We want to use native webview's on Android/iOS

kcef is also 500mb download which is quite big on mobile

PR #2038, opened 2026-05-16
Switch to JCEF

cpiber:
Needs some more testing, but moves to pure JCEF, with CEF download mostly copied from KCEF. Will test some more tomorrow

I think it would also make sense to create a test that verifies that the JCEF version matches the pinned JBR release for CEF, and to automatically use the version from the code here for the docker, while look into that tomorrow


server/build.gradle.kts
```diff
@@ -159,6 +163,15 @@ buildConfig {
...
     buildConfigField("String", "DISCORD", quoteWrap("https://discord.gg/DDZdqZWaHA"))
+
+    val jcefDep =
+        project.configurations
+            .getByName("implementation")
+            .allDependencies
+            .matching {
+                it.group.equals("org.jetbrains.intellij.deps.jcef") && it.name.equals("jcef")
+            }.firstOrNull()
+    buildConfigField("String", "JCEF_VERSION", quoteWrap(jcefDep?.version ?: ""))
```
Syer10: libs.jcef.version.get()

cpiber: Had to extract it to a field in `libs.versions`, otherwise I couldn't get the version, only set it?

```
  Line 166:     buildConfigField("String", "JCEF_VERSION", quoteWrap(libs.jcef.version.get()))
                                                                               ^ None of the following candidates is applicable:
                                                                                 
                                                                                   fun PluginDependencySpec.version(version: String?): PluginDependencySpec:
                                                                                     Unresolved reference. None of the following candidates is applicable because of a receiver type mismatch:
                                                                                   fun PluginDependencySpec.version(version: String?): PluginDependencySpec
                                                                                 
                                                                                   fun PluginDependencySpec.version(version: Provider<String>): PluginDependencySpec:
                                                                                     Unresolved reference. None of the following candidates is applicable because of a receiver type mismatch:
                                                                                   fun PluginDependencySpec.version(version: Provider<String>): PluginDependencySpec

  Line 166:     buildConfigField("String", "JCEF_VERSION", quoteWrap(libs.jcef.version.get()))
                                                                                          ^ No value passed for parameter 'name'.
```

cpiber:
The core code stayed mostly the same, with only minor helper methods to be added.
Largest changes come from having to download and extract CEF ourselves now. I've modified this to now also check that the downloaded CEF version is compatible.

Some troubles:

- CEF seems to have dropped pre-Windows 10 support. Do we care? Works on 10+
- On my ancient Mac, `cef_initialize` segfaults (ironically in `cef_dump_without_crashing_throttled`). I am not sure if this is because the device is old, or some other problem, but since it's inside CEF itself with no input from us, there's probably nothing we can do. But it would be nice to at least have an option to disable CEF, as #1856 proposes

cpiber:
> create a test that verifies that the JCEF version matches the pinned JBR release for CEF

At the moment, it seems tests aren't run at all. Maybe it would make sense to always run them for PRs?

> to automatically use the version from the code here for the docker

I'm thinking of putting the JBR version into the manifest and extracting that in the docker build, do you have any better idea?

Syer10:
> CEF seems to have dropped pre-Windows 10 support. Do we care? Works on 10+

I don't think we should care, major browsers have already dropped support for versions below 10 so we should follow it.

> On my ancient Mac, `cef_initialize` segfaults

Is it a Intel Mac or a M1 Mac? If its ARM I think that will be an issue.

>At the moment, it seems tests aren't run at all. Maybe it would make sense to always run them for PRs?

Tests are currently broken and outdated, we can remove old REST API ones but we still need to figure out the startup issues, having Tests run on PRs would be good but hasn't been a priority to fix.

> I'm thinking of putting the JBR version into the manifest and extracting that in the docker build, do you have any better idea?

Putting it in the manifest is a good plan, don't have any problems with that.

cpiber:
> Is it a Intel Mac or a M1 Mac? If its ARM I think that will be an issue.

Intel.

> Tests are currently broken and outdated, we can remove old REST API ones but we still need to figure out the startup issues, having Tests run on PRs would be good but hasn't been a priority to fix.

Currently working on it :)

Syer10:
> Intel.

Its probably fine to ignore then

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,552 @@
...
+            }
+        }
+        this.delete()
+    }
+
+object CEFManager {
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default + Dispatchers.IO)
+    private val applicationDirs = Injekt.get<ApplicationDirs>()
+    private val cefDir = Path(applicationDirs.dataRoot) / "bin/kcef"
+    private const val JBR_VERSION = "jbr-release-25.0.3b475.60"
```
Syer10: What is the update path for this?

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ xmlserialization = "0.91.3"
 ktlint = "1.8.0"
 koin = "4.2.1"
 moko = "0.26.4"
+jcef = "144.0.15-g72717cf-chromium-144.0.7559.172-api-1.21-262-b34"
```
Syer10, line 20: How does this version relate to the other jcef version defined in CEFManager?

AndroidCompat/src/main/java/xyz/nulldev/androidcompat/webkit/CefHelper.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import org.cef.CefApp
+import org.cef.CefClient
+import org.cef.CefSettings.LogSeverity
+import org.cef.SystemBootstrap
+import org.cef.handler.CefAppStateHandler
+import kotlin.concurrent.thread
+import kotlin.io.path.Path
+import kotlin.io.path.div
```
Syer10: Optimize imports

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,552 @@
...
+import java.io.FileOutputStream
+import java.io.IOException
+import java.net.http.HttpResponse
+import java.nio.file.Files
+import java.nio.file.LinkOption
+import java.util.Arrays
+import kotlin.concurrent.thread
+import kotlin.io.path.Path
+import kotlin.io.path.div
+import kotlin.streams.asSequence
```
Syer10, line 67: Optimize imports

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,552 @@
...
+                            "--disable-dev-shm-usage",
+                            // #1723 support Widevine (incomplete)
+                            "--enable-widevine-cdm",
+                            // #1736 JCEF does implement stack guards properly
+                            "--change-stack-guard-on-fork=disable",
+                        ),
+                    )
+                    cefSettings.apply {
+                        windowless_rendering_enabled = true
+                        cache_path = (Path(applicationDirs.dataRoot) / "cache/kcef").toString()
```
Syer10: Does this need to be absolute? Can use .absolutePathString()

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,552 @@
...
+            @Serializable
+            data class Asset(
+                val name: String = "",
+                @SerialName("browser_download_url") val downloadUrl: String = "",
+            )
+        }
+    }
+
+    // based on https://github.com/DatL4g/KCEF/blob/master/kcef/src/main/kotlin/dev/datlag/kcef/step/extract/TarGzExtractor.kt
+    internal data object TarGzExtractor {
```
Syer10, line 362: The following seems to switch between Java NIO Path and Java Util File a lot, is this worth fixing so its all Java NIO Path?

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
cpiber: `isInstallationValid` will return `false` for existing installations, since `bin/kcef/release` does not exist, and so it will re-download. If the `JBR_VERSION` is changed, the same happens.

gradle/libs.versions.toml
cpiber: Not directly, unfortunately. JBR includes a version of JCEF, which is described in the `release` file. I do not know of any way of relating them...

Syer10: What would I need to do if I wanted to update JCEF? Where would I get the versions needed?

cpiber: They usually publish JCEF on maven and JBR on https://github.com/JetBrains/JetBrainsRuntime/releases at the same time, so you can just try using the latest release. I will create a test to verify that those two match.

Syer10: I'll merge the other PRs first then and then you can start doing that with the fixed tests

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
cpiber: Changed it

cpiber:
Should be done. Left a TODO for download progress, not sure if needed/wanted?

@KolbyML I also included the toggle from #1856 (but renamed it to enable**d** for consistency)

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+import kotlin.io.path.listDirectoryEntries
+import kotlin.io.path.outputStream
+import kotlin.io.path.readLines
+import kotlin.io.path.readText
+import kotlin.io.path.writeText
+import kotlin.streams.asSequence
+
+private val logger = KotlinLogging.logger {}
+
+internal fun Path.deleteDir(): Result<Boolean> =
```
Syer10: This could be replaced with the Kotlin `Path.deleteRecursively()`

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default + Dispatchers.IO)
+    private val applicationDirs by lazy { Injekt.get<ApplicationDirs>() }
+    private val cefDir by lazy { Path(applicationDirs.dataRoot) / "bin/kcef" }
+    private val releaseFile by lazy { cefDir / "release" }
+
+    fun init() = scope.launch { initAsync() }
+
+    private suspend fun initAsync(): CefApp =
+        try {
+            if (!serverConfig.kcefEnabled.value) {
```
Syer10, line 97: Usually with settings we like to have it toggleable when the app is running and not require a restart, is that possible with CEF?

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+
+            val app = CefApp.getInstance(config.getAppArgs(), config.cefSettings, config.getServerExe())
+            CefHelper.cefApp.value = Result.success(app)
+            logger.debug { "CEF app created" }
+
+            Runtime.getRuntime().addShutdownHook(
+                thread(start = false) {
+                    logger.debug { "Shutting down CEF" }
+                    app.dispose()
+                    logger.debug { "KCEF shutdown complete" }
```
Syer10:
```suggestion
                    logger.debug { "CEF shutdown complete" }
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+
+        val downloadUrl =
+            client.newCall(request).awaitSuccess().use { response ->
+                if (!response.isSuccessful) throw IOException("Unexpected code $response")
+                GithubReleaseTransform.transform(response)
+            }
+
+        val tempDownload =
+            withContext(Dispatchers.IO) {
+                Files.createTempDirectory("cef")
```
Syer10: Kotlin has an extension function for this, `createTempDirectory()`

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+
+                            if (currentEntry != null) {
+                                val file = installDir / currentEntry.name
+                                if (!file.validate(installDir)) {
+                                    throw CefException("bad archive")
+                                }
+
+                                if (currentEntry.isDirectory) {
+                                    file.createDirectories()
+                                    file.toFile().setExecutable(true, false)
```
Syer10:
```suggestion
                                    file.setPosixFilePermissions(
                                        file.getPosixFilePermissions() + setOf(
                                            PosixFilePermission.OWNER_EXECUTE,
                                            PosixFilePermission.GROUP_EXECUTE,
                                            PosixFilePermission.OTHERS_EXECUTE
                                        )
                                    )
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+                                    val data = ByteArray(bufferSize.toInt())
+                                    BufferedOutputStream(
+                                        Files.newOutputStream(file, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE),
+                                        bufferSize.toInt(),
+                                    ).use { dest ->
+                                        while (tarIn.read(data, 0, bufferSize.toInt()).also { count = it } != -1) {
+                                            dest.write(data, 0, count)
+                                        }
+                                    }
+                                    file.toFile().setExecutable(true, false)
```
Syer10:
```suggestion
                                    file.setPosixFilePermissions(
                                        file.getPosixFilePermissions() + setOf(
                                            PosixFilePermission.OWNER_EXECUTE,
                                            PosixFilePermission.GROUP_EXECUTE,
                                            PosixFilePermission.OTHERS_EXECUTE
                                        )
                                    )
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            downloadedFile.deleteExisting()
+        }
+
+        fun move(installDir: Path) {
```
Syer10, line 451: This can be replaced with the Kotlin `Path.moveTo(Path)`

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+                                }
+
+                                if (currentEntry.isDirectory) {
+                                    file.createDirectories()
+                                    file.toFile().setExecutable(true, false)
+                                } else {
+                                    var count: Int
+                                    val data = ByteArray(bufferSize.toInt())
+                                    BufferedOutputStream(
+                                        Files.newOutputStream(file, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE),
```
Syer10:
```suggestion
                                        file.outputStream(StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE),
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+    // based on https://github.com/DatL4g/KCEF/blob/master/kcef/src/main/kotlin/dev/datlag/kcef/step/extract/TarGzExtractor.kt
+    internal data object TarGzExtractor {
+        internal fun Path.validate(parent: Path): Boolean =
+            runCatching {
+                this.normalize().startsWith(parent)
+            }.getOrNull() ?: false
+
+        internal fun Path.isSymlink(): Boolean =
+            runCatching {
+                Files.isSymbolicLink(this)
```
Syer10:
```suggestion
                this.isSymbolicLink()
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+            }.getOrNull() ?: runCatching {
+                !Files.isRegularFile(this, LinkOption.NOFOLLOW_LINKS)
```
Syer10:
```suggestion
                !this.isRegularFile(LinkOption.NOFOLLOW_LINKS)
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+            }.getOrNull() ?: false
+
+        internal fun Path.getRealFile(): Path =
+            if (isSymlink()) {
+                runCatching {
+                    Files.readSymbolicLink(this)
```
Syer10:
```suggestion
                    this.readSymbolicLink()
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+
+            return if (targetFile == null) {
+                false
+            } else {
+                this == targetFile || runCatching {
+                    sourceFile.absolute() == targetFile.absolute() ||
+                        Files.isSameFile(
+                            sourceFile,
+                            targetFile,
+                        )
```
Syer10:
```suggestion
                        sourceFile.isSameFileAs(targetFile),
```

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+                                val file = installDir / currentEntry.name
+                                if (!file.validate(installDir)) {
+                                    throw CefException("bad archive")
+                                }
+
+                                if (currentEntry.isDirectory) {
+                                    file.createDirectories()
+                                    file.toFile().setExecutable(true, false)
+                                } else {
+                                    var count: Int
```
Syer10: If possible change this to a standard .copyTo function

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
```diff
@@ -0,0 +1,554 @@
...
+                    .Builder()
+                    .url(downloadUrl)
+                    .build()
+
+            // TODO: progress?
+            downFile.outputStream().use { output ->
+                client.newCall(downloadRequest).awaitSuccess().use { response ->
+                    response.body.byteStream().use { input -> input.copyTo(output) }
+                }
+            }
```
Syer10: Progess can be done with the `.newCachelessCallWithProgress()` function

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
cpiber: I did notice that, but also that it's marked as experimental. Is that fine?

Syer10: Yep, its fine

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
cpiber: Partially. We could destroy the app instance, but I'm not sure how much sense that makes. Can look into it though

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
cpiber: I assume you meant to comment on `Path.move` above, the `move` function figures out which directories to keep and moves them to a flat structure

server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt
cpiber: Reverting this due to exception on Windows:

```
java.lang.UnsupportedOperationException
        at java.base/sun.nio.fs.WindowsFileSystemProvider.readAttributes(Unknown Source)
        at java.base/java.nio.file.Files.readAttributes(Unknown Source)
        at java.base/java.nio.file.Files.getPosixFilePermissions(Unknown Source)
        at suwayomi.tachidesk.server.util.CEFManager$TarGzExtractor.extract(CEFManager.kt:396)
        at suwayomi.tachidesk.server.util.CEFManager.downloadRelease$server(CEFManager.kt:229)
        at suwayomi.tachidesk.server.util.CEFManager$downloadRelease$1.invokeSuspend(CEFManager.kt)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
        at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:124)
        at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:798)
```

Edit: Left in, but with try/catch

Syer10: Ya Windows doesn't use the Posix file attributes so its uneeded. Put it in a try/catch or runCatching

Syer10 (changes requested):
It needs a changelog entry and the last thing about toggling the setting during runtime, otherwise I think its good.
The PR for docker should be done before this is merged I think.

server/src/test/kotlin/suwayomi/tachidesk/CefTest.kt
```diff
@@ -0,0 +1,40 @@
...
+                    }
+                }
+            startKoin {
+                modules(module)
+            }
+            try {
+                CEFManager.downloadRelease(tempDownload)
+                assertTrue { CEFManager.isInstallationValid(tempDownload / "release") }
+            } finally {
+                tempDownload.deleteDir()
```
Syer10: Need to fix the test

cpiber: Ups, good catch. Should I also enable the tests in the PR-CI?

Syer10: Ya, lets do that. I suggest disabling everything in the masstest package though.

cpiber:
I have implemented the setting to change at runtime, not sure how useful that is though. JCEF does not support shutting down and starting up again (https://github.com/JetBrains/jcef/blob/5b93e5b916068316f1c8e7f8a59bf958d5ffd6e1/java/org/cef/CefApp.java#L363-L364). I believe CEF also does not support that, it does not even allow to call `cef_initialize` multiple times.

One nice thing about the new JCEF: It now supports remote server, so CEF can be run out-of-process (or even on another server). That could be a nice usecase for some people (and CEF cannot bring down Suwayomi!)

Syer10 (approved):
Sounds great, once docker is ready I'll merge this

PR #2049, opened 2026-05-17
Fix tests

cpiber:
Some actual fixes:

- `ServerSetup` installed `JavalinJackson` instead of `JavalinJackson3` (`JavalinSetup` uses the correct one), which cannot deal with kotlin data classes (broke deserialization of `FilterChange` in `Search`, not tested if reproducible in running instance)
- Fix `PaginatedList` equality broken by 5b0426a94ca49278e3080a67da29d4e7c9723c70 (unknown if this has any effects in real setups)

`CloudflareTest` is disabled for now since it depends on having a source that requires Cloudflare, and we don't ship extensions anymore. `TestExtensionCompatibility` is running but useless for the same reason.


Syer10 (approved):
TestExtensionCompatibility is mainly just a mass extension tester I made to try to find broken extensions in Suwayomi to fix, its not actually a real test we should run regularly,

PR #2077, opened 2026-05-30
Add Flaresolverr info to troubleshooting

cpiber:
Let me know if you want these split into separate PRs.

- Add Flaresolverr info to Troubleshooting
- Insert a permalink to Configuring page for stable on wiki push.  
  Last release, it happened several times that people were confused that a config option did not work, because the page was updated for preview and the option was not yet available in stable (e.g. for auth modes)


Syer10 (approved):
Its probably fine to leave it combined in this PR

PR #2100, opened 2026-06-08
feat(download): improve chapter download filenames

zeedif:
This pull request refactors the CBZ filename generation in `ChapterDownloadHelper.kt` to prevent file collisions. It trims whitespace, adds safe fallbacks (using the formatted chapter number or index) when a chapter name is missing, and moves the scanlator name to brackets at the end of the filename to maintain sequential file sorting on disk. Additionally, it sanitizes invalid OS characters with underscores to prevent download failures across different operating systems.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
@@ -71,13 +71,46 @@ object ChapterDownloadHelper {
                     .select(ChapterTable.columns + MangaTable.columns)
                     .where { ChapterTable.id eq chapterId }
                     .firstOrNull() ?: throw IllegalArgumentException("ChapterId $chapterId not found")
+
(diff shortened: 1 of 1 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/ChapterDownloadHelper.kt
```diff
@@ -71,13 +71,46 @@ object ChapterDownloadHelper {
...
+                    if (scanlatorName != null) {
+                        append(" [")
+                        append(scanlatorName)
+                        append("]")
+                    }
+                    append(".cbz")
+                }
+
+            // Sanitize filename for OS compatibility
+            val safeFileName = fileName.replace(Regex("[\\\\/:*?\"<>|]"), "_")
```
Syer10:
```suggestion
            val safeFileName = SafePath.buildValidFilename(fileName)
```

Syer10 (changes requested):

Syer10 (approved):

PR #2103, opened 2026-06-09
Add SyncYomi trigger settings and chapter sync hooks

dodmann:
## Summary

- Add `syncOnChapterRead`, `syncOnChapterOpen`, `syncOnWebUIStart` and `syncOnWebUIResume` boolean settings to `ServerConfig` (proto numbers 96–99, group `SYNCYOMI`, privacy-safe)
- Trigger a SyncYomi sync automatically in `ChapterMutation` when a chapter is marked as read (`isRead == true`) or when a chapter is opened in the reader
- Improve SyncYomi push log messages: clarify success outcome and 412 precondition-failed scenario with descriptive context instead of bare debug messages

## Test plan

- [ ] Configure SyncYomi in server settings and enable `syncOnChapterRead`
- [ ] Mark a chapter as read via the API — verify sync is triggered
- [ ] Enable `syncOnChapterOpen` and open a chapter — verify sync is triggered
- [ ] Confirm `syncOnWebUIStart` and `syncOnWebUIResume` settings are exposed in the GraphQL API
- [ ] Check server logs to confirm the improved push messages appear on success and 412 responses


Syer10:
@schroda Do you want to implement sync triggers in WebUI instead of the server? I feel like that would be better since it would give the UI's control of this.

schroda:
I agree, should be done client side (which includes the koreader sync stuff as well, which is currently done the same way this pr intends to implement it).

Though, as of now, the client can only trigger a sync which runs in the background.
Which makes pushing changes to the sync server the only thing that makes sense.
I don't see what pulling would achieve, especially considering that the caused mutations never reach the client anyway (which is an overall issue with the sync implementation).

It also seems like a big overhead to trigger a full sync for everything, instead of just syncing the specific chapter/manga/whatever.
Especially considering that it uses the backup logic, which, in case postgres is used, just takes forever.
Though, I have no clue if syncyomi even supports anything besides a full sync

dodmann:
Syncyomi doesn't support partial syncs, AFAIK, this is the current implementation for the tachiyomi/komikku, etc. I just tried to replicate it. I don't see a way to trigger the syncs without doing it server-side as well (I've done the WebUI implementation as well), the interface just gives the user more control over the sync triggers. I personally use this as a mean to sync my android devices with my suwayomi, as my library is quite large.

PR #2110, opened 2026-06-15
add reportSyncEvent for SyncYomi service

herowinb:
Since SyncYomi service run on Suwayomi don't have notification so the only way to know when it run success or have problem is look at the logs.
I add reportSyncEvent like TachiyomiSY, so it can sent status to SyncYomi and from SyncYomi sent notification to other like Discord.
image


Syer10 (changes requested):
You need to run `./gradlew ktlintformat`

herowinb:
> You need to run `./gradlew ktlintformat`

Thanks, I not really familiar with Kotlin.

Syer10 (approved):

PR #2113, opened 2026-06-15
Fix Postgres Backups

Syer10:
ChapterTable.toDataClass seems to cause quite slowdown when using Postgres, I am thinking of refactoring Suwayomi in the future to remove excess usages of the .toDataClass functions, as they were made for the V1 API.


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - (**API**) Fix GraphQL `Filter` `notAll` and `notAny` being inversed
 - (**API**) Fix GraphQL `Filter` causing an UnsupportedOperationException when passing an empty list as a `Any` filter value
 - (**Build**) Fix CURL failing silently in builds
+- (**Backup/Database**) Fix backup creation slowdown when using PostgreSQL
```
schroda: It's not just postgres, H2 as well (~2k entries goes down to ~3s instead of ~40s)

schroda (approved):

PR #2120, opened 2026-06-17
Extension API 1.6

Syer10:
I need testers to make sure everything works properly, I can provide a jar to anyone that needs one. @Robonau @schroda @cpiber 

PRs to this branch would be welcome as well. I am sure I missed some things.

Currently trying to get SourceTable.message and SourceTable.contentRating to be filled properly somehow.

Here is a test extension using the 1.6 Extension API:
[tachiyomi-en.asurascans-v1.6.62-release.apk.zip](https://github.com/user-attachments/files/29028394/tachiyomi-en.asurascans-v1.6.62-release.apk.zip)


docs/Configuring-Suwayomi‐Server.md
```diff
@@ -159,15 +159,20 @@ server.systemTrayEnabled = true
...
 - `server.maxLogFileSize = "10mb"` sets the maximum size of a log file - values are formatted like: 1 (bytes), 1KB (kilobytes), 1MB (megabytes), 1GB (gigabytes)
 - `server.maxLogFolderSize = "100mb"` sets the maximum size of all saved log files - values are formatted like: 1 (bytes), 1KB (kilobytes), 1MB (megabytes), 1GB (gigabytes)
-- `server.extensionRepos` is a list of extension repositories for custom sources. Uses the same format as Mihon; each entry is expected to be a string URL pointing to a JSON file representing the repository.
+
+### Extension/Source
+```
+server.extensionStores = []
+server.maxSourcesInParallel = 6
+```
+- `server.extensionStores` is a list of extension stores(previously called repositories) for custom sources. Uses the same format as Mihon; each entry is expected to be a string URL pointing to a JSON or PROTOBUF file representing the repository.
```
cpiber:
```suggestion
- `server.extensionStores` is a list of extension stores (previously called repositories) for custom sources. Uses the same format as Mihon; each entry is expected to be a string URL pointing to a JSON or PROTOBUF file representing the repository.
```

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/SourceQuery.kt
```diff
@@ -118,7 +125,7 @@ class SourceQuery {
                 andFilterWithCompareEntity(SourceTable.id, id),
                 andFilterWithCompareString(SourceTable.name, name),
                 andFilterWithCompareString(SourceTable.lang, lang),
-                andFilterWithCompare(SourceTable.isNsfw, isNsfw),
+                // andFilterWithCompareEnum(SourceTable.contentRating, contentRating)
```
cpiber: Seems unfinished

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/SourceQuery.kt
```diff
@@ -91,14 +93,17 @@ class SourceQuery {
...
         val isNsfw: Boolean? = null,
+        val contentRating: ContentRating? = null,
     ) : HasGetOp {
         override fun getOp(): Op<Boolean>? {
             val opAnd = OpAnd()
             opAnd.eq(id, SourceTable.id)
             opAnd.eq(name, SourceTable.name)
             opAnd.eq(lang, SourceTable.lang)
-            opAnd.eq(isNsfw, SourceTable.isNsfw)
+            opAnd.andWhere(isNsfw) { if (it) SourceTable.contentRating greaterEq 3 else SourceTable.contentRating less 3 }
```
cpiber: This seems inconsistent with the `GraphQLDeprecated` warning and how other calls translate it, I think a simple `==`/`!=` would be better (or adjust e.g. `ExtensionType.kt`)

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
```diff
@@ -41,35 +42,31 @@ class SourceType(
...
         name = row[SourceTable.name],
         lang = row[SourceTable.lang],
-        iconUrl = Extension.getExtensionIconUrl(sourceExtension[ExtensionTable.apkName]),
+        message = row[SourceTable.message],
+        contentRating = ContentRating.valueOf(row[SourceTable.contentRating]),
+        iconUrl = Extension.proxyExtensionIconUrl(sourceExtension[ExtensionTable.pkgName]),
         supportsLatest = catalogueSource.supportsLatest,
         isConfigurable = catalogueSource is ConfigurableSource,
-        isNsfw = row[SourceTable.isNsfw],
+        isNsfw = row[SourceTable.contentRating] >= 3,
```
cpiber: Same as above, this is inconsistent

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Source.kt
```diff
@@ -49,10 +49,10 @@ object Source {
                     id = it[SourceTable.id].value.toString(),
                     name = it[SourceTable.name],
                     lang = it[SourceTable.lang],
-                    iconUrl = getExtensionIconUrl(sourceExtension[ExtensionTable.apkName]),
+                    iconUrl = proxyExtensionIconUrl(sourceExtension[ExtensionTable.pkgName]),
                     supportsLatest = catalogueSource.supportsLatest,
                     isConfigurable = catalogueSource is ConfigurableSource,
-                    isNsfw = it[SourceTable.isNsfw],
+                    isNsfw = it[SourceTable.contentRating] >= 3,
```
cpiber: Inconsistent

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Source.kt
```diff
@@ -70,13 +70,10 @@ object Source {
...
                 lang = source[SourceTable.lang],
-                iconUrl =
-                    getExtensionIconUrl(
-                        extension[ExtensionTable.apkName],
-                    ),
+                iconUrl = proxyExtensionIconUrl(extension[ExtensionTable.pkgName]),
                 supportsLatest = catalogueSource.supportsLatest,
                 isConfigurable = catalogueSource is ConfigurableSource,
-                isNsfw = source[SourceTable.isNsfw],
+                isNsfw = source[SourceTable.contentRating] >= 3,
```
cpiber: As above

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/ExtensionStoreService.kt
```diff
@@ -0,0 +1,223 @@
...
+        return try {
+            val response = network.client.newCall(GET(indexUrl)).awaitSuccess()
+            response.body
+                .source()
+                .use { source ->
+                    try {
+                        protoBuf.decodeFromByteArray<NetworkExtensionStore>(source.peek().readByteArray())
+                    } catch (e: IllegalArgumentException) {
+                        logger.debug { "Failed to decode as protobuf, trying JSON" }
+                        if (forceV2) throw e
```
cpiber: This `if` seems to be too early, V2 should allow json, just not the old format, no?

Mihon also has it later: https://github.com/mihonapp/mihon/blob/6552ffe3158f8127ca265f8621ad326be9e173cb/data/src/main/java/mihon/data/extension/service/ExtensionStoreService.kt#L41

server/src/main/kotlin/eu/kanade/tachiyomi/source/model/SMangaImpl.kt
```diff
@@ -2,24 +2,28 @@
...
 
-    override var status: Int = 0
-
-    override var thumbnail_url: String? = null
-
     override var update_strategy: UpdateStrategy = UpdateStrategy.ALWAYS_UPDATE
 
     override var initialized: Boolean = false
+
+    override var memo: JsonObject = JsonObject(emptyMap())
```
cpiber: Why not use `JsonObject.EMPTY` here?

server/src/main/kotlin/eu/kanade/tachiyomi/source/model/SChapterImpl.kt
```diff
@@ -2,14 +2,18 @@
...
 
-    override var date_upload: Long = 0
-
     override var chapter_number: Float = -1f
 
     override var scanlator: String? = null
+
+    override var date_upload: Long = 0
+
+    override var memo: JsonObject = JsonObject(emptyMap())
```
cpiber: Why not use `JsonObject.EMPTY` here?

cpiber (changes requested):
From checking the code only, will do more testing now

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -276,30 +276,38 @@ class ServerConfig(
...
-            } else {
-                null
-            }
-        },
+        deprecated =
+            SettingsRegistry.SettingDeprecated(
+                replaceWith = "extensionStores",
+                message = "Replaced with extensionStores",
+                migrateConfigValue = {
+                    (it.unwrapped() as? List<String>)
```
cpiber, line 290:
```
ServerConfig.kt:290:37 Unchecked cast of 'Any!' to 'List<String>'.
```

cpiber (changes requested):
Minor thing, the setting migration introduces a new warning that could be suppressed

The last thing is the unfinished filter above, is that intentionally left out for the moment? Otherwise LGTM

cpiber:
Actually, there seems to be some syncing problem of the extensions. I changed the repo list via GQL (WebUI preview), but that was not reflected in the server.conf nor was the updated repo fetched. This doesn't always happen, and may not be related to this PR at all. Specifically, this seems to happen if WebUI is open, server is restarted, and WebUI is then used to change the settings without reload the page, and without opening the Browse > Extensions page. As soon as the page is refreshed, the repo is queried and extensions are populated, ~~until then no extensions are available (all obsolete)~~[^1] (and also the server.conf is not updated).

[^1]: This seems inconsistent, I cannot get it to reproduce reliably, sometimes opening the Browse > Extensions page causes the list to be refreshed (but not server.conf), sometimes all are obsolete

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ExtensionStoreMutation.kt
schroda, line 1: Are you supposed to use these and updating it via the settings mutation is wrong? Shouldn't the list setting then be exluded from the update mutation? What happens if one uses the reset setting mutation?

The mutations here also need to return the settings since they are mutating them

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/ExtensionQuery.kt
```diff
@@ -89,29 +95,44 @@ class ExtensionQuery {
...
             opAnd.eq(name, ExtensionTable.name)
+            opAnd.eq(extensionLib, ExtensionTable.extensionLib)
             opAnd.eq(versionName, ExtensionTable.versionName)
-            opAnd.eq(versionCode, ExtensionTable.versionCode)
+            opAnd.eq(versionCode?.toLong(), ExtensionTable.versionCode)
+            opAnd.eq(versionCodeLong, ExtensionTable.versionCode)
             opAnd.eq(lang, ExtensionTable.lang)
-            opAnd.eq(isNsfw, ExtensionTable.isNsfw)
+            opAnd.eq(
+                isNsfw?.let { if (it) ContentRating.PORNOGRAPHIC.ordinal else ContentRating.SAFE.ordinal },
```
schroda: `erotica` doesn't count as nsfw?

server/src/main/kotlin/suwayomi/tachidesk/manga/model/dataclass/ExtensionInfo.kt
```diff
@@ -0,0 +1,38 @@
...
+data class ExtensionSource(
+    val id: Long,
+    val name: String,
+    val lang: String,
+    val homeUrl: String,
+    val message: String?,
+    val contentRating: ContentRating,
+)
+
+enum class ContentRating {
```
schroda: Why does this enum exist multiple times?

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/ExtensionStoreType.kt
```diff
@@ -0,0 +1,84 @@
...
+        name = row[ExtensionStoreTable.name],
+        badgeLabel = row[ExtensionStoreTable.badgeLabel],
+        signingKey = row[ExtensionStoreTable.signingKey],
+        contactWebsite = row[ExtensionStoreTable.contactWebsite],
+        contactDiscord = row[ExtensionStoreTable.contactDiscord],
+        indexUrl = row[ExtensionStoreTable.indexUrl],
+        isLegacy = row[ExtensionStoreTable.isLegacy],
+    )
+
+    fun extension(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<ExtensionNodeList> =
```
schroda:
```suggestion
    fun extensions(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<ExtensionNodeList> =
```

schroda (commented):
Feels like the settings are now misused to store the extension store urls

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ExtensionStoreMutation.kt
cpiber: The settings and the DB are being synced on any change. I agree that it is weird to have two places, but the config is the only thing the user can modify without a UI, so making the list of extension repos configurable is kinda important IMO.

Alternatively, the settings mutation could fake the url list setting and only write to DB.

Adding/removing stores via the settings mutation works btw, I am using preview WebUI without problems with this.

schroda: Yeah, I saw that later as well.
But that in itself is a problem as well. Changing the setting causes a side mutation (updating the store) that the client never gets informed about

All this syncing makes [this ](https://github.com/Suwayomi/Suwayomi-Server/pull/2120#pullrequestreview-4519348669) even more problematic.
The extensions urls do already not really belong to settings imo, with these changes this just gets even worse.

I don't really see a problem with extensions stores requiring a UI. You already can't install any extensions without it anyway.

cpiber: The extension repo may provide a new URL regardless, so even if we don't keep the setting, the UI may miss a change as-is. Any regular fetch of the store may cause the update.

schroda: Not sure I can follow, what do you mean the repo can provide a new URL?

When would the repos get fetched/updated if not triggered by the user? Which would be done through a mutation that returns the mutated data

cpiber: When the repo updates to the new store format, they keep the existing index.min.json but the new store URL is automatically detected and preferred.

I was relatively certain that extension updates are checked periodically, I'm not at my PC so won't check now. If it is only user interaction, then I agree, having a single place makes it easier, but what about compatibility with older clients?
For that matter, mutations by other clients don't notify your client in any case, so WebUI can just choose to ignore the setting and use the dedicated mutation always.
Gn8 will respond tomorrow to further comments

Syer10: Lets exclude the extensionStores setting in the GQL API, so it can only be updated in the one place.

I've added ExtensionStores to the result of fetchExtensions since that is where it will be updated if the url changes.

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/ExtensionQuery.kt
Syer10: This isn't the final version of the enum, the final version will mean that the last one equals NSFW.

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -276,30 +276,39 @@ class ServerConfig(
...
+                                        (it.groupValues.getOrNull(5)?.ifBlank { null } ?: "index.min.json")
+                                }
+                            } else {
+                                it
+                            }
+                        }
+                },
+            ),
+        readMigrated = { extensionStores.value },
+        setMigrated = { extensionStores.value = (extensionStores.value + it).distinct() },
```
cpiber:
```suggestion
        setMigrated = { extensionStores.value = it?.distinct().orEmpty() },
```

cpiber (changes requested):

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -276,30 +276,39 @@ class ServerConfig(
...
-        itemToValidValue = { url ->
-            if (url.matches(repoMatchRegex)) {
-                url
-            } else {
-                null
-            }
-        },
+        deprecated =
+            SettingsRegistry.SettingDeprecated(
+                replaceWith = "extensionStores",
```
cpiber: Should `replaceWith` still be set, since the setting is not offered via GQL? More elegant would be to indicate the relevant mutations

cpiber (approved):

Syer10:
I'll leave this open until a final version of the Extension API is out, most of the changes should be done and further updates should be minimal unless someone finds other issues.

Syer10:
I'll be merging tomorrow so be sure to get any last minute reviews in if you need to @schroda @cpiber

cpiber (approved):

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ExtensionStoreDataLoader.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.exposed.v1.jdbc.transactions.transaction
+import suwayomi.tachidesk.graphql.types.ExtensionNodeList
+import suwayomi.tachidesk.graphql.types.ExtensionNodeList.Companion.toNodeList
+import suwayomi.tachidesk.graphql.types.ExtensionStoreType
+import suwayomi.tachidesk.graphql.types.ExtensionType
+import suwayomi.tachidesk.manga.model.table.ExtensionStoreTable
+import suwayomi.tachidesk.manga.model.table.ExtensionTable
+import suwayomi.tachidesk.server.JavalinSetup.future
+
+class ExtensionStoreDataLoader : KotlinDataLoader<String, ExtensionStoreType> {
```
schroda, line 19: What is the difference between `ExtensionStoreDataLoader` and `ExtensionStoreForExtension`? Looks identical

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ExtensionStoreDataLoader.kt
```diff
@@ -0,0 +1,77 @@
...
+                            .where { ExtensionStoreTable.indexUrl inList ids }
+                            .map { ExtensionStoreType(it) }
+                            .associateBy { it.indexUrl }
+                    ids.map { (extensionStoreByIndexUrl[it]) }
+                }
+            }
+        }
+}
+
+class ExtensionForExtensionStore : KotlinDataLoader<String, ExtensionNodeList> {
```
schroda:
```suggestion
class ExtensionsForExtensionStore : KotlinDataLoader<String, ExtensionNodeList> {
```

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ExtensionStoreDataLoader.kt
```diff
@@ -0,0 +1,77 @@
...
+    override val dataLoaderName = "ExtensionForExtensionStore"
```
schroda:
```suggestion
    override val dataLoaderName = "ExtensionsForExtensionStore"
```

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0057_AddNewExtensionApiFields.kt
```diff
@@ -0,0 +1,88 @@
...
+import suwayomi.tachidesk.manga.model.dataclass.ContentWarning
+import suwayomi.tachidesk.server.database.migration.helpers.MAYBE_TYPE_PREFIX
+import suwayomi.tachidesk.server.database.migration.helpers.UNLIMITED_TEXT
+import suwayomi.tachidesk.server.database.migration.helpers.toSqlName
+import suwayomi.tachidesk.server.serverConfig
+
+@Suppress("ClassName", "unused")
+class M0057_AddNewExtensionApiFields : SQLMigration() {
+    fun postgresRename(): String =
+        "ALTER TABLE EXTENSION RENAME COLUMN " + "repo".toSqlName() + " TO " + "store_index_url".toSqlName() + ";"
```
schroda, line 21: Should we create a index for the `store_index_url` since we query extensions for stores?

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0058_AddExtensionStore.kt
```diff
@@ -0,0 +1,32 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import de.neonew.exposed.migrations.helpers.AddTableMigration
+import org.jetbrains.exposed.v1.core.Table
+import org.jetbrains.exposed.v1.core.dao.id.IntIdTable
+
+@Suppress("ClassName", "unused")
+class M0058_AddExtensionStore : AddTableMigration() {
+    private class ExtensionStoreTable : IntIdTable() {
+        val indexUrl = varchar("index_url", 2048)
```
schroda: Should `indexUrl` have a unique constraint?

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
```diff
@@ -519,4 +520,12 @@ fun applicationSetup() {
     GlobalScope.launch {
         CEFManager.init()
     }
+
+    serverConfig.subscribeTo(
```
schroda, line 524: Is this subscription still needed? I think on startup the sync would have to be run only once to sync any potential changes not done via the gql mutations.
Otherwise, it would run the prefs to db sync unnecessarily - gql mutation -> updates db and pref -> subscription triggered -> updates db

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ExtensionStoreMutation.kt
schroda: I still would just remove them from the settings completely once the deprecated extensionRepo gets dropped

server/src/main/kotlin/suwayomi/tachidesk/server/ServerSetup.kt
cpiber: It is still needed in case one uses the old settings mutation (e.g. stable WebUI).
For the new mutation, the subscription executes but does nothing, since the DB is already up-to-date.

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ExtensionStoreDataLoader.kt
Syer10: Removed ExtensionStoreForExtension

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ExtensionStoreDataLoader.kt
Syer10: Fixed

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0058_AddExtensionStore.kt
Syer10: Done

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ExtensionStoreDataLoader.kt
Syer10: Done

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0057_AddNewExtensionApiFields.kt
Syer10: Added

Syer10:
@schroda last chance for any final reviews. Will merge in 30 mins

PR #2182, opened 2026-07-13
Add ability to install jar versions of extensions

Syer10:


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
```diff
@@ -48,241 +49,421 @@ import suwayomi.tachidesk.manga.model.table.ExtensionTable
...
-                val extensionLang =
-                    when (langs.size) {
-                        0 -> ""
-                        1 -> langs.first()
-                        else -> "all"
-                    }
+            try {
+                uninstallExtension(pkgName)
+            } catch (_: Throwable) {
+            }
```
cpiber, line 383: By removing the jar before calling `uninstallExtension` you're causing it to not execute `unregisterSource`, is that intentional?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/AndroidManifestParser.kt
cpiber, line 1: For future reference: ApkParsers is deprecated (https://github.com/hsiafan/apk-parser is archived since 2020), so it might be good to change the apk path to also use this class

cpiber (approved):
LGTM

Haven't tested manual jar upload since WebUI doesn't support it, but I assume that should also work

Syer10:
> Haven't tested manual jar upload since WebUI doesn't support it, but I assume that should also work

Yep, tested with a Postman alternative, here is the relevant CURL
```
curl --request POST \
  --url http://localhost:4567/api/graphql \
  --header 'content-type: multipart/form-data' \
  --form 0=@0 \
  --form 'operations={"query": "mutation MyMutation($extensionFile: Upload!) { installExternalExtension(input: {extensionFile: $extensionFile}) { extension { apkName } } }", "variables": {"extensionFile": null}}' \
  --form 'map={"0": ["variables.extensionFile"]}'
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/AndroidManifestParser.kt
Syer10: I will have to take a look at this at some point, sadly the AndroidManifest in a apk is not a standard xml, but a binary mapped axml type. Its much more complicated to parse properly.

cpiber: Ah, I didn't check properly, so the new lib `com.github.Aliucord:axml` isn't actually used, remove it then?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
Syer10: I believe that's something that @schroda's PR will fix. I won't change it for now since that's how it behaved previously.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/AndroidManifestParser.kt
Syer10: It actually is used, it provides the `ArscParser` used in `ResourceArscIconParser`

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
cpiber: Criticizing my own code xD sure!

schroda (approved):

Issue #1181, opened 2024-12-04
[Bug] AniList Json error when sending tracking information


WarDrake:
## Device information
- Suwayomi-Server version: v1.1.1-r1636
- Server Operating System: Proxmox 8.0.4
- Server Desktop Environment: N/A
- Server JVM version: bundled with docker Image
- Client Operating System: Windows 11
- Client Web Browser: Google Chrome 131.0.6778.85

## Steps to reproduce
1. Login to AniList
2. Read a chapter

### Expected behavior
The chapter should be sent over to the Anilist API

### Actual behavior
The chapter is sent but a JSON error prevents it from working

## Other details
Here's the docker log for the action
suwayomi-1    | 00:19:45.216 [DefaultDispatcher-worker-8] ERROR suwayomi.tachidesk.manga.impl.track.tracker.anilist.Anilist -- loadOAuth err
suwayomi-1    | kotlinx.serialization.json.internal.JsonDecodingException: Expected start of the object '{', but had 'EOF' instead at path: $
suwayomi-1    | JSON input:
suwayomi-1    |         at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
suwayomi-1    |         at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:32)
suwayomi-1    |         at kotlinx.serialization.json.internal.AbstractJsonLexer.fail(AbstractJsonLexer.kt:580)
suwayomi-1    |         at kotlinx.serialization.json.internal.AbstractJsonLexer.fail$default(AbstractJsonLexer.kt:578)
suwayomi-1    |         at kotlinx.serialization.json.internal.AbstractJsonLexer.fail$kotlinx_serialization_json(AbstractJsonLexer.kt:231)
suwayomi-1    |         at kotlinx.serialization.json.internal.AbstractJsonLexer.fail$kotlinx_serialization_json$default(AbstractJsonLexer.kt:226)
suwayomi-1    |         at kotlinx.serialization.json.internal.AbstractJsonLexer.unexpectedToken(AbstractJsonLexer.kt:223)
suwayomi-1    |         at kotlinx.serialization.json.internal.StringJsonLexer.consumeNextToken(StringJsonLexer.kt:79)
suwayomi-1    |         at kotlinx.serialization.json.internal.StreamingJsonDecoder.beginStructure(StreamingJsonDecoder.kt:102)
[...]

PR #2184, opened 2026-07-14
Fix/shared preferences saving corrupting xml file

### LINKED ISSUE #2179: Download queue lost on hard stop: queue state file is truncate-written non-atomically on every queue update
### Device information
- Server version: v2.3.2238 (Stable, r2238)
- Server platform: Docker (ghcr image) on Synology DSM 7.x, JVM/Linux
- Client: web UI / GraphQL

### Symptom

The entire download queue is lost when the server process is killed while downloads are active. In our case a routine container recreate (image update) wiped a 2,224-chapter queue: after restart the queue was empty and `restoreAndResumeDownloads()` had nothing to restore.

### Root cause: queue persistence is a non-atomic truncate-write

The download queue is persisted through the SharedPreferences compat shim into
`settings/suwayomi.tachidesk.manga.impl.download.DownloadManager.xml`.

`DownloadManager.saveDownloadQueue()` fires on effectively every queue mutation (enqueue/dequeue/reorder/download-finished via `saveQueueFlow`), and each save runs this write in `JavaSharedPreferences`'s `PropertiesSettings.onModify`:

```kotlin
file.outputStream().use {
    properties.storeToXML(it, null)
}
```

`file.outputStream()` truncates the file in place before `storeToXML` streams the new content — there is no write-to-temp + atomic rename. With a large queue the file spends most of wall-clock time in a truncated/partial state while downloads are running (we measured the file at 0 bytes roughly 80% of the time during active downloading, polling once per second on a ~2.2k-item queue).

Any hard stop in that window — `docker stop` timeout escalating to SIGKILL, OOM kill, power loss, host crash — persists an empty or truncated XML. On next boot the queue loads as empty (or fails to parse and is discarded by the catch in the loader), and the queue is gone.

The adjacent branch has the same hazard in reverse: `if (properties.isEmpty) file.deleteIfExists()` can drop the file entirely on a transiently-empty snapshot.

### Suggested fix

Standard atomic-persist pattern in the `onModify` writer:

1. Write to a sibling temp file (`<name>.xml.tmp`) in the same directory.
2. Flush + close.
3. `Files.move(tmp, file, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING)`.

That guarantees readers (including the next boot) only ever see either the previous complete snapshot or the new complete snapshot, never a truncated one. Optionally debounce the save flow — persisting the full set on every downloaded chapter is a lot of churn for large queues — but atomicity is the correctness fix; debouncing is just a nicety.

This would apply to all shared-prefs-backed settings written through `JavaSharedPreferences`, not only the download queue — the queue is just the file with by far the highest write frequency, so it's the one that gets caught.

### Workarounds we run meanwhile (in case they help others)

- `stop_grace_period: 60s` on the container + always stopping the downloader and waiting for the state file to settle before any restart/update.
- A rebuild script that re-enqueues from library-not-downloaded chapter state after a wipe (recovered 2,222 of the 2,224 lost items).

schroda:
fixes #2179, #1181

### DIFF
--- a/AndroidCompat/src/main/java/xyz/nulldev/androidcompat/androidimpl/CustomContext.java
+++ b/AndroidCompat/src/main/java/xyz/nulldev/androidcompat/androidimpl/CustomContext.java
@@ -177,7 +177,7 @@ public class CustomContext extends Context {
     public SharedPreferences getSharedPreferences(@NotNull File file, int mode) {
         String path = file.getAbsolutePath().replace('\\', '/');
         int firstSlash = path.indexOf("/");
-        return new JavaSharedPreferences(path.substring(firstSlash));
+        return getSharedPreferences(path.substring(firstSlash), mode);
     }
 
     @Override
--- a/AndroidCompat/src/main/java/xyz/nulldev/androidcompat/io/sharedprefs/JavaSharedPreferences.kt
+++ b/AndroidCompat/src/main/java/xyz/nulldev/androidcompat/io/sharedprefs/JavaSharedPreferences.kt
@@ -57,17 +57,19 @@ class JavaSharedPreferences(
         PropertiesSettings(
             properties,
             onModify = { properties ->
-                try {
-                    if (properties.isEmpty) {
-                        file.deleteIfExists()
-                    } else {
-                        file.createParentDirectories()
-                        file.outputStream().use {
-                            properties.storeToXML(it, null)
+                synchronized(key) {
+                    try {
+                        if (properties.isEmpty) {
+                            file.deleteIfExists()
+                        } else {
+                            file.createParentDirectories()
+                            file.outputStream().use {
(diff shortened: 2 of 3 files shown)


Syer10 (commented):
What does this change in general? Around the app we use preferences and I don't think I kept track of if I used .apply() or .commit(). Do we need to verify existing usages?

schroda:
It prevents file corruption by
- synchronizing file writes
- writing the changes to a tmp file first

It implements `commit` and `apply` properly by making
- `commit` write the edits to the memory and the file synchronously
- `apply` write the edits immediately in the memory and asynchronously to the file

We are using `apply` everywhere. So considering the intent of it, if at all, the change should be aligning the usages with the actual expectation of how the function should behave.

Syer10 (approved):

PR #2187, opened 2026-07-21
perf: Add chapter indexes and optimize data loaders

Seanstoppable:
## Summary

Adds database indexes on ChapterTable/PageTable and rewrites 6 chapter data loaders to use SQL window functions instead of in-memory grouping.

### Database indexes (M0060)
- `chapter(manga, source_order)` — covers sorted chapter listing
- `chapter(manga, read, source_order)` — covers unread chapter queries
- `chapter(manga, last_read_at DESC)` — covers "last read" lookups
- `chapter(manga, fetched_at DESC, source_order DESC)` — covers "latest fetched"
- `chapter(manga, date_upload DESC, source_order DESC)` — covers "latest uploaded"
- `page(chapter)` — covers page lookups by chapter

### Data loader optimization
Rewrote 6 data loaders (`LastReadChapter`, `LatestReadChapter`, `LatestFetchedChapter`, `LatestUploadedChapter`, `FirstUnreadChapter`, `HighestNumberedChapter`) to use `ROW_NUMBER() OVER(PARTITION BY manga ...)` instead of fetching all chapters into memory and grouping with Kotlin's `groupBy`.

**Before:** For N manga, fetches ALL matching chapters into memory, groups by manga, picks first from each group.  
**After:** SQL returns exactly 1 row per manga via window function — the DB does the filtering.

### Code quality
- Introduced `sealed interface ChapterFilter` for type-safe SQL filter construction (no raw string interpolation)
- Added `SortOrder.sql` extension property for reusable sort-order mapping

## Impact
- Reduces memory pressure for large libraries (e.g., 5000+ manga × hundreds of chapters)
- Indexes speed up all paginated chapter queries used by the WebUI
- Compatible with both H2 and PostgreSQL
- This should also address/help #2236 by not loading so many chapters into memory for each dataload

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
@@ -291,7 +291,7 @@ class HighestNumberedChapterForMangaDataLoader : KotlinDataLoader<Int, ChapterTy
  * With appropriate indexes, this executes as an index scan returning only N rows
  * (one per manga) instead of all chapters for the requested manga.
  */
-private fun firstChapterPerManga(
+internal fun firstChapterPerManga(
     mangaIds: List<Int>,
     orderBy: List<Pair<org.jetbrains.exposed.v1.core.Column<*>, SortOrder>>,
     filter: Op<Boolean>? = null,
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
@@ -172,13 +172,12 @@ class LastReadChapterForMangaDataLoader : KotlinDataLoader<Int, ChapterType> {
             future {
                 transaction {
                     addLogger(Slf4jSqlDebugLogger)
-                    val lastReadChaptersByMangaId =
-                        ChapterTable
-                            .selectAll()
-                            .where { (ChapterTable.manga inList ids) }
-                            .orderBy(ChapterTable.lastReadAt to SortOrder.DESC)
-                            .groupBy { it[ChapterTable.manga].value }
-                    ids.map { id -> lastReadChaptersByMangaId[id]?.let { chapters -> ChapterType(chapters.first()) } }
+                    val chaptersByMangaId =
+                        firstChapterPerManga(
+                            mangaIds = ids,
+                            orderBy = listOf(ChapterTable.lastReadAt to SortOrder.DESC),
+                        )
+                    ids.map { chaptersByMangaId[it] }
                 }
             }
         }
@@ -192,13 +191,13 @@ class LatestReadChapterForMangaDataLoader : KotlinDataLoader<Int, ChapterType> {
             future {
                 transaction {
                     addLogger(Slf4jSqlDebugLogger)
-                    val latestReadChaptersByMangaId =
-                        ChapterTable
-                            .selectAll()
-                            .where { (ChapterTable.manga inList ids) and (ChapterTable.isRead eq true) }
-                            .orderBy(ChapterTable.sourceOrder to SortOrder.DESC)
-                            .groupBy { it[ChapterTable.manga].value }
-                    ids.map { id -> latestReadChaptersByMangaId[id]?.let { chapters -> ChapterType(chapters.first()) } }
+                    val chaptersByMangaId =
+                        firstChapterPerManga(
+                            mangaIds = ids,
+                            filter = ChapterFilter.IsRead,
+                            orderBy = listOf(ChapterTable.sourceOrder to SortOrder.DESC),
+                        )
+                    ids.map { chaptersByMangaId[it] }
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


schroda:
exposed supports [window functions](https://www.jetbrains.com/help/exposed/sql-functions.html#window-functions)

Seanstoppable:
@schroda got it, thanks!

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0061_AddChapterIndexes.kt
Syer10, line 1: Rebase on master. We now have a v60 migration so this needs to be renamed.

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
```diff
@@ -269,18 +270,57 @@ class HighestNumberedChapterForMangaDataLoader : KotlinDataLoader<Int, ChapterTy
...
         }
 }
+
+/**
+ * Fetches at most one chapter per manga using Exposed's ROW_NUMBER() window function,
+ * avoiding the previous pattern of loading all chapters and grouping in memory.
+ * With appropriate indexes, this executes as an index scan returning only N rows
+ * (one per manga) instead of all chapters for the requested manga.
+ */
+private fun firstChapterPerManga(
```
Syer10: This function should be more generalized so we could use it in more places with other tables.

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0060_AddChapterIndexes.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (C) Contributors to the Suwayomi project
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import de.neonew.exposed.migrations.helpers.SQLMigration
+
+@Suppress("ClassName", "unused")
+class M0060_AddChapterIndexes : SQLMigration() {
```
schroda: I think we can add indexes for `is_read`, `is_bookmarked` and `is_downloaded` as well

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0061_AddChapterIndexes.kt
```diff
@@ -0,0 +1,23 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import de.neonew.exposed.migrations.helpers.SQLMigration
+
+@Suppress("ClassName", "unused")
+class M0060_AddChapterIndexes : SQLMigration() {
+    override val sql =
+        """
+        CREATE INDEX IF NOT EXISTS chapter_manga_source_order ON CHAPTER (manga, source_order);
```
schroda, line 16: @Syer10 when do we have to care for the `table` and `column` names for the different databases (`toSqlName`)?

Syer10: At this point we don't use any databases where it matters most of the time. I believe PostgreSQL uses lowercase table names and H2 is case insensitive while internally uses full caps. Though I could be misremembering. 

It can matter in certain queries or with certain table names, but we currently don't *need* to use it most of the time.

If we wanted to we could enforce toSqlName for a standardization purposes, but I don't feel like it's needed at this point.

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0060_AddChapterIndexes.kt
Seanstoppable: We currently don't seem to be using these server side. But now I am realizing that most filtering is happening client side, and I am trying to improve performance for listing reasonably sized libraries. Want me to expand this to do filtering server side where we can leverage indexing more and send less data over the wire?

server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/ChapterDataLoader.kt
Seanstoppable: Can you expand on your thought process/example of where else this would be useful? Currently seems to be fairly limited to the manga/chapters experience.

server/src/main/kotlin/suwayomi/tachidesk/server/database/migration/M0060_AddChapterIndexes.kt
schroda: The `MangaType` has `downloadCount`, `unreadCount` and `bookmarkCount` and the webUI also requests filtered chapter ids for actions like download, delete, etc.

Seanstoppable: Ok, indexes added

server/src/main/kotlin/suwayomi/tachidesk/graphql/queries/MangaQuery.kt
```diff
@@ -219,6 +231,11 @@ class MangaQuery {
                 andFilterWithCompare(MangaTable.lastFetchedAt, lastFetchedAt),
                 andFilterWithCompare(MangaTable.chaptersLastFetchedAt, chaptersLastFetchedAt),
                 andFilterWithCompareEntity(CategoryMangaTable.category, categoryId),
+                hasChapterFilter(hasUnreadChapters, ChapterTable.isRead eq false),
```
schroda: I'd say if we add filters like this, it should be `IntFilters`. `BooleanFilter` makes this to restrictive (e.g. not being able to filter for manga that have all chapters downloaded)

Though, besides this, I think this doesn't really belong to this pr and should be its own

Seanstoppable: For clarity, you would prefer me to split apart the graphql changes that would leverage the indexes into a seperate PR? Happy to do so, but want to make sure that is the request.

Syer10: That would be the better option.

Seanstoppable: Done.

PR #2203, opened 2026-07-25
Return 404 for missing downloads

Seanstoppable:
I've been moving around my library, and in some cases it isn't clear to me if an error when trying to load something is because I've moved the files and not migrated in the UI, or something else

My proposal is to return 404s here, instead of 521s, which should also include the reason for clarity

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
@@ -21,7 +21,7 @@ import org.jetbrains.exposed.v1.jdbc.transactions.transaction
 import org.jetbrains.exposed.v1.jdbc.update
 import suwayomi.tachidesk.graphql.types.DownloadConversion
 import suwayomi.tachidesk.manga.impl.util.getChapterCachePath
-import suwayomi.tachidesk.manga.impl.util.source.GetSource.getSourceOrStub
+import suwayomi.tachidesk.manga.impl.util.source.GetSource.getSourceOrNull
 import suwayomi.tachidesk.manga.impl.util.storage.ImageResponse.getImageResponse
 import suwayomi.tachidesk.manga.impl.util.storage.ImageUtil
 import suwayomi.tachidesk.manga.model.table.ChapterTable
@@ -118,7 +118,7 @@ object Page {
             return imageFile.inputStream() to (ImageUtil.findImageType { imageFile.inputStream() }?.mime ?: "image/jpeg")
         }
 
-        val source = getSourceOrStub(mangaEntry[MangaTable.sourceReference])
+        val source = getSourceOrNull(mangaEntry[MangaTable.sourceReference])!!
         source as HttpSource
 
         if (pageEntry[PageTable.imageUrl] == null) {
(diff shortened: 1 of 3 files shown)


Syer10 (commented):
I'm not really sure I see the point of this. We don't support the legacy REST API and everything has moved to GraphQL which doesn't support status codes.

Seanstoppable:
The graphql API for pages returns a list of urls
https://github.com/Suwayomi/Suwayomi-Server/blob/d7027199888bfe051864b3c28fdcb6d3ce320839/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/ChapterMutation.kt#L460

Which calls manga controller
https://github.com/Suwayomi/Suwayomi-Server/blob/master/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/MangaController.kt#L479

Which calls these, and currently triggers 500 errors, rather than 404s with more readable/debuggable error messages

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
```diff
@@ -119,7 +120,9 @@ object Page {
         }
 
         val source = getSourceOrStub(mangaEntry[MangaTable.sourceReference])
-        source as HttpSource
+        if (source !is HttpSource) {
+            throw NotFoundResponse("Source not installed and chapter not downloaded")
+        }
```
Syer10: You will need to manually import getSource
```suggestion
        val source = getSource(mangaEntry[MangaTable.sourceReference])!!
        source as HttpSource
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/ChaptersFilesProvider.kt
```diff
@@ -78,7 +79,7 @@ abstract class ChaptersFilesProvider<Type : FileType>(
         val images = getImageFiles().filter { it.getName() != COMIC_INFO_FILE }.sortedBy { it.getName() }
 
         if (images.isEmpty()) {
-            throw Exception("no downloaded images found")
+            throw NotFoundResponse("no downloaded images found")
```
Syer10:
```suggestion
            throw NoSuchElementException("no downloaded images found")
```

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/fileProvider/impl/FolderProvider.kt
```diff
@@ -34,7 +35,7 @@ class FolderProvider(
         val chapterFolder = File(getChapterDownloadPath(mangaId, chapterId))
 
         if (!chapterFolder.exists()) {
-            throw Exception("download folder does not exist")
+            throw NotFoundResponse("download folder does not exist")
```
Syer10:
```suggestion
            throw NoSuchElementException("download folder does not exist")
```

Syer10 (changes requested):
I see your point, I would prefer not to use Javalin-specific exceptions though. We have NoSuchElementException and NullPointerException mapped to 404 so these should work.

Syer10 (approved):

Issue #2221, opened 2026-07-28
Improve startup times


riyaniii:
### Describe your suggested feature

The server starts up too slow for it to be used to make client apps, about 7-10 seconds, please look into this

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have updated the (**[Suwayomi-WebUI](https://github.com/suwayomi/suwayomi-webui/releases/latest)** and **[Suwayomi-Server](https://github.com/suwayomi/suwayomi-server/releases/latest)**) to the latest versions
- [x] I have filled out all of the requested information in this form, including specific version numbers.

coreofcogwork:
There has already been more than one client app made for suwayomi so this obviously isn't true
[https://github.com/Suwayomi/Tachidesk-Sorayomi](https://github.com/Suwayomi/Tachidesk-Sorayomi)
[https://github.com/Suwayomi/Suwayomi-Tsumiru](https://github.com/Suwayomi/Suwayomi-Tsumiru)

A 7-10 second startup time is miniscule compared to the amount of time someone would spend on an app, unless they are repeatedly  opening the app and closing it again a few seconds later. Startup time also varies greatly depending on hardware / how you run the server, I generally get 4-5 second startup times running it in docker. The devs have definitely spent time trying to make it as quick as possible, just after a certain point it is almost impossible to make spinning up a whole database and server any faster.

riyaniii:
Navidrome, TDLib all startup in milliseconds, maybe you can look into them

coreofcogwork:
I was bored so decided to look through logs to see what is taking time for me
~0.5s: starting up android main loop
~1s: loading settings, connecting to database, performing migrations/validating database, loading queued downloads, setting up cron jobs
~2s: setting up chromium embedded framework (seems to be overlapped with the next two things)
~0.5s: requesting extension repo and getting response
~0.5s: starting javalin
overlapped with other stuff: ~1s: creating webui folder, verifying install, loading webui (presumably this won't happen if you aren't using webui)

Some delays are from external components starting up (javalin, cef). It feels to me like it ought to be possible to parallelise this more, such as starting cef and requesting the extension repo as soon as settings are loaded, and starting javalin earlier, but of course I don't really know anything about this, I may just be waffling nonsense.

riyaniii:
[...]

PR #2231, opened 2026-07-29
Start graphql schema generation immediately on startup

schroda:
Takes multiple seconds to generate and would otherwise delay the javalin startup, because it gets generated when setting up the api endpoints

improves #2221 (haven't found anything else besides this)

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/Main.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/Main.kt
@@ -7,10 +7,13 @@ package suwayomi.tachidesk
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import suwayomi.tachidesk.graphql.server.GraphQLSchemaProvider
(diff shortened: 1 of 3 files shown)


server/src/main/kotlin/suwayomi/tachidesk/Main.kt
```diff
@@ -7,10 +7,19 @@ package suwayomi.tachidesk
...
+import suwayomi.tachidesk.graphql.server.SchemaProvider
 import suwayomi.tachidesk.server.JavalinSetup.javalinSetup
 import suwayomi.tachidesk.server.applicationSetup
 
 fun main() {
+    // Trigger graphql schema generation as early as possible, takes multiple seconds, delaying the javalin startup otherwise
+    @OptIn(DelicateCoroutinesApi::class)
+    GlobalScope.launch {
+        SchemaProvider.getSchema()
+    }
```
Syer10: Make this happen in a `SchemaProvider.init()` function

Syer10 (changes requested):

Syer10 (approved):

PR #2238, opened 2026-07-30
Fix: Add early return for unchanged directory in updateDir

vtorres-t:
image

When updating manga with downloaded chapters, the "downloaded" status disappears, and all chapters are marked as not downloaded (set to `false`).

This occurs when the remote title and the original title are the same.

This has been fixed by treating the operation as successful when the source and destination directories are identical.


schroda:
There is already a pr for that #2223

PR #2242, opened 2026-08-03
Create db indexes for foreign keys

schroda:


Syer10 (approved):

Syer10 (commented):
Maybe add a changelog

schroda:
Same reason as the privacy safe changes, don't think an internal change is adding much there

PR #2243, opened 2026-08-03
Merge manga chapter stats data loader

schroda:
The bigger the batch of the data loaders, the bigger the performance impact of running a data loader for each chapter stat.

E.g. the webui query to load the mangas of a category with 79 entries takes 169ms compared to 702ms


Syer10 (approved):

Syer10 (commented):
Maybe add a changelog

schroda:
Don't think it's needed - [reason](https://github.com/Suwayomi/Suwayomi-Server/pull/2242#issuecomment-5285244996)

PR #2244, opened 2026-08-03
Update privacy safe flags of settings

schroda:


Syer10 (approved):

Syer10 (commented):
Maybe add a changelog

schroda:
Since it's only internal, I don't think it's needed

PR #2248, opened 2026-08-04
Add batch mutation for reordering chapter downloads

schroda:

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -33,6 +33,7 @@ import org.jetbrains.exposed.v1.core.inList
 import org.jetbrains.exposed.v1.jdbc.select
 import org.jetbrains.exposed.v1.jdbc.selectAll
 import org.jetbrains.exposed.v1.jdbc.transactions.transaction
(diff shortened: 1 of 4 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -478,6 +479,10 @@ object DownloadManager {
         notifyAllClients(false, chapterDownloads.toList().map { DownloadUpdate(DownloadUpdateType.DEQUEUED, it) })
     }
 
+    fun reorder(reorders: List<ChapterDownloadReorder>) {
+        reorders.forEach { reorder(it.chapterId, it.to) }
+    }
```
Syer10: Should this use its own function so that `notifyAllClients` and `triggerSaveDownloadQueue` is not called multiple times?

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
schroda: Aligned it with the way `enqueue` works

Syer10 (approved):
Missed this one, also conflicts

PR #2249, opened 2026-08-04
Feature/gql provide platform information

schroda:

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/server/util/Platform.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/server/util/Platform.kt
@@ -1,40 +1,173 @@
 package suwayomi.tachidesk.server.util
 
-data class OSInfo(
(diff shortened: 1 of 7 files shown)


server/src/main/kotlin/suwayomi/tachidesk/server/util/Platform.kt
```diff
@@ -1,132 +1,225 @@
...
+    private fun matches(value: String): Boolean = aliases.any { value.startsWith(it, ignoreCase = true) }
+
+    companion object {
+        private val types =
+            listOf(
+                ::AMD64,
+                ::I386,
+                ::ARM64,
+                ::ARM,
+            )
```
Syer10, line 222: Maybe add RISC-V, we have provided docker builds of it in the past.

Syer10 (changes requested):

schroda:
This is only tested on windows and linux

Syer10 (approved):

Issue #2241, opened 2026-08-01
Allow cookies obtained from opening the extension in webview to be used in accessing the extension instead of flaresolverr


tubaboy26:
### Describe your suggested feature

MangaFire in particular seems to only serve its captcha when you access the main url. If you access the JSON api, it will respond with the following JSON. There is no captcha and therefore flaresolverr cant do anything with this. The repos solution on Mihon is to have you open the webview and then mihon caches the resulting cookie.

<img width="292" height="101" alt="Image" src="https://github.com/user-attachments/assets/fcc38df7-e1e0-49c4-82e5-96be2d437a5b" />

What would solve this would be replicating the functionality of Mihon and allowing the user to open the website in webview and store the cookies in cache.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have updated the (**[Suwayomi-WebUI](https://github.com/suwayomi/suwayomi-webui/releases/latest)** and **[Suwayomi-Server](https://github.com/suwayomi/suwayomi-server/releases/latest)**) to the latest versions
- [x] I have filled out all of the requested information in this form, including specific version numbers.

schroda:
It works already like that

PR #2251, opened 2026-08-05
Match WebView and network user agents

JoJoKorok:
## Summary

- Configure the JCEF WebView to use the same default User-Agent as network requests.
- Keep transferred Cloudflare clearance cookies associated with the User-Agent that obtained them.
- Add a changelog entry for the WebView behavior change.

## Why

Suwayomi already transfers WebView cookies into the persistent network cookie store. However, Cloudflare clearance cookies are tied to the User-Agent that obtained them. If JCEF and the network client use different User-Agents, a transferred clearance cookie can be rejected.

Related to #2241.

## Testing

- `git diff --check`
- `./gradlew.bat ktlintCheck`
- `./gradlew.bat :server:test :server:shadowJar`
- Manual isolated MangaFire test: JCEF initialized, MangaFire loaded in WebView, and the source reloaded successfully after closing WebView; the corresponding API requests returned HTTP 200.

Cloudflare did not present a challenge on the non-VPN connection during the final manual test, so the complete challenge-to-cookie recovery path could not be reproduced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - (**Source/API**) Fix graphql browse mutation (`fetchSourceManga`) with filters including nested group changes
 - (**HTML**) Fix HTML being sent with gibberish when using non-latin characters
 - (**OPDS**) Fix OPDS search charset
+- (**WebView**) Reuse WebView cookies with the matching network user agent
 
 ## [v2.3.2243] - 2026-07-13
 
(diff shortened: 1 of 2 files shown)


cpiber:
This is already done, in a better way: https://github.com/Suwayomi/Suwayomi-Server/blob/eb2dc0b19a9571b27c02bebc5c883e404b7bd7fb/AndroidCompat/src/main/java/xyz/nulldev/androidcompat/webkit/KcefWebViewProvider.kt#L438

Your version just sets the default user agent, but does not change dynamically

JoJoKorok:
Thank you for pointing that out. I missed the per-request User-Agent handling in KcefWebViewProvider and the dynamic http.agent updates. You're right that my change only sets it during initialization. I'll close this draft.

PR #2252, opened 2026-08-05
Handle preservation of cbz chapter downloads

schroda:
The preservation only worked for normal downloads because of the missing file extension. For cbz it returned early with success result because it couldn't find the downloaded file


Syer10 (approved):

Syer10 (commented):
Add a changelog

schroda:
I don't think we should add fixes for[ preview only changes](https://github.com/Suwayomi/Suwayomi-Server/blob/master/CHANGELOG.md?plain=1#L10) to the changelog

PR #2256, opened 2026-08-07
Make JCEF, system tray, and browser launch opt-in by default

Sparshagrawal07:
## Summary
- Default `kcefEnabled`, `systemTrayEnabled`, and `initialOpenInBrowserEnabled` to `false` so HTTP/WebUI startup does not load native desktop components.
- Skip all JCEF/`jawt` native loads when KCEF is disabled; warn on macOS if it is enabled.
- Document the macOS Tahoe ARM64 `SIGTRAP`/`EXC_BREAKPOINT` crash in `cef_initialize`, and that existing `server.conf` values are unchanged.

## Context
On macOS 26.x (Tahoe) ARM64, async CEF init after Javalin start terminates the JVM inside `libjcef.dylib` → `cef_initialize`. Crash reports confirm this is native Chromium/JCEF, not Kotlin server code. Linux has `catch_abort.so` for SIGTRAP; macOS does not.

## Test plan
- [ ] Fresh install / no `kcefEnabled` in conf: server starts, log shows CEF skipped, WebUI + `/api/v1/settings/about` work
- [ ] Existing conf with `kcefEnabled=true` still enables CEF (defaults do not rewrite user conf)
- [ ] With `kcefEnabled=false` on macOS Tahoe: process stays alive (no SIGTRAP)
- [ ] Enabling tray / open-in-browser still works when set to `true`

### DIFF
--- a/docs/Configuring-Suwayomi‐Server.md
+++ b/docs/Configuring-Suwayomi‐Server.md
@@ -11,6 +11,9 @@ Suwayomi will create a default configuration file when one doesn't exist, you ca
 ### I am running Suwayomi in a headless environment (docker, NAS, VPS, etc.)
 - Set `server.systemTrayEnabled` to false, it will prevent Suwayomi to attempt to create a System Tray icon.
 - Set `server.initialOpenInBrowserEnabled`to false, it will prevent Suwayomi to attempt to open a browser on startup.
+- Set `server.kcefEnabled` to false, it will prevent native JCEF/Chromium (WebView) initialization. **Required on macOS Tahoe (26.x) ARM64** where `cef_initialize` has been observed to terminate the JVM with `SIGTRAP`/`EXC_BREAKPOINT`.
+
+These three settings default to `false` (opt-in desktop features). Existing `server.conf` files keep whatever values were already written.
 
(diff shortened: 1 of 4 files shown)


Sparshagrawal07:
In MacOS Tahoe Latest release due to EXC_BREAKPOINT (SIGTRAP)
The webUI breaks again and again 
This makes it much more stable for future releases as well as all other desktop releases as well

cpiber:
Setting `systemTrayEnabled` and `initialOpenInBrowserEnabled` to false is a bad idea, because most people use Suwayomi on Windows, and have no way to stop it otherwise (aside from killing via task manager). WebView I can agree with.

schroda:
If the webview is an macos only issue, then it should only be disabled for macos by default and not for everything

Syer10 (changes requested):
1. `kcefEnabled` is needed for many sources to work, currently we should just force disable it for MacOS until we can fix it.
2. `systemTrayEnabled` and `initialOpenInBrowserEnabled` are vital for GUI users and need to be enabled by default. Headless users can disable it themselves if they want to.

PR #2261, opened 2026-08-09
Add a simple mutation to clear cache & cookies

cpiber:
Cloudflare can get into infinite verification loops if cookies are present and the UA string changes, in which case both our cookie store as well as CEF's cache need to be cleared

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/WebviewMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/WebviewMutation.kt
@@ -7,10 +7,10 @@ import suwayomi.tachidesk.server.ApplicationDirs
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 import java.net.CookieHandler
+import kotlin.io.path.ExperimentalPathApi
 import kotlin.io.path.Path
 import kotlin.io.path.deleteRecursively
 import kotlin.io.path.div
-import kotlin.io.path.ExperimentalPathApi
 
 @OptIn(ExperimentalPathApi::class)
 class WebviewMutation {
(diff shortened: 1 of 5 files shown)


server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/WebviewMutation.kt
```diff
@@ -0,0 +1,38 @@
...
+import java.net.CookieHandler
+import kotlin.io.path.ExperimentalPathApi
+import kotlin.io.path.Path
+import kotlin.io.path.deleteRecursively
+import kotlin.io.path.div
+
+@OptIn(ExperimentalPathApi::class)
+class WebviewMutation {
+    private val applicationDirs by lazy { Injekt.get<ApplicationDirs>() }
+    private val cacheDir by lazy { Path(applicationDirs.dataRoot) / "cache" }
```
Syer10: At this point lets make it a shared path in ApplicationDirs

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/WebviewMutation.kt
cpiber: Do you mean a `cacheDir` variable there? and have https://github.com/Suwayomi/Suwayomi-Server/blob/eb2dc0b19a9571b27c02bebc5c883e404b7bd7fb/server/src/main/kotlin/suwayomi/tachidesk/server/util/CEFManager.kt#L131 reference the `cacheDir / "kcef"`?

Syer10: Yes

Syer10 (approved):

schroda:
How about clearing the cookies in case the ua changes?

cpiber:
> How about clearing the cookies in case the ua changes?

The problem is that clearing just the cookies causes a CloudFlare challenge loop, at least just `cookie_store.xml`. So we would need to also delete `cacheDir` like this mutation does, but that could cause unexpected behaviour (since that includes other data like local storage etc).

Also, this would cause a logout on unrelated sources, which would otherwise continue to work.

Syer10 (commented):
Add a changelog

cpiber:
I've placed it under `Extension`, since clearing cookies affects everything, not just WebView, let me know if you think there's a better category

Syer10:
Maybe a Cache/API category would work?

PR #2268, opened 2026-08-11
ref(server): send full `POST` body in `CloudflareInterceptor`

D-Brox:
Instead of sending an empty string. This doesn't affect current behaviour, but allows the use other implementations of the flaresolverr API that accept json in the POST requests, so a few extra extensions that require POST in sources with CF protection.


Syer10 (commented):
~I was under the understanding that Flaresolverr only supports FormBody data, why does this work? Is it branch specific?~

Nvm I read the description again and now see why it works.

Syer10 (approved):

Syer10 (commented):
Add a changelog

PR #2279, opened 2026-08-15
Handle unsupported source search

Seanstoppable:
I've noticed several sources don't support search, and in the UI it looks kind of ugly/like an error exposing the stack trace

This PR makes it return a much better error message.

image

### DIFF
--- a/server/src/main/kotlin/eu/kanade/tachiyomi/source/Source.kt
+++ b/server/src/main/kotlin/eu/kanade/tachiyomi/source/Source.kt
@@ -108,4 +108,19 @@ interface Source {
     fun fetchPageList(chapter: SChapter): Observable<List<Page>> = throw UnsupportedOperationException()
 }
 
+suspend fun Source.searchManga(
+    page: Int,
+    query: String,
+    filters: FilterList,
+): MangasPage =
+    try {
+        getSearchManga(page, query, filters)
+    } catch (e: UnsupportedOperationException) {
+        throw SearchNotSupportedException(e)
+    }
+
+class SearchNotSupportedException(
+    cause: UnsupportedOperationException,
+) : IllegalArgumentException("Search not supported by this source", cause)
+
 // fun Source.icon(): Drawable? = Injekt.get<ExtensionManager>().getAppIconForSource(this)
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SourceMutation.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/SourceMutation.kt
@@ -7,6 +7,7 @@ import androidx.preference.EditTextPreference
 import androidx.preference.ListPreference
 import androidx.preference.MultiSelectListPreference
 import androidx.preference.SwitchPreferenceCompat
+import eu.kanade.tachiyomi.source.searchManga
 import org.jetbrains.exposed.v1.core.LikePattern
 import org.jetbrains.exposed.v1.core.Op
 import org.jetbrains.exposed.v1.core.and
@@ -265,7 +266,7 @@ class SourceMutation {
             val mangasPage =
                 when (type) {
                     FetchSourceMangaType.SEARCH -> {
-                        source.getSearchManga(
+                        source.searchManga(
                             page = page,
                             query = query.orEmpty(),
                             filters = updateFilterList(source, filters),
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/server/TachideskGraphQLServer.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/server/TachideskGraphQLServer.kt
@@ -10,6 +10,8 @@ package suwayomi.tachidesk.graphql.server
 import com.expediagroup.graphql.generator.execution.FlowSubscriptionExecutionStrategy
 import com.expediagroup.graphql.server.execution.GraphQLRequestHandler
 import com.expediagroup.graphql.server.execution.GraphQLServer
+import com.expediagroup.graphql.server.extensions.toGraphQLError
+import eu.kanade.tachiyomi.source.SearchNotSupportedException
 import graphql.ExceptionWhileDataFetching
 import graphql.GraphQL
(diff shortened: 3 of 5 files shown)


schroda:
Isn't the UI showing something like `Exception while fetching data (/fetchSourceManga) : ... [show more]`?
So one has to click on "show more" to see the stack trace.

This is the standard error behavior. Changing it for only one error specifically seems kinda pointless to me.
The reason the stack trace is added in the first place is to be able to spot places where we throw a `Stub!` error in the AndroidCompat, because that made it really cumbersome to find the location (#1319)

Seanstoppable:
In this case, it isn't really exceptional/an error. It is that search isn't supported for the source. Showing a clearer error, rather than forcing someone to double check by expanding a stack trace is, imo, a better UX.
In my experience showing full stack traces is usually the last resort when you aren't otherwise able to classify the error.
Probably some other exceptions where we should do the same (403s from sources? Timeouts? Neither of those really need/benefit from a full stack trace in the UI do they?)

schroda:
How is it not an exception/error? The source does not tell you if it supports the search functionality or not (unlike for `latest`). The only way to find out is literally to call the function and see if you get some response or not.

Each client is free in how it handles the error message. The stack trace is easily hidden (as is it done in the webui) or can also be completely omitted.

Like I already mentioned, the stack trace is a "debugging band aid" because the server throws a bunch of unhelpful exceptions.
So instead of starting to implement some workaround for specific exceptions to omit the stack trace, the thrown exceptions should be updated to have a meaningful message instead. That way the stack trace doesn't need to be added in the first place.

In this case that would mean updating the actual thrown exception instead of adding a new exception and logic to wrap the original one and rethrow it to just make the workaround work.

So the way I see it, this workaround will just lead to more logic being added in the future to handle more specific errors instead of fixing the root cause.

Seanstoppable:
Yeah, you are exactly correct in "The only way to find out is literally to call the function and see if you get some response or not.". So barring going and making plugin implementation changes so we can do something like not even try to call this, or returning a more specific exception in their implementations, doing something like this is the "best" that a client can do.

We can't map ALL UnimplementedMethodExceptions to a known error, we can only map UnimplementedMethodExceptions of this specific function. I.e. it seems straightforward to me that unimplemented is search is not supported. "the thrown exceptions should be updated to have a meaningful message instead. That way the stack trace doesn't need to be added in the first place." is pretty much exactly what I am doing, without going back and trying to change the plugins spec and all implementations. Even if there was a better error message, you'd still need to map it for user friendly UX here.

Also correct that "Each client is free in how it handles the error message.". Other implementors just return things like "No results found" which is also not quite correct.

Particularly for non-technical users (I am assuming technical people are sharing access to non-technical people), I would imagine that non-technical responses are preferable to technical ones.

schroda:
> "the thrown exceptions should be updated to have a meaningful message instead. That way the stack trace doesn't need to be added in the first place." is pretty much exactly what I am doing, without going back and trying to change the plugins spec and all implementations

I was under the impression that we throw the exception

My point is still the same. If you want to improve the returned error, then that should be reusable going forward for **any** exception/error, not just **one specific** one.

How is this going to expand in the future? Add logic to the graphql exception handler for each and every exception so that we do not add a stack trace?
The graphql exception handler doesn't care that the source search wasn't supported. It shouldn't even know about that specific exception.

Seanstoppable:
I mean, the REST API was mapping specific errors so that they would go from a 5xx to a more meaningful error, no?
I'm not sure how this is meaningfully different.

In general, my belief is that showing stack traces to end users are bad (and as an end user they are ugly), and there is a difference between an error that is somewhat expected and/or understood(Unsupported is the APIs only way to indicate search is not supported) vs not (some other random error we don't understand).

If you are saying you'd rather have this be a more general set of errors, like maybe UnretryableException and RetryableException that are rendered better for a larger set of scenarios, that makes sense.

If you are saying you'd rather just have the stack trace, I'll close this out.

Robonau:
> In general, my belief is that showing stack traces to end users are bad (and as an end user they are ugly), and there is a difference between an error that is somewhat expected and/or understood(Unsupported is the APIs only way to indicate search is not supported) vs not (some other random error we don't understand).

as someone that does allot of support for suwayomi a stack trace is far more useful than a simple error that looks nice, when someone has an issue i want to be able to fix it immediately when possible, obfuscating the error does nothing but add the extra work of walking the user through finding the error in logs

Seanstoppable:
@Robonau Yeah, I guess I get not wanting to have people dive into the logs in some cases. 
"Search not supported" doesn't need/deserve a stack trace, imo, and currently I have to open a stack trace to see if that is the case.
A 403 unauthorized also probably doesn't need a stack trace, it would benefit from a clearer unauthorized exception rather than something like "The server returned 403"
500s are likely the same, as that is "The server has an error, retry later" with a retry button, rather than a deep stack trace

Alternately, my problem is that the error message isn't great such that you HAVE to jump into the stack trace for full details and the rough edge I am feeling could possibly be addressed another way.

kptestfu23-oss:
Keeping raw stack traces in the UI because they’re useful for support feels backwards. Users shouldn’t have to read an exception dump just to learn that a source doesn’t support search. That belongs in logs, not in the normal error message.

And saying this should wait for a more generic error system doesn’t really justify keeping the current behavior. That system doesn’t exist right now. This PR fixes a real user-facing issue now.

If the implementation needs to be generalized later, fine. Keeping bad UX until the perfect abstraction exists makes no sense.

Robonau:
> And saying this should wait for a more generic error system doesn’t really justify keeping the current behavior. That system doesn’t exist right now. This PR fixes a real user-facing issue now.

no, it doesn't fix a non issue

kptestfu23-oss:
> no, it doesn't fix a non issue

There is clearly an issue being addressed here: a known, expected condition is currently presented in a way that requires the user to inspect technical details to understand it.

The PR gives that condition an explicit user-facing message while preserving the exception in the server logs.

If the argument is that showing stack traces in the UI is better for support, that's a valid tradeoff to discuss. But saying there is no issue at all ignores the actual UX problem this PR is addressing.

PR #2281, opened 2026-08-15
Pause downloads during extension updates

Seanstoppable:
## Summary

- Pause active downloads while updating extensions.
- Restore the queue only when it was running before the update.
- Add success and failure-path tests for pause/resume behavior.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
@@ -60,6 +60,25 @@ import kotlin.time.Duration.Companion.seconds
 
 private val logger = KotlinLogging.logger {}
 
+internal class DownloadPauseController(
+    private val isStarted: () -> Boolean,
+    private val stop: suspend () -> Unit,
+    private val start: () -> Unit,
+) {
+    suspend fun <T> run(block: suspend () -> T): T {
+        val wasStarted = isStarted()
+
+        try {
+            stop()
+            return block()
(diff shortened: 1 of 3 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -539,6 +558,13 @@ object DownloadManager {
         refreshDownloaders()
     }
 
+    suspend fun <T> withDownloadsPaused(block: suspend () -> T): T =
+        DownloadPauseController(
```
schroda, line 562: Why is this extracted into a class?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
```diff
@@ -539,6 +558,13 @@ object DownloadManager {
         refreshDownloaders()
     }
 
+    suspend fun <T> withDownloadsPaused(block: suspend () -> T): T =
```
schroda, line 561: While this is running it should not be possible to start downloads

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/download/DownloadManager.kt
Seanstoppable: Mostly for testability.

Seanstoppable:
I don't really hit this error myself and don't find myself interested in iterating here.

PR #2284, opened 2026-08-15
config: Fix migration of extensionRepos to extensionStores

berquist:
I noticed that extensions weren't getting picked up when doing the version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - (**WebView**) Fix that UI-WebView did not share user agent (background WebView unchanged)
 - (**WebView**) Fix authentication with subpath option
 - (**Category**) Fix library/category counts not matching the actual number of manga due to duplicate category-manga rows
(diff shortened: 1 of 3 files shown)


schroda:
Can you change it so that `replaceWith` needs to be explicitly set to `null`

schroda:
In case `replaceWith` is null while either `migrateConfig` or `migrateConfigValue` we should also shutdown the server. Currently that is only done in case `replaceWith` is set but neither `migrateConfig` or `migrateConfigValue`

I guess better would be to change it to something like this

```kotlin
sealed interface SettingDeprecated {
    val message: String

    data class MigrateConfigValue(
        val replaceWith: String,
        override val message: String,
        val migrateConfigValue: (value: ConfigValue) -> Any?,
    ) : SettingDeprecated

    data class MigrateConfig(
        val replaceWith: String,
        override val message: String,
        val migrateConfig: (value: ConfigValue, config: ConfigDocument) -> ConfigDocument,
    ) : SettingDeprecated

    data class NoMigration(
        override val message: String,
    ) : SettingDeprecated
}
```

Syer10 (approved):

Syer10:
@schroda That can probably be handled in a separate PR

Issue #1547, opened 2025-07-26
Split tall images


Afam007:
### Describe your suggested feature

Offer an option for splitting tall images into smaller ones on download and also serve to client if possible

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have updated to the **[latest version](https://github.com/suwayomi/suwayomi-server/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.

vrtgs:
I agree, the web-UI with very tall images starts to get choppy and look like a slideshow

Issue #645, opened 2023-08-08
[Feature Request] Tall image splitter for long strip chapters after download


BaccanoMob:
The feature is already in Tachiyomi for better performance. 

Its more beneficial for Suwayomi extension users to read in mobile faster as well as on other devices.

PR #2288, opened 2026-08-17
Add option to split tall images when downloading chapters

yohanim:
## Summary
- Adds a server-side setting, `splitTallImages` (opt-in, defaults to `false`), that splits overly tall "long strip" webtoon pages into several smaller JPEG images right after a page is downloaded.
- Splitting logic (`TallImageSplitter`) is a JVM port of the heuristic used by Mihon's "split tall images" feature: an image is split when its height is more than 3x its width **and** exceeds 5000px, dividing it into equal-height bands (the last band absorbs any remainder).
- Hooked into `Page.getPageImageDownload()`, right after a downloaded page (converted or not) is saved to the chapter's download cache folder.
- No changes were needed to page-count bookkeeping: `ChapterForDownload.getChapterDownloadReady()` already resyncs `Chapter.pageCount` to the number of files actually on disk, so chapters with split pages report the correct page count to clients automatically.

Towards #1547 and #645.

## Scope note
This PR only covers the server side (Kotlin). The Suwayomi-WebUI project (separate repository) will need its own follow-up to expose a toggle for this setting in the Downloads settings screen — the new setting is already exposed automatically through the GraphQL `settings`/`setSettings` API in the meantime.

## Test plan
- [x] Added `TallImageSplitterTest` covering the split-threshold boundary conditions and an end-to-end split of a synthetic 12000px-tall image, verifying output file count, naming, and total height.
- [ ] Could not run a full Gradle build/test suite in the development environment used for this change (unrelated local JDK/Windows networking issue); please run CI/`./gradlew :server:test` to confirm.

### DIFF
--- a/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -219,14 +219,6 @@ class ServerConfig(
         defaultValue = false,
     )
 
-    val splitTallImages: MutableStateFlow<Boolean> by BooleanSetting(
-        protoNumber = 98,
-        group = SettingGroup.DOWNLOADER,
-        privacySafe = true,
-        defaultValue = false,
-        description = "Split long images into smaller pages after downloading a chapter",
-    )
-
     val downloadsPath: MutableStateFlow<String> by PathSetting(
         protoNumber = 16,
         group = SettingGroup.DOWNLOADER,
@@ -1145,6 +1137,14 @@ class ServerConfig(
                 specificType = "List<String>",
             ),
         description = "List of extension store index URLs",
+    )
+
+    val splitTallImages: MutableStateFlow<Boolean> by BooleanSetting(
+        protoNumber = 98,
+        group = SettingGroup.DOWNLOADER,
+        privacySafe = true,
+        defaultValue = false,
+        description = "Split long images into smaller pages after downloading a chapter",
     )
 
     /** ****************************************************************** **/
--- a/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
+++ b/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
@@ -219,6 +219,14 @@ class ServerConfig(
         defaultValue = false,
     )
 
+    val splitTallImages: MutableStateFlow<Boolean> by BooleanSetting(
+        protoNumber = 98,
+        group = SettingGroup.DOWNLOADER,
+        privacySafe = true,
+        defaultValue = false,
+        description = "Split long images into smaller pages after downloading a chapter",
+    )
+
     val downloadsPath: MutableStateFlow<String> by PathSetting(
         protoNumber = 16,
         group = SettingGroup.DOWNLOADER,
--- a/docs/Configuring-Suwayomi‐Server.md
+++ b/docs/Configuring-Suwayomi‐Server.md
@@ -80,6 +80,7 @@ server.excludeEntryWithUnreadChapters = true
 server.autoDownloadNewChaptersLimit = 0
 server.autoDownloadIgnoreReUploads = false
 server.downloadConversions = {}
+server.splitTallImages = false
 ```
 - `server.downloadAsCbz = true` configures Suwayomi to automatically compress chapters into CBZ.
 - `server.downloadsPath = ""` the path where manga downloads will be stored, if the value is empty, the default directory `downloads` inside [the data directory](https://github.com/Suwayomi/Suwayomi-Server/wiki/The-Data-Directory) will be used. If you are on Windows the slashes `\` needs to be doubled(`\\`) or replaced with `/`
@@ -114,6 +115,7 @@ server.downloadConversions = {}
   
   This is an example curl command for what Suwayomi-Server will send to the conversion url: `curl -X POST "http://localhost:9999/convert" -F "image=@cat.png;type=image/png"`
 - `server.serveConversions = {}` configures optional image conversions before serving the image to the client. It follows the same format as `server.downloadConversions`.
+- `server.splitTallImages = false` controls if Suwayomi should split overly tall "long strip" pages into several smaller images after downloading a chapter, to improve reader performance.
 
 
 ### Updater
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt
@@ -24,6 +24,7 @@ import suwayomi.tachidesk.manga.impl.util.getChapterCachePath
 import suwayomi.tachidesk.manga.impl.util.source.GetSource.getSourceOrNull
 import suwayomi.tachidesk.manga.impl.util.storage.ImageResponse.getImageResponse
 import suwayomi.tachidesk.manga.impl.util.storage.ImageUtil
+import suwayomi.tachidesk.manga.impl.util.storage.TallImageSplitter
 import suwayomi.tachidesk.manga.model.table.ChapterTable
 import suwayomi.tachidesk.manga.model.table.MangaTable
 import suwayomi.tachidesk.manga.model.table.PageTable
@@ -195,6 +196,7 @@ object Page {
         val conversions = serverConfig.downloadConversions.value
         if (conversions.isEmpty() || !downloadCacheFolder.exists()) {
             inputStream.close()
+            splitTallImageIfNeeded(downloadCacheFolder, fileName)
             return
         }
         val defaultConversion = conversions["default"]
@@ -203,6 +205,7 @@ object Page {
                 ?: defaultConversion
         if (conversion == null) {
             inputStream.close()
+            splitTallImageIfNeeded(downloadCacheFolder, fileName)
             return
         }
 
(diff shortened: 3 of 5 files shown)


server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
```diff
@@ -219,6 +219,14 @@ class ServerConfig(
         defaultValue = false,
     )
 
+    val splitTallImages: MutableStateFlow<Boolean> by BooleanSetting(
```
Syer10: Move this to the bottom of the file under 97

Syer10 (changes requested):
Does this use the same logic as the Mihon version? It would be best if they were as similar as possible since it is battle tested.

yohanim:
Hi back @Syer10 , I've checked and asked a good comparison with references so that we can go look into files.
It's definitely based on the same algorithm, only real differences are the image library to do things and the choices to make because it's don on server side (I now have an idea to improve the fixed high it proposed and replace it with a ratio based on width ti make it even closer to mihon one)

I'll propose the improved version with the moving of the item at the right place (silly me I pushed fast to have something to discuss befor leaving 😄 )


## Files to compare

**Mihon** (`mihonapp/mihon`, `main` branch)
- [`TallImageSplitCalculator.kt`](https://github.com/mihonapp/mihon/blob/main/core/common/src/main/kotlin/tachiyomi/core/common/util/system/TallImageSplitCalculator.kt) — threshold calculation (full file, 17 lines)
- [`ImageUtil.kt#L188-L300`](https://github.com/mihonapp/mihon/blob/main/core/common/src/main/kotlin/tachiyomi/core/common/util/system/ImageUtil.kt#L188-L300) — `isTallImage`, `splitTallImage`, `splitData`, `SplitData`, `splitImageName`
- [`ImageUtil.kt#L557-L561`](https://github.com/mihonapp/mihon/blob/main/core/common/src/main/kotlin/tachiyomi/core/common/util/system/ImageUtil.kt#L557-L561) — `optimalImageHeight` (derived from the device screen)
- [`Downloader.kt#L544-L559`](https://github.com/mihonapp/mihon/blob/main/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt#L544-L559) — hook in the downloader (called at [line 450](https://github.com/mihonapp/mihon/blob/main/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt#L450))
- [`DownloadPreferences.kt#L22`](https://github.com/mihonapp/mihon/blob/main/domain/src/main/java/tachiyomi/domain/download/service/DownloadPreferences.kt#L22) — the setting (`split_tall_images`, default `true`)

**This PR** (`yohanim/Suwayomi-Server`, `feat/split-tall-images` branch)
- [`TallImageSplitter.kt`](https://github.com/yohanim/Suwayomi-Server/blob/feat/split-tall-images/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/util/storage/TallImageSplitter.kt) — the full algorithm (entire file)
- [`Page.kt#L196-L263`](https://github.com/yohanim/Suwayomi-Server/blob/feat/split-tall-images/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Page.kt#L196-L263) — hook in the download pipeline
- [`ServerConfig.kt#L222-L228`](https://github.com/yohanim/Suwayomi-Server/blob/feat/split-tall-images/server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt#L222-L228) — the setting (`splitTallImages`, default `false`)
- [`TallImageSplitterTest.kt`](https://github.com/yohanim/Suwayomi-Server/blob/feat/split-tall-images/server/src/test/kotlin/suwayomi/tachidesk/TallImageSplitterTest.kt) — tests

## The algorithm (same core logic on both sides)

1. **Decide whether an image should be split**: an image is "too tall" if `height > width × 3` (aspect ratio gate) **AND** its height exceeds a threshold `optimalHeight` such that the computed part count is `> 1`.
2. **Compute the part count**: `partCount = (height - 1) / optimalHeight + 1` (the `-1` avoids creating a useless empty extra part when `height` is an exact multiple of `optimalHeight`).
3. **Split into equal-height bands**: `partHeight = height / partCount`, each band spans the full image width; the **last band absorbs the division remainder** so the summed heights exactly match the original height.
4. **Write each band** out as a separate image file, then **delete the original** once every part has been written successfully.
5. **Safety**: if writing fails partway through, any partial split files already created are deleted and the original image is left untouched — a page can never go missing because of a failed split.

## Differences between the two implementations

| Aspect | Mihon (Android) | Suwayomi-Server (JVM, our port) |
|---|---|---|
| **Image decoding API** | `android.graphics.BitmapFactory` / `BitmapRegionDecoder` — decodes **only the needed region** per band, without loading the whole image into memory | `javax.imageio.ImageIO` — decodes the **entire** image into a `BufferedImage`, then uses `getSubimage()` to extract each band (no direct JVM-standard equivalent of `BitmapRegionDecoder`) |
| **Height threshold (`optimalImageHeight`)** | **Dynamic**, based on the device's screen: `2 × max(screen height, screen width)` — varies per phone/tablet | **Fixed** constant of `5000px` — a server has no screen to query, so a reasonable value was hardcoded |
| **Output format** | Always JPEG, **100%** quality (`Bitmap.compress(JPEG, 100, ...)`) | Always JPEG, **ImageIO writer's default** quality (~0.75–0.8); images with an alpha channel are flattened onto an opaque RGB background before writing (JPEG doesn't support alpha) |
| **Split file naming** | `{prefix}__{001}.jpg` (double underscore before the index) | `{fileName}.{001}.jpg` (dot before the index) — a deliberate choice to stay compatible with Suwayomi-Server's existing "page already downloaded" detection, which looks for files starting with `"$fileName."` |
| **Animated image detection** | Explicit check via a separate image-decoding library (`isAnimatedAndSupported`) | Generic check via `ImageReader.getNumImages(true) > 1` (works for multi-frame GIFs, for instance) |
| **File abstraction** | `com.hippo.unifile.UniFile` (Android abstraction over SAF/storage) | Plain `java.io.File` (no UniFile dependency in Suwayomi-Server) |
| **Hook point** | `Downloader.splitTallImageIfNeeded()`, called right after each page is saved to the app's temporary download folder | `Page.splitTallImageIfNeeded()`, called at the end of `getPageImageDownload()`, right after the page (converted or not) is saved to the chapter's download cache folder |
| **Setting default** | `true` (enabled by default) | `false` (disabled by default) — a deliberate choice so existing server users see no behavior change |
| **Impact on page count** | Not a concern: the Android reader just displays the files in a local folder in order, with no strict server-side "page N of M" contract | A potential issue was identified and verified: `ChapterForDownload.getChapterDownloadReady()` already recomputes `pageCount` from the actual number of files on disk, so no further change to that mechanism was needed |

**In short**: the core math of the splitting logic (threshold, part-count calculation, height distribution) was ported almost verbatim. The differences mostly come from the target platform — Android, with a screen to size against and memory-efficient `Bitmap` region-decoding APIs, versus a headless JVM server that has to rely on standard `ImageIO` APIs and a fixed threshold.

yohanim:
Done,
* documentation done for CI and quality of use
* item at the right place
* even closer to pure mihon algorithm (using a 21/9 ratio instead of the phones one)
* improved a bit splitted output image (reproduced original type and compression, else use the JPG 100% compression choise of mihon)

yohanim:
afterwards, we'd have to add the option in the webui.
I forced the option as inactive by default to not have unintended behavior change for all users from day one so we'd need UI option to set it active for some time before taking the same route of activated by default as mihon did 😄

Syer10:
Don't worry about the webui part. the WebUI dev will handle that when they need to

schroda:
> No changes were needed to page-count bookkeeping: ChapterForDownload.getChapterDownloadReady() already resyncs Chapter.pageCount to the number of files actually on disk, so chapters with split pages report the correct page count to clients automatically.

So what happens if I start reading a chapter while it's downloading?

yohanim:
## Known limitation: concurrent read/download race on the shared page cache

While analyzing this feature I found a pre-existing race condition that this PR makes more visible, and want to document it here rather than silently ship around it.

**The issue**: `ChapterTable.isDownloaded` stays `false` for the *entire* duration of a chapter download. While it's `false`, reading that chapter (WebUI, or any client) goes through `Page.getPageImage()`'s live-fetch path, which reads/writes the *same* per-chapter cache directory (`getChapterCachePath`) and the *same* filename convention as the download job (`ChaptersFilesProvider.downloadImpl`) - with no locking between the two.

Specifically for this PR's splitting feature, that means:
- If the download job has already split a page into `page.001.jpg`, `page.002.jpg`, ... by the time a concurrent live read looks up that page in the cache, the prefix-based cache lookup (`ImageResponse.findFileNameStartingWith`) can match one of those fragments and serve it as if it were the whole page - a visibly wrong/cropped image, with no error surfaced.
- If a live read caches a page first, the download job's own "page already exists" check will skip calling `getPageImageDownload()` for it entirely, so that page is silently never split (no error either, just quietly not split).

This isn't introduced by the splitting logic itself - the same shared-cache-without-locking pattern already exists for `downloadConversions` - but splitting turns a previously benign symptom ("page wasn't converted") into a more visible one ("reader shows a cropped image").

I confirmed this isn't just a theoretical edge case: Suwayomi-Tsumiru's "save chapter to device" flow (`offline_download_providers.dart`, `saveChapterToDevice`) deliberately enqueues the chapter on the server's download queue *and* fetches pages itself via the same `fetchChapterPages` mutation at the same time, explicitly not waiting on one before the other. So this is the default behavior whenever a Tsumiru user downloads a not-yet-server-downloaded chapter to their device, not a rare corner case.

Fixing this properly needs some form of per-chapter locking between the live read path and the download job, which is bigger in scope than this feature PR, so I've filed it separately: #2289. Flagging here so reviewers have the context when deciding whether this feature should wait on that fix or can land independently.

yohanim:
In short, it is a pre-existing problem as it also is present for image post-processing.
it has an edge behavior fro readers and for tsumiru download features.
to be corrected independently

server/server-config/src/main/kotlin/suwayomi/tachidesk/server/ServerConfig.kt
yohanim: see next comment for the analyse of the answer. same algorythm, reimplementation because of other technology

PR #2296, opened 2026-08-21
Feature/import sources for uninstalled extensions

schroda:

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
@@ -51,7 +51,6 @@ class SourceType(
     val isNsfw: Boolean,
     val displayName: String,
     val homeUrl: String?,
-    @GraphQLDeprecated("", ReplaceWith("homeUrl"))
     val baseUrl: String?,
 ) : Node {
     constructor(row: ResultRow, sourceExtension: ResultRow, source: Source?) : this(
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
@@ -65,7 +65,7 @@ class SourceType(
         isNsfw = row[SourceTable.contentWarning] >= ContentWarning.MIXED.ordinal,
         displayName = source.toString(),
         homeUrl = runCatching { (source as? HttpSource)?.getHomeUrl() }.getOrNull(),
-        baseUrl = runCatching { (source as? HttpSource)?.baseUrl }.getOrNull(),
+        baseUrl = row[SourceTable.baseUrl],
     )
 
     fun manga(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<MangaNodeList> =
--- a/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
@@ -54,13 +54,13 @@ class SourceType(
     @GraphQLDeprecated("", ReplaceWith("homeUrl"))
     val baseUrl: String?,
 ) : Node {
-    constructor(row: ResultRow, sourceExtension: ResultRow, source: Source) : this(
+    constructor(row: ResultRow, sourceExtension: ResultRow, source: Source?) : this(
         id = row[SourceTable.id].value,
         name = row[SourceTable.name],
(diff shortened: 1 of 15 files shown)


server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
```diff
@@ -51,21 +51,20 @@ class SourceType(
...
         name = row[SourceTable.name],
         lang = row[SourceTable.lang],
         contentWarning = ContentWarning.valueOf(row[SourceTable.contentWarning]),
         iconUrl = Extension.proxyExtensionIconUrl(sourceExtension[ExtensionTable.pkgName]),
-        supportsLatest = source.supportsLatest,
+        supportsLatest = source?.supportsLatest ?: false,
         isConfigurable = source is ConfigurableSource,
         isNsfw = row[SourceTable.contentWarning] >= ContentWarning.MIXED.ordinal,
         displayName = source.toString(),
         homeUrl = runCatching { (source as? HttpSource)?.getHomeUrl() }.getOrNull(),
```
schroda: Should the `homeUrl` return the `baseUrl` in case the source is not installed? I.e. keeping `baseUrl` deprecated?

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
```diff
@@ -528,24 +530,48 @@ object Extension {
...
 
-            val extensionId =
+            updateExtensionSourcesDatabase(
+                pkgName,
+                httpSources.map {
+                    ExtensionSource(
+                        id = it.id,
+                        name = it.name,
+                        lang = it.lang,
+                        baseUrl = (it as HttpSource).baseUrl,
```
Syer10: Wrap this in a runCatching

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
```diff
@@ -84,10 +83,7 @@ class SourceType(
 
 @Suppress("ktlint:standard:function-naming")
 suspend fun SourceType(row: ResultRow): SourceType? {
-    val catalogueSource =
-        GetSource
-            .getSourceOrNull(row[SourceTable.id].value)
-            ?: return null
+    val catalogueSource = GetSource.getSourceOrNull(row[SourceTable.id].value, false)
```
schroda, line 87: We have to get rid of this call because it basically nullifies the data loader. Which makes e.g. the `extensions` query super slow when selecting  the `sources` field.

`getSourceOrNull` makes two sql queries to get the source and the extension.

So I think we should add `supportsLatest`, `isConfigurable` and `homeUrl` to the `SourceTable`

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
Syer10: homeUrl is the new name of baseUrl in the newer extension libs. So only homeUrl should be available.

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
Syer10: Adding supportsLatest and isConfigurable sounds good to me, see my other comment for homeUrl

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
schroda: I'll do that in another PR since that issue exists already without these changes

server/src/main/kotlin/suwayomi/tachidesk/graphql/types/SourceType.kt
schroda: Fixed

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/ExtensionsList.kt
```diff
@@ -217,6 +217,8 @@ object ExtensionsList {
                 if (removeExtensions.isNotEmpty()) {
                     ExtensionTable.deleteWhere { ExtensionTable.pkgName inList removeExtensions }
                 }
+
+                uniqueExtensions.forEach { Extension.updateExtensionSourcesDatabase(it.pkgName, it.sources) }
```
schroda, line 221: I need to optimize this before this PR is ready

PR #2308, opened 2026-08-30
feat(sync): support SyncYomi protocol v2

kaiserbh:
Splitting this up for easier review:

- [x] #2310 — backup model proto3 defaults
- [x] #2311 — version trigger fixes (one migration)
- [x] #2312 — sync restore adopts server versions, flags passthrough, category reorder
- [x] protocol v2 itself — fresh PR once the above merge; this draft then closes

This branch stays as the integration branch until then.

PR #2312, opened 2026-08-31
fix(sync): adopt server versions on sync restore, pass flags through, fix category reorder

kaiserbh:
SyncYomi merges by version only, but the sync restore merged like a backup file (read || read, max progress) and never took the remote version, so a row that lost a merge once stayed behind the server forever — and rows written by the restore weren't flagged `is_syncing`, letting the triggers inflate versions during the restore itself.

Restores now carry a `SyncRestoreMode`: `NONE` keeps today's backup-file behaviour, `ADOPT` overwrites rows whose remote version is at least the local one and keeps newer local rows for the next upload, and `CONVERGE` runs once (tracked by a `sync_schema` pref) to repair version counters skewed by the old triggers — best merged in the same release as #2311.

Suwayomi also exported category `flags`, `chapterFlags`, `viewer` and `viewer_flags` as zeros and dropped them on restore, so whenever its copy won a merge the other client's library sort, chapter filters and reading mode reset to defaults; they're now stored as opaque passthrough columns (new migration). And the WebUI category reorder did raw `sort_order` arithmetic, which broke once a restore adopted TachiyomiSY's 0-based orders (a real category landed on order 0 next to the default category); it's now position-based.

Migration is numbered M0063 on this branch; whichever of this and #2311 merges second gets renumbered on rebase. Split out of #2308.


server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/CategoryMutation.kt
```diff
@@ -340,31 +340,8 @@ class CategoryMutation {
...
-                }
-
-                CategoryTable.update({ CategoryTable.id eq categoryId }) {
-                    it[CategoryTable.order] = position
-                }
-            }
-        }
-
-        Category.normalizeCategories()
+        // position-based: raw sort_order values can collide after a sync adopts a peer's 0-based orders
```
schroda: Can you give an example of what exactly this is fixing?

kaiserbh: TachiyomiSY uses 0 index based, and Suwayomi uses 1, so when we are moving categories around it will be out of position and won't sync properly.

`BackupCategoryHandler.restore` writes the incoming `order` values as-is (sync restore or a plain restore of an SY backup), so after that a real category sits at `sort_order = 0` next to the Default category:

    Default (id 0)  sort_order 0
    A               sort_order 0   <- adopted from the backup
    B               sort_order 1
    C               sort_order 2

Drag C to the top: the WebUI sends position 1, the old code shifts everything with `order >= 1` up (B 1→2, C 2→3) and sets C = 1 — but A at 0 never moves, so after `normalizeCategories()` A is still first and the drag does nothing. Position 0 itself can't even be requested because of the `position > 0` require. With sync on it's worse: the renumbering still bumps versions, so the wrong order wins the merge and spreads to the other device.

Moving by position in the ordered non-default list behaves the same on a healthy 1-based table and correctly on an adopted 0-based one (the first move renumbers everything back to 1-based).

Syer10 (approved):

server/src/main/kotlin/suwayomi/tachidesk/graphql/mutations/CategoryMutation.kt
schroda: So we basically have a "broken" category ordering in the database after a sync as long as the user is not reordering the categories in suwayomi?

I guess it's not a problem because we force the base ordering by the category id when quering categories, but I wouldn't call it ideal either

kaiserbh: Yeah exactly - it sits there until the user reorders, then the first move renumbers everything back to 1-based. Adopting the orders verbatim was deliberate so the local rows match the server copy and we don't re-restore categories every sync.

Agreed it's not ideal though - I can add an id tiebreak to the `orderBy(sort_order)` queries so ties at least display deterministically. Let me know if you want me to?

@schroda

kaiserbh: The ideal fix imo is refactoring Suwayomi to 0-based ordering like the rest of the ecosystem (Mihon/SY/J2K), with the Default category excluded from the ordering entirely - it's already excluded from reorder/upload and pinned first by id anyway. Then adopted peer orders are just canonical instead of "broken", and this whole collision class disappears.

I'm willing to do that refactor - I'll open it as a separate PR focused on just that, so this one stays about fixing the reorder itself.

schroda: Would it make sense to convert the category order when syncing? I.e. when sending make it 0-based and when receiving convert it to be 1-based?

kaiserbh: Hmm yea I think that's actually a good idea. Let me implement that here and push a fix.

kaiserbh: Cool fixed and tested seems to be working.

Thanks for the suggestion @schroda

schroda (approved):

PR #2319, opened 2026-09-01
Fix extensions incorrectly marked as updatable

schroda:
Due to an extension store fixing a version code mismatch, some extensions are now incorrectly marked as updatable. However, these are still pointing to the same version already installed. This makes updating them impossible. The only solution to fix this manually is reinstalling the extension.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/ExtensionsList.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/ExtensionsList.kt
@@ -143,7 +143,8 @@ object ExtensionsList {
 
                                         this[ExtensionTable.hasUpdate] = false
                                     }
-                            }
+                            }.toExecutable()
+                            .execute(this@transaction)
 
                         BatchUpdateStatement(ExtensionTable)
                             .apply {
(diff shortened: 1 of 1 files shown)


Syer10:
🤔 
Do we want to run something like this before every fetch extensions from store instead?

schroda:
Makes sense. It would ensure that this issue would fix itself in the future.

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/ExtensionsList.kt
```diff
@@ -119,12 +121,30 @@ object ExtensionsList {
...
+                        BatchUpdateStatement(ExtensionTable)
+                            .apply {
+                                installedExtensionsToUpdate
+                                    .filter { it.second[ExtensionTable.hasUpdate] }
+                                    .forEach { (_, extension) ->
+                                        addBatch(EntityID(extension[ExtensionTable.id].value, ExtensionTable))
+
+                                        this[ExtensionTable.hasUpdate] = false
+                                    }
+                            }
```
Syer10:
```suggestion
                            }.toExecutable()
                            .execute(this@transaction)
```

Syer10 (changes requested):

Syer10 (approved):

PR #2320, opened 2026-09-01
Ignore extension file deletion issues during installation/update

schroda:
When uninstalling/updating an extension, we unload the jar. However, there is no guarantee that the loaded jar gets fully unloaded. A jar that was not fully unloaded won't be able to be deleted on windows due to its strict file locking. On linux this is not an issue because it allows one to delete jars that are still loaded.

### DIFF
--- a/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
+++ b/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
@@ -7,6 +7,8 @@ package suwayomi.tachidesk.manga.impl.extension
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+import android.app.Application
+import android.content.Context
 import android.content.pm.PackageInfo
 import android.net.Uri
 import eu.kanade.tachiyomi.network.GET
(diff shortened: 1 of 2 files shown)


server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
```diff
@@ -687,6 +700,30 @@ object Extension {
         }
     }
 
+    private val filesToDeleteOnStartup = ConcurrentHashMap.newKeySet<String>()
+    private val preferences = Injekt.get<Application>().getSharedPreferences(Extension.javaClass.name, Context.MODE_PRIVATE)
+    private const val FILES_TO_DELETE = "files_to_delete"
+
+    fun cleanupExtensionFiles() {
+        logger.debug { "Cleanup extension files that could not be deleted during extension update/deinstallation" }
+        preferences.getStringSet(FILES_TO_DELETE, emptySet())?.forEach { Path(it).deleteIfExists() }
```
Syer10: Doesn't this need to clean the `FILES_TO_DELETE` preference? Also should surround the deletion with a try/catch.

Syer10 (changes requested):

server/src/main/kotlin/suwayomi/tachidesk/manga/impl/extension/Extension.kt
schroda: fixed

Syer10 (approved):

